Issue #112, opened 2023-10-20
Create User Profile Page - Backend


DhananjayThomble:
Develop the backend logic to support the user profile page on the frontend, allowing users to view and edit their profile information.

**Tasks:**
- Create API endpoints for retrieving and updating user profile information.
- Implement the necessary database operations to retrieve and update user data.
- Ensure proper validation and error handling on the backend.

Honey2339:
Interesting , I would like to contribute

Honey2339:
Im setting up the application and i came across Email Provider , Can u help me how to setup the Email Provider part?

samuelogboye:
@Honey2339 if you need help, ill love to join you

Honey2339:
@samuelogboye Yeah sure! , i need Email related environment variables so that i can test the development server

DhananjayThomble:
> @samuelogboye Yeah sure! , i need Email related environment variables so that i can test the development server

You can refer this : https://youtube.com/shorts/_INB5cm29LQ?si=Na-rImQSXKfxP2HR

Alternatively, you can even use your own Gmail account. Just search for "how to use Nodemailer with Gmail"

Honey2339:
Please review the above mentioned PR

DhananjayThomble:
> Please review the above mentioned PR

Thanks @Honey2339, I've merged the PR. I'd like to ask if you could add validators for this API. You can use the signup API as a reference.

PR #118, opened 2023-10-21
#112 Added Endpoint to retrieve user data and updating user data

Honey2339:
1- Added a endpoint to see current user data
2- Added a endpoint to change email
3- Added a endpoint to change the name

### DIFF
--- a/backend/controllers/CurrentUserController/getCurrentUser.controller.js
+++ b/backend/controllers/CurrentUserController/getCurrentUser.controller.js
@@ -1,20 +1,21 @@
-import jwt from 'jsonwebtoken';
+import jwt from "jsonwebtoken";
 import User from "../../models/UserModel.js";
 import dotenv from "dotenv";
 dotenv.config();
 
 export const getCurrentUser = (req, res) => {
   // Check if a token is present in the request headers
-  const token = req.headers.authorization;
+  // split the token from the "Bearer" string
+  const token = req.headers.authorization?.split(" ")[1];
 
   if (!token) {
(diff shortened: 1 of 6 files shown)


backend/controllers/CurrentUserController/getCurrentUser.controller.js
```diff
@@ -0,0 +1,33 @@
+import jwt from 'jsonwebtoken';
+import User from "../../models/UserModel.js";
+import dotenv from "dotenv";
+dotenv.config();
+
+export const getCurrentUser = (req, res) => {
+  // Check if a token is present in the request headers
+  const token = req.headers.authorization;
```
DhananjayThomble, line 8: We can reuse `middlewares/authMiddleware.js` wherever we need to extract and verify a JWT.

backend/routes/auth.route.js
```diff
@@ -15,5 +18,8 @@ router.post("/signup", validateSignup, validationErrorHandler, signup);
 router.post("/forgot-password", forgetPasswordValidator, validationErrorHandler, forgotPassword);  //For Sending the password reset request
 router.post("/reset-password", resetPassword);   // For Reseting the password
 router.get("/verify-email", verifyEmail);
+router.get("/current-user" , getCurrentUser)
+router.post("/change-email" , changeEamil)
+router.post("/change-name" , changeName)
```
DhananjayThomble, line 23: I think here 'patch' will be more appropriate than the 'post' since you're only updating a part of the user resource (either the email or the name), it would be more appropriate to use PATCH

DhananjayThomble (approved):

Issue #290, opened 2026-08-29
Make the documentation true


DhananjayThomble:
Final phase of the architecture epic. For a project meant to teach, documentation describing a system that does not exist is worse than none â€” and under the open-source framing it actively misleads operators.

## Missing file

`docs/DEPLOYMENT.md` links to `docs/ARCHITECTURE.md`, **which does not exist.**

## Claims in DECISIONS.md that the code contradicts

| Claim | Reality |
| --- | --- |
| "The counter is a DynamoDB conditional write" | A Postgres column updated by the rollup worker |
| "Background work: SQS + Lambda" | EventBridge only. No SQS, and no AWS SDK in any app |
| "CloudFront already absorbs volume" | `CACHING_DISABLED` â€” it absorbs nothing |
| "GeoIP uses `CloudFront-Viewer-Country`" | Not forwarded by the origin request policy; always null |
| "The redirect reads link config from DynamoDB" | `PostgresLinkResolver` is the only implementation |
| Redis rejected because "DynamoDB, SQS or CloudFront serve every use" | True for the AWS profile only; being applied to all of them |

Phases 3 through 7 make most of these true. The rest must be corrected rather than left.

## Also

- Write `docs/ARCHITECTURE.md` around the three profiles and the port boundaries, since that is now the thing worth explaining.
- Record the per-profile decisions the epic introduces, including the egress call and the click-accounting call.
- The cost table in the deployment plan claims marginal cost per 1M redirects falls from $1.57 to ~$1.10 after adding a WAF. WAF bills per request inspected, including edge-served ones, so that figure should rise. Recompute before publishing.
- State plainly that at ~$15.50/month the $100 of credits lasts about six and a half months, not twelve.

Issue #278, opened 2026-08-29
Bound the failure-mode bill: reserved concurrency, budget alarms, log retention


DhananjayThomble:
Phase 3 of the architecture epic. Three independent guard rails, none of which exist. `grep` for `reservedConcurrentExecutions`, `CfnBudget` and `logRetention` across `infra/` returns nothing.

## Reserved concurrency

No function has a concurrency cap. `db.t4g.micro` allows ~112 connections, minus 3 reserved for superuser, so ~109 usable â€” and one is needed by the migrate task. Combined with the 10-second timeout amplifier, uncapped concurrency is a self-inflicted outage with a bill attached.

Set redirect **80**, api **20**, worker **2**. That totals 102 and leaves headroom.

Note: the original audit proposed 100/20/2, which is 122 and exceeds the connection ceiling it was intended to respect. The cap is right, the arithmetic was not.

## Budgets

No AWS Budgets alarm anywhere. Credits mask overspend until they are gone. Add budgets at $25 / $50 / $75 with an SNS email subscription.

Worth recording in the epic: at the current ~$15.50/month the $100 of credits lasts about six and a half months, not twelve. The premise that drove the serverless design is not currently satisfied by it.

## Log retention

CDK's default is Never Expire, and Fastify logs two JSON lines per request at `info`. At 100M redirects/month that is ~60 GB/month ingested with storage compounding forever.

- Retention of two weeks on all three log groups. Use an explicit `logs.LogGroup` with the `logGroup` prop rather than the deprecated `logRetention` prop, which provisions a custom-resource Lambda.
- `LOG_LEVEL=warn` for the redirect function specifically.

## Done when

- Concurrency is capped on all three functions and total possible connections stay under the RDS ceiling.
- A load generator past the cap is rejected quickly rather than producing 10-second timeouts.
- Budget alarms deliver to a subscribed address.
- No log group retains indefinitely.

Issue #267, opened 2026-08-29
Architecture: three deployment profiles, container-first core


DhananjayThomble:
## Why

Two pre-deployment audits found that the CDK stack as written would deploy a product that does not work at all: every redirect returns 404 through CloudFront, every country-routing rule silently misses, and clicks are written into a query Lambda suspends mid-flight. None of it is visible to the 218 tests or to CI, because none of it exists until CloudFront is in front.

The deeper finding is not any single bug. The codebase already contains the right design — `LinkResolver`, `ClickSink`, `ProjectionTarget` and `MailerPort` are real ports with real seams — and only the local-development adapter was built for each half of it. The CDK stack then deploys those local adapters to Lambda as if they were production. The result has the cost profile of a serverless design and the latency and concurrency profile of one small Postgres box.

This epic reframes the work around the actual goal: **SnapURL is an open-source product that other people and organisations will host on their own AWS accounts or their own hardware.** That makes the ports-and-adapters core the product rather than an implementation detail, and it makes "Lambda or containers" a per-deployment choice rather than a decision made once.

## The architecture

A stateless core behind ports, with the raw event stream partitioned and the read path cached. Three deployment profiles, shipped in this order:

| Profile | Who runs it | Adapters |
| --- | --- | --- |
| **Single node** — the reference implementation | Most self-hosters | Postgres throughout, in-memory cache, in-process click batching |
| **Horizontally scaled, cloud-agnostic** | Organisations on EKS / GKE / k3s / bare metal | Postgres primary + replicas, Redis for cache, counters and queue, optional columnar analytics |
| **AWS-native serverless** | Cheapest-possible AWS, and the edge-latency showcase | Lambda, DynamoDB projection, SQS, CloudFront Function + KeyValueStore |

The AWS edge design is kept. A CloudFront Function plus KeyValueStore in front of a DynamoDB projection fed by an outbox is roughly how a shortener is built at scale, and that judgement was correct. What changes is its position: it becomes the **last** profile finished rather than the first, because it is the one fewest operators will run and the only one that cannot be verified on a laptop.

## Ordering, and why

Sequenced to put the expensive-to-reverse work first. Code authorship is cheap; migrations against live data, operational burden carried by other people, and insecure defaults already in someone else's hands are not.

### Phase 1 — Data model

Migrations get strictly harder once real data exists. After that they are backfills. These four ceilings break identically in every profile.

[...]

PR #293, opened 2026-08-29
Partition click_events by day, and expire it by dropping partitions

### LINKED ISSUE #268: Partition click_events by day, and expire it with DETACH/DROP instead of DELETE
Phase 1 of the architecture epic. **The single highest-value scale change in the codebase**, and it is independent of any cloud.

## Problem

`packages/database/src/schema/analytics.ts` declares `click_events` as a plain heap table. There is no partitioning anywhere in the schema â€” `grep -i partition packages/database` returns nothing.

At 1,000 req/s that is roughly 86M rows per day into one table. Two consequences:

1. `pruneRetention` in `apps/worker/src/jobs/rollup.ts` expires rows with `DELETE FROM click_events USING workspaces ...`. A bulk `DELETE` at that volume generates WAL faster than a replica can consume it, bloats the heap, and outruns autovacuum. Retention is currently a default of **3 years**, so this only ever gets worse.
2. `click_events_link_time_idx` on `(link_id, occurred_at DESC)` is a single btree spanning every row ever recorded. At hundreds of millions of rows it is far larger than any reasonable working set.

## Fix

- Convert `click_events` to a declaratively range-partitioned table on `occurred_at`, one partition per day (or week â€” see below).
- Pre-create partitions ahead of time in the maintenance job; a missing partition must never fail an insert.
- Replace the retention `DELETE` with `ALTER TABLE ... DETACH PARTITION` followed by `DROP TABLE`. Constant-time, no WAL storm, no vacuum debt.
- Indexes become per-partition, so they stay resident.
- Note the primary key: `uuidv7()` on `id` alone cannot remain the PK on a partitioned table unless the partition key is included. The PK becomes `(id, occurred_at)`, and the FKs pointing at `click_events` (there are none today) stay unaffected.

## Open sub-decision

Day vs week partitions. Day is better for retention granularity and index size; week means 52 partitions/year instead of 365 and less planner overhead. Recommend **day**, with retention rounding to whole partitions.

## Done when

- A migration converts the table without data loss, and is safe to run against a populated database.
- `pruneRetention` no longer issues a row-level `DELETE` against `click_events`.
- The maintenance job provisions future partitions, and a test proves an insert into a not-yet-created day still succeeds.
- The retention setting is documented as rounding to whole partitions.

DhananjayThomble:
## What does this PR do?

Makes `click_events` a RANGE-partitioned table, one partition per UTC day, and turns retention from a row-level `DELETE` into a partition `DETACH` + `DROP`.

This is the first item of #267 (Phase 1, data model), and it is first for a reason that is briefly true and will stop being true: nothing is deployed yet, so this is a schema change. Once there is a production database it becomes a backfill.

Fixes #268

## Why

At the volume this table is designed for — roughly 86M rows a day at 1,000 clicks per second — two things break, and one change fixes both.

**Retention.** `DELETE FROM click_events USING workspaces` generates WAL faster than a replica can consume it, bloats the heap, and outruns autovacuum. The default retention is three years, so the problem only compounds. Dropping a partition is a catalogue change: constant time, almost no WAL, nothing left for vacuum to chase.

**Index size.** `click_events_link_time_idx` was a single btree spanning every row ever recorded. Declared on the partitioned parent it becomes per-partition, so the part actually being queried stays resident in memory.

## Two things that are easy to get wrong

Both are covered by tests, because both are silent when wrong.

**A `DEFAULT` partition exists permanently.** Without one, an insert for a day with no partition *fails* — meaning the redirect path would stop recording clicks because a housekeeping job did not run. That trade is the wrong way round, so the table always has somewhere to put a row.

**`click_events_ensure_partition` builds detached, drains, then attaches.** Postgres scans the default partition on `ATTACH` and refuses if any row there belongs to the incoming range. A naive `CREATE TABLE ... PARTITION OF` therefore breaks *permanently* the first time the default is used, and the table cannot be repaired without manual surgery. The function also tolerates losing a race to a concurrent worker — with reserved concurrency of 2 (#278) and EventBridge retries, that is an ordinary Tuesday rather than a corner case.

## Retention keeps two mechanisms, deliberately

Retention is **per workspace**; a partition is **shared**. So a day can only be dropped once it is past the *longest* retention any workspace has configured. Workspaces retaining less than that maximum still need their rows removed individually from the partitions that survive.

When every workspace uses the same retention — the common case, and the default — `rowsDeleted` is zero and the whole job is a partition drop. The two counts are logged separately because they cost wildly different amounts, so a steadily non-zero `clickRowsDeleted` is a visible signal that mixed retention is in use rather than something to discover from a disk graph.

## Notes for review

- **Day boundaries are explicitly UTC**, matching the rollups, which all aggregate on `(occurred_at at time zone 'UTC')::date`. Letting `date::timestamptz` adopt the session TimeZone would file a click under the wrong day for anyone east of Greenwich.
- **The primary key becomes `(id, occurred_at)`** because Postgres requires the partition key in every unique constraint on a partitioned table. The constraint name is the one `drizzle-kit` generates, so the next `drizzle-kit generate` sees no drift. Worth knowing: `id` is no longer unique on its own as far as Postgres is concerned, so a lookup by `id` that does not also constrain `occurred_at` fans out across every partition. Nothing does that today — the rollup joins through its temp batch table — and it is the thing to check before adding one.
- **Drizzle has no vocabulary for partitioning**, so the parent, the children and the two maintenance functions are hand-written SQL in the migration. The schema declaration carries only what changes about the table's *shape*, which is why every existing query is untouched.
- **`ensureClickPartitions` runs first in `runMaintenance`**, before anything that can fail, so a pass that dies halfway has still done the one job the redirect path depends on.
- The `.gitignore` commit is unrelated housekeeping, kept separate: `.gh-token` was ignored only via `.git/info/exclude`, which lives in one clone and protects no contributor.

## Type of change

- [x] Chore (refactoring code, workflow improvements)

## How should this be tested?

No new environment variables. `DATABASE_URL` as usual; the partition tests skip without it, matching `rollup.test.ts`.

```
pnpm db:up && pnpm db:migrate
```

Both migration paths were exercised locally against a real Postgres 18:

- **Empty database** — the full chain 0000–0007 applies; the table comes out partitioned with a `DEFAULT` plus 15 day partitions, and the PK is `click_events_id_occurred_at_pk`.
- **Populated database** — run against 87 existing rows: all 87 preserved, nothing stranded in the `DEFAULT` partition, `click_events_legacy` dropped.

The drop path was verified for its boundary rather than just its happy case: a partition whose range ends *on* the cutoff is kept, one ending before it is dropped, and the `DEFAULT` and future partitions survive a retention pass.

`apps/worker/src/jobs/partitions.test.ts` adds 10 tests and was run three times consecutively to confirm it is repeatable — the first version was not, because partitions are schema objects and the fixture workspace cascade does not remove them.

Results: **228 tests pass** (218 before, 10 new). `scripts/smoke.sh` 105/105 and `scripts/smoke-redirect.sh` 35/35.

## Checklist

`docs/DECISIONS.md` will need a line about partitioned retention, but that is #290's job — it is currently wrong about six other things and is being corrected in one pass rather than piecemeal.


DhananjayThomble:
## Review response — 81009d3

A design review returned NEEDS_CHANGES on the previous head with one critical finding. CI was green at the time, which is worth noting on its own: none of this was visible to the test suite.

The critical one was mine twice over.

**`rollupClicks` lost partition pruning.** It stamped `rolled_up_at` matching on `id` alone. With a composite `(id, occurred_at)` primary key, `id` is not unique as far as Postgres is concerned, so the planner cannot prune — it opens every partition. On a three-year database that is ~1100 relations, on the job EventBridge invokes **every minute**. Before partitioning it was a plain primary-key lookup, so this PR would have made the hottest job in the system dramatically worse while claiming to make the table faster.

And the comment I wrote in `analytics.ts` said *"Nothing does that today — the rollup joins on the temp batch table."* That was false when I wrote it. I asserted an audit I had not actually performed. It now names the query it was wrong about.

Fixed by matching the full key and bounding `occurred_at` by the batch's own min and max as sub-selects — they become InitPlans, which the executor can prune with. `rollup_batch` is `select *`, so `occurred_at` was already there.

### The rest, by severity

**ATTACH raced the drain.** The drain ran before any lock was taken, so a click for the target day committing in between routed to the default and failed the ATTACH validation scan. Not exotic — draining only matters when the default already holds rows for that day, which means the day is live and traffic is still arriving. Now takes `ACCESS EXCLUSIVE` on the default first, then drains, then attaches. That also stops a large backlog being copied at full WAL cost and rolled back on every pass with no progress.

**The ATTACH handler caught the wrong SQLSTATEs.** It caught `invalid_table_definition` / `invalid_object_definition`. The failures this design actually produces are `check_violation` (default holds a row in range), `wrong_object_type` (already a partition) and `datatype_mismatch` (stale orphan) — all three uncaught. Since provisioning runs first in the maintenance pass, any of them would have skipped salt rotation and every prune behind it.

**Expiry blocked the write path.** `DETACH` takes `ACCESS EXCLUSIVE` on the **parent**, so it blocks inserts into every other partition too, and `DETACH CONCURRENTLY` is unavailable *because* this table has a DEFAULT partition — the safety net that makes inserts unfailable is what forecloses the low-lock detach. The old function looped every spent partition in one transaction with no timeout and no cap. Steady state is one a day; the moment the longest retention drops it becomes hundreds under one lock hold. Now split into a lister plus a single-partition dropper, one transaction each, bounded timeout, capped per pass, driven from TypeScript. `pruneRetention` also takes a session advisory lock so two overlapping passes cannot collide on a partition one already dropped.

**Two retention holes.**

The drop path ignored `rolled_up_at`. The `DELETE` it replaced deliberately spared un-counted rows; dropping the partition instead loses those clicks from `click_events` *and* the rollups at once, silently under-counting forever. Spent partitions holding pending rows are now skipped until the rollup catches up.

Rows stranded in the DEFAULT partition were unreachable by **either** mechanism — the drop pass skips the default by design, and the `DELETE` had an `occurred_at >= cutoff` bound excluding exactly the default rows past every retention. That bound pruned nothing anyway: after a drop pass every attached day partition is already at or above the cutoff, and a lower bound cannot prune a default partition at all. So it cost a retention guarantee and bought no elimination. Removed.

**Smaller:** provisioning look-back widened from 1 day to 7 to match the migration's backfill, so a gap longer than a day is repaired instead of leaving rows in the default permanently; a reused orphan is checked for column compatibility, since `ADD COLUMN` does not reach a detached table; `SET LOCAL lock_timeout` is restored rather than leaking to the caller's transaction (in the migration it covered the full-table copy and the legacy `DROP`); `current_date` replaced with `(now() at time zone 'UTC')::date`, since every bound is UTC-explicit and mixing them is the bug this migration's own header warns about; and provisioning failure is isolated so it cannot take `rotateSalts` down with it.

### Tests

Two added for the paths carrying data-loss risk that had none:

- a spent partition holding un-rolled-up clicks is **not** dropped, and becomes droppable once counted
- a workspace retaining less than the longest retention still has its rows deleted from a surviving partition — the entire reason `rowsDeleted` exists, previously unexercised

230 tests pass (218 before this PR, 12 new).

### Deliberately not in this PR

Three findings are real but are runbook or policy rather than code-local, and are filed separately so this PR does not grow another axis:

- **The migration's lock footprint on a large existing table.** Drizzle wraps all pending migrations in one transaction, so `RENAME` → partition creation → full copy → `DROP` is a single `ACCESS EXCLUSIVE` hold on `click_events`, plus `SHARE ROW EXCLUSIVE` on `links` and `workspaces` from the two `ADD FOREIGN KEY` statements. Footprint is ~5 lock slots per day of history, so past roughly 3.5 years it risks exhausting the default shared lock table. Not blocking here — this table has never been deployed — but it matters for self-hosters with existing data.
- **Retention became a global setting.** The cutoff is `max(retention_years)` and the contract allows up to 100 years, so one workspace can disable partition drops for the whole install and revert everyone to the row-level `DELETE` this PR exists to remove.
- **`fileParallelism: false` only serialises within the worker package.** `pnpm -r test` still runs packages concurrently against the same database. Nothing else writes clicks today, so nothing collides, but the guarantee is narrower than the comment implies.

Issue #324, opened 2026-08-30
Partition maintenance: lock-order inversion, silent contention, and an over-long default lock hold


DhananjayThomble:
Found while re-reviewing #293 after it merged. Both parts confirmed empirically against Postgres 18.

## 1. Lock-order inversion between the two maintenance functions

The two functions acquire the same two locks in opposite orders:

- `click_events_ensure_partition` takes **default → parent** (`LOCK TABLE click_events_default IN ACCESS EXCLUSIVE MODE`, then `ATTACH PARTITION`).
- `click_events_drop_partition` takes **parent → default** (`DETACH PARTITION` needs both; verified it times out at exactly 2.003s against a held `ACCESS EXCLUSIVE` on the default).

Two overlapping maintenance passes — one in `ensure`, one in `prune` — can therefore deadlock.

The consequences are bounded, which is why this is a medium and not a high: both handlers catch `deadlock_detected`, and both are capped by a 2s `lock_timeout`, so the outcome is "nothing happened this pass" rather than an aborted maintenance cycle or lost data.

Worth fixing by making both take the locks in the same order, so the deadlock is structurally impossible rather than merely survivable.

## 2. Contention is silent

This is the part that matters more, and it compounds the above.

`click_events_drop_partition` returns `false` for three different outcomes:

- lock timeout (`lock_not_available`)
- deadlock (`deadlock_detected`)
- already dropped by a concurrent pass (`undefined_table`)

`pruneRetention` then reports only `partitionsDropped`, so **sustained lock contention is indistinguishable from "nothing needed dropping"** — both read `partitionsDropped: 0`. Retention can stall indefinitely with no signal.

The same ambiguity makes the test assertion `expect(result.partitionsDropped).toBe(0)` weaker than it looks: it passes whether the partition was correctly skipped for holding un-rolled-up rows, or simply because a lock could not be taken.

Fix: distinguish the outcomes. Either return a reason from the SQL function rather than a boolean, or count timeouts and deadlocks separately in `RetentionResult` and log them. A non-zero contention count is the signal that partition maintenance is losing to write traffic, which is exactly the thing an operator needs to see and currently cannot.

## 3. The provisioning lock is held longer than its comment implies

[...]

Issue #283, opened 2026-08-29
Small correctness and cost items found during audit


DhananjayThomble:
Phase 3 of the architecture epic. Independent, each worth its own commit, none large enough for its own issue.

| Fix | Where | Why |
| --- | --- | --- |
| Split `runMaintenance` off the 1-minute schedule | `apps/worker/src/lambda.ts`, `infra/lib/snapurl-stack.ts` | The handler calls `runFrequent` **and** `runMaintenance` on every invocation, and the EventBridge rule fires every minute â€” so the hourly job runs 60x per hour against a `t4g.micro` that is also serving redirects |
| Parallelise the two resolver queries | `apps/redirect/src/resolver.ts` | `resolve()` awaits the link row, then the routing rules, sequentially. The second usually returns zero rows. One RTT saved on every redirect |
| `Architecture.ARM_64` | all three functions | ~20% cheaper per GB-s and typically faster. The RDS instance is already Graviton |
| Redirect memory to 1024 MB | `infra/lib/snapurl-stack.ts` | 512 MB is roughly a third of a vCPU. Double the memory finishes in half the time for the same GB-s, and halves cold start |
| Drop unused `argon2` | `apps/redirect/package.json` | Declared, never imported. A native module bloating the hot-path image for nothing |
| Volume cap on `click_events` | `apps/worker/src/jobs/rollup.ts` | `pruneRetention` only deletes by age, default **3 years**. Storage autoscales to 50 GB and then RDS goes read-only, and allocated storage can never be reduced. Largely superseded by partitioning, but the cap is still needed |

Issue #277, opened 2026-08-29
Fire-and-forget click writes are lost, and deadlock the warm Lambda instance


DhananjayThomble:
Phase 3 of the architecture epic. This is the actual mechanism behind the "unbounded failure-mode bill" finding, and it is reachable by ordinary traffic with no attacker involved.

## Problem

`apps/redirect/src/main.ts` calls `void record(...)` â€” deliberately fire-and-forget, on the documented reasoning that a redirect must not wait on analytics. Under the Lambda Web Adapter the sandbox **freezes the moment Fastify responds**, suspending the in-flight `INSERT`.

Three consequences, and the third is the expensive one:

1. Clicks are silently lost.
2. A suspended query holds a Postgres backend the server cannot reclaim.
3. `DATABASE_POOL_MAX=1` in production, and postgres.js *queues* when the pool is saturated. The next request on that warm instance waits behind a query that will never complete, until the function's 10-second timeout. `connect_timeout: 10` in `packages/database/src/client.ts` is the fuse.

So every warm redirect instance breaks itself after serving one request, and each broken invocation bills 10 seconds instead of ~20 ms â€” a 500x jump on the failure path.

## Fix

Phase 3 (now): `await` the click write. It is one round trip to a database in the same AZ, single-digit milliseconds, and vastly cheaper than a deadlocked instance. Keep the failure path non-fatal so analytics being down still cannot break a redirect.

Phase 7 (later): the SQS `ClickSink` adapter the port was designed for. An HTTP send with a bounded flush survives freeze/thaw in a way a Postgres `INSERT` does not.

## Done when

- No database write is left unawaited across a response boundary.
- A load test shows steady-state latency and no 10-second invocations.
- The integration harness asserts that a redirect produces exactly one `click_events` row.

PR #325, opened 2026-08-30
Replace retention's session advisory lock with an expiring lease

### LINKED ISSUE #323: pruneRetention's session advisory lock can strand and silently stop retention forever
Found while re-reviewing #293 after it merged. Latent, not currently biting, and the failure mode is silent — which is why it is worth fixing before it does.

## Problem

`pruneRetention` in `apps/worker/src/jobs/rollup.ts` guards the pass with a **session-level** advisory lock:

```ts
select pg_try_advisory_lock(hashtext('click_events_prune_retention')) as locked
...
finally {
  await db.execute(sql`select pg_advisory_unlock(hashtext('click_events_prune_retention'))`);
}
```

Session-scoped rather than transaction-scoped is the right call for the mechanism — the drops deliberately span several transactions so the parent `ACCESS EXCLUSIVE` lock is released between them, and `pg_try_advisory_xact_lock` would release at the first commit.

The problem is that a session advisory lock lives on the **connection**, and nothing bounds how long it can be held:

- postgres.js pools connections, and the AWS profile sets `DATABASE_POOL_MAX=1`.
- The `finally` covers a thrown error, but not the process disappearing. A Lambda frozen or reclaimed mid-invocation — a hazard this codebase already documents on the redirect path, and the whole reason #277 existed — leaves the lock held on a backend that is still alive from Postgres's point of view.
- Every later pass then takes the early return.

Retention stops permanently and silently. There is no error, no log line, and `partitionsDropped: 0` is indistinguishable from "nothing to drop". The first symptom is a disk graph, or RDS going read-only at the 50 GB autoscale ceiling.

## Why not simply remove it

I tried that first and it was wrong. `b4ad8b5` (#283) added `MAX_RETAINED_DAYS`, an age-independent volume cap with its **own** drop loop that drops partitions `click_events_spent_partitions` never listed. With two independent drop loops in one function, the pass-level guard is load-bearing rather than decorative. `click_events_drop_partition` catching `undefined_table` makes each individual drop idempotent, but that is not the same as making the pass safe.

## Fix: a lease that cannot outlive its holder

Replace the advisory lock with a lease row — a single-row table (or a row in an existing settings table) holding `locked_until timestamptz`:

- **Acquire** is a conditional update: `update ... set locked_until = now() + interval '5 minutes' where name = 'prune_retention' and locked_until < now() returning 1`. Got a row, you hold the lease.
- **Release** sets `locked_until = now()`.
- **Crash** needs no cleanup: the lease expires on its own.

The TTL should comfortably exceed a worst-case pass — `MAX_PARTITION_DROPS` drops at a 2s `lock_timeout` each, plus the cap loop — so a slow pass never has its lease stolen mid-run. Five minutes matches the worker Lambda's own timeout, which is a natural ceiling on how long a pass can be alive at all.

This is the standard shape for cross-transaction mutual exclusion where the holder can vanish, and it is strictly better than an advisory lock here because it degrades to "someone else runs it in five minutes" instead of "nobody ever runs it again".

## Done when

- No advisory lock in `pruneRetention`.
- A killed worker mid-pass does not prevent the next pass from acquiring — worth an explicit test, since this is the exact case the current code gets wrong.
- Contention is observable rather than silent (see the sibling issue on `partitionsDropped: 0` being ambiguous).

DhananjayThomble:
## What does this PR do?

Replaces the session-scoped advisory lock guarding `pruneRetention` with a lease that expires by the clock, so a worker that dies mid-pass can no longer stop retention from ever running again.

Fixes #323

## Why

`pruneRetention` guarded its pass with `pg_try_advisory_lock`. Session-scoped rather than transaction-scoped was the right instinct — the drops deliberately span several transactions so the `ACCESS EXCLUSIVE` lock a `DETACH` takes on the parent is released between them, and a `_xact_` variant would release at the first commit.

The problem is that a session advisory lock lives on the **connection**, and nothing bounds how long it is held. The `finally` covers a thrown error but not the process disappearing. `postgres.js` pools connections and the AWS profile runs `DATABASE_POOL_MAX=1`, so a worker frozen or reclaimed mid-pass — the same hazard that motivated awaiting click writes in #277 — strands the lock on a backend Postgres still considers healthy. Every later pass then takes the early return.

Retention stops permanently, and **silently**: no error, nothing in the logs, and `partitionsDropped: 0` reads identically to "nothing to drop". The first symptom would be a disk graph, or RDS going read-only at the 50 GB autoscale ceiling.

## Why not just remove the guard

That was my first instinct and it is wrong. #283 added `MAX_RETAINED_DAYS`, a volume cap with its **own** drop loop targeting partitions `click_events_spent_partitions` never lists. With two independent drop loops in one function, per-drop idempotence (`click_events_drop_partition` catching `undefined_table`) is not sufficient — the guard has to be at pass level.

So the guard stays, and stops being able to outlive its holder.

## How it works

A `job_leases` row carries `locked_until`. Acquiring is a single conditional upsert, releasing sets it to `now()`, and a crash needs no cleanup because the row is already timestamped to expire. The worst case degrades from *"nobody runs it ever again"* to *"somebody runs it a few minutes late"*.

Two details that are easy to get wrong, both covered by tests:

**Acquire is one statement.** `insert ... on conflict do update ... where locked_until < now()` lets Postgres serialise contenders on the row, and the loser's update is filtered out rather than overwriting the winner's claim. A read-then-write would let both callers come away believing they hold it.

**Release matches on `holder`.** A pass that overruns its TTL no longer owns the lease. Releasing unconditionally would hand the new holder's lease away and let a third pass start alongside it — manufacturing exactly the concurrent execution the lease exists to prevent.

The table is deliberately generic rather than retention-specific: any job that must not overlap and cannot fit in one transaction has the same problem. For work that *does* fit in one transaction, `pg_try_advisory_xact_lock` is still the better tool and the code says so.

## On the TTL

300 seconds, matching the worker Lambda's own timeout — the real ceiling on how long a pass can be alive at all. A pass cannot legitimately run long anyway: `MAX_PARTITION_DROPS` drops across the two loops, each bounded by a 2s `lock_timeout`, so even a pass where every drop times out finishes well inside the lease.

## Type of change

## How should this be tested?

No new environment variables. `DATABASE_URL` as usual; the lease suite skips without it, matching the other worker suites.

```
pnpm db:up && pnpm db:migrate   # 0012_job_leases
```

Migration `0012_job_leases` is additive — a new table, no changes to existing ones — so it applies cleanly to a populated database.

Ten new tests in `apps/worker/src/jobs/lease.test.ts`, including the regression this PR exists for:

- **a holder that dies without releasing does not stop the next pass** — the exact case the advisory lock got wrong, forced by expiring the row directly since nothing else produces that state on demand
- a second holder is refused while the lease is live, and the job does not run at all
- the lease is released on the way out, so sequential passes do not block each other
- a job that throws still releases, and the error is not swallowed
- an overrunning pass does not release a lease someone else has taken
- leases for different job names are independent
- **mutual exclusion under three racing callers**

That last one asserts *peak concurrency*, not "exactly one acquired". Worth calling out: my first version asserted the latter and failed, because two passes acquiring in **sequence** is correct — the first releases on its way out. Counting acquisitions proves nothing; only overlap does.

Results: **495 tests pass** across all packages. The worker suite was run three times consecutively to confirm the concurrency paths are stable.

## Notes for review

- `RETENTION_LEASE_SECONDS` sits next to `MAX_PARTITION_DROPS` and `MAX_RETAINED_DAYS` so the three bounds that shape a pass are read together.
- Not acquiring the lease returns `{ partitionsDropped: 0, rowsDeleted: 0 }` rather than throwing — it is the mechanism working, not a failure. Note this keeps the ambiguity #324 describes, where zeroes cannot be distinguished from lock contention; that is deliberately left to #324 rather than half-solved here.
- `HOLDER` includes the Lambda function name, pid and a random suffix. Correctness only needs it unique per holder; the readable prefix is for reading the table later.

## Checklist

`docs/DECISIONS.md` should record why cross-transaction jobs use a lease while single-transaction ones use an advisory lock, but that belongs in #290, which is already correcting several other claims in one pass.

### DIFF
--- a/apps/worker/src/jobs/lease.test.ts
+++ b/apps/worker/src/jobs/lease.test.ts
@@ -39,7 +39,10 @@ describeDb("withLease", () => {
   }
 
   beforeEach(async () => {
-    handle ??= createDatabase({ url: DATABASE_URL!, max: 2 });
+    // Room for several genuinely concurrent callers: the overlap tests need
+    // more than one connection in flight or they serialise on the pool and
+    // stop testing overlap at all.
+    handle ??= createDatabase({ url: DATABASE_URL!, max: 4 });
     db = handle.db;
     await db.execute(sql`delete from job_leases where name like 'test_lease%'`);
   });
@@ -132,6 +135,78 @@ describeDb("withLease", () => {
     // otherwise lock itself out permanently.
     const next = await withLease(db, NAME, 60, async () => "after failure");
     expect(next.acquired).toBe(true);
+  });
+
+  it("derives the expiry from ttlSeconds, so the lease frees itself on time", async () => {
+    /* Every other test writes `locked_until` by hand, which means none of them
+       touch the TTL — reading the argument as minutes instead of seconds, or
+       ignoring it entirely, would pass the whole suite. This is the only case
+       that pins the unit. */
+    const held = await withLease(db, NAME, 1, async () => "held");
+    expect(held.acquired).toBe(true);
+
+    // Immediately after release the lease is free, so re-take it and let this
+    // one lapse on the clock rather than by releasing.
+    await db.execute(sql`
+      update job_leases set locked_until = now() + make_interval(secs => 1), holder = 'ttl-holder'
+      where name = ${NAME}
+    `);
+
+    const tooSoon = await withLease(db, NAME, 60, async () => "should not run");
+    expect(tooSoon.acquired).toBe(false);
+
+    await new Promise((resolve) => setTimeout(resolve, 1200));
+
+    const afterExpiry = await withLease(db, NAME, 60, async () => "expired naturally");
+    expect(afterExpiry.acquired).toBe(true);
+    expect(afterExpiry.value).toBe("expired naturally");
+  });
+
+  it("does not release a lease taken over by another pass in the same process", async () => {
+    /* The sibling of the test below, and the one that actually bites, because
+     * the guard cannot see it: both passes live in one process.
+     *
+     * The release matches on a holder token. Were that token per *process*
+     * rather than per *acquisition*, two overlapping passes inside one worker
+     * would share the string — so the overrunning pass's release would match the
+     * takeover's row and free a lease that is still in use, letting a third pass
+     * run alongside it.
+     *
+     * The takeover has to still be **in flight** when the outer pass exits. An
+     * earlier version of this test awaited the takeover inside the outer job; it
+     * released itself first, leaving nothing live to stomp, and the test passed
+     * against the per-process bug it was written to catch. Verified by mutation:
+     * pin the token to a constant and this fails. */
+    let takeoverInFlight = false;
+    let takeoverAcquired = false;
+    let takeoverPromise: Promise<unknown> = Promise.resolve();
+
+    await withLease(db, NAME, 60, async () => {
+      // This pass overruns: its lease lapses while it is still working.
+      await expireLease();
+
+      // Deliberately not awaited — it must outlive the outer pass.
+      takeoverPromise = withLease(db, NAME, 60, async () => {
+        takeoverInFlight = true;
+        await new Promise((resolve) => setTimeout(resolve, 400));
+        takeoverInFlight = false;
+      }).then((r) => {
+        takeoverAcquired = r.acquired;
+      });
+
+      // Long enough for the takeover to have acquired and entered its job.
+      await new Promise((resolve) => setTimeout(resolve, 80));
+    });
+
+    // The outer pass's release has now run. The takeover is still working.
+    const intruder = await withLease(db, NAME, 60, async () => "third pass");
+
+    await takeoverPromise;
+
+    expect(takeoverAcquired).toBe(true);
+    // The assertion: nobody got in while the takeover held the lease.
+    expect(intruder.acquired && takeoverInFlight).toBe(false);
+    expect(intruder.acquired).toBe(false);
   });
 
   it("does not release a lease that has already been taken by someone else", async () => {
--- a/apps/worker/src/jobs/lease.test.ts
+++ b/apps/worker/src/jobs/lease.test.ts
@@ -1,0 +1,197 @@
+import { afterAll, beforeEach, describe, expect, it } from "vitest";
+import { createDatabase, sql, type Database } from "@snapurl/database";
+import { withLease } from "./lease.js";
+
+/* ============================================================
+   Job leases, against a real Postgres.
+
+   The property that matters is the one a session-scoped advisory
+   lock did not have: a holder that vanishes must not keep the
+   job from ever running again. That is why the expiry case has
+   its own test — it is the regression this file exists for.
+
+   Runs only when DATABASE_URL is set, matching the other worker
+   suites.
+   ============================================================ */
+
+const DATABASE_URL = process.env.DATABASE_URL;
+const describeDb = DATABASE_URL ? describe : describe.skip;
+
+describeDb("withLease", () => {
+  let handle: ReturnType<typeof createDatabase>;
+  let db: Database;
+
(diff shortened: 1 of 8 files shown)


DhananjayThomble:
## Review response — f062cd0

Design review came back APPROVED with 7 non-blocking findings. Three were defects in the code this PR adds, so they are fixed here rather than deferred. Two are pre-existing and filed separately.

### The holder token was per-process, not per-acquisition

A real defect, and the comment beside it asserted a property the code did not have.

Release matches on a holder token so an overrunning pass cannot free a lease somebody else has since taken. That token was computed **once per process**, which makes the guard blind to the case where the somebody else is *another pass in the same process*: the two share the string, so the overrunning pass's release matches the takeover's row, frees a lease still in use, and lets a third pass start alongside it. Precisely the concurrent execution the guard exists to prevent.

Not reachable through `pruneRetention` today — `everyN` carries a `running` flag and Lambda serialises invocations per environment — but `withLease` is exported as a general-purpose helper and documented the property unconditionally. Now generated per acquisition.

Release also nulls `holder`, so a clean exit and a dead holder are no longer identical in the table.

### The TTL was never tested

Every test wrote `locked_until` by hand, so **nothing asserted the expiry derives from `ttlSeconds` at all**. Reading the argument as minutes instead of seconds, or ignoring it entirely, would have passed all nine tests. There is now a case that acquires with a 1-second TTL, confirms a second caller is refused, waits, and confirms the lease frees itself on the clock.

### My first test for the holder bug was worthless

Worth calling out, because it is the more useful lesson.

I wrote a same-process takeover test, and it **passed against the bug it was written to catch**. It awaited the takeover inside the outer job, so the takeover released itself before the outer pass exited — there was no live lease left to stomp.

Rewritten so the takeover is still *in flight* when the outer pass returns, then verified by mutation: pinning the token to a constant fails the test, restoring per-acquisition passes it.

```
WITH BUG (holder pinned to a constant)   -> 1 failed | 10 passed
WITH FIX                                 -> 80 passed, 3 consecutive runs
```

A test that passes with and without the fix proves nothing, and the only way to know which you have is to try it. I'd rather do that check on every regression test than trust that a new assertion is load-bearing.

### The TTL comment overstated its own argument

It bounded a pass by the drop loops' 2s `lock_timeout`. Those *are* bounded, at roughly 40s worst case. The row-level `DELETE` the pass ends on is not — it carries neither `lock_timeout` nor `statement_timeout`, and it does real work exactly when workspaces use mixed retention settings.

The honest bound is the Lambda's five-minute ceiling on AWS, and the comment now says so, along with the fact that the compose and Helm profiles have no process ceiling and a long pass there can be overtaken.

### Not fixed here

**#326** — the volume-cap loop's `sql.raw` partition probe can raise `undefined_table` if a concurrent pass drops that partition between the listing and the probe, and nothing catches it. The line is pre-existing from #283, but this PR makes it *reachable*: the old advisory lock never expired, so two passes could never overlap. Allowing a lapsed lease to be taken over is the intended trade, and it turns that probe into a live path. Filed with the fix (a `to_regclass` guard, plus isolating `pruneRetention` in `runMaintenance` the way provisioning already is).

**#324** — already open, and the sibling problem: `click_events_drop_partition` returns `false` for lock timeout, deadlock and already-dropped alike, so a pass losing every drop is indistinguishable from one with nothing to do. Deliberately not half-solved here.

One more thing the review surfaced that is worth recording: the original bug was **broader than this PR's description**. Both advisory-lock calls were session-scoped and `createDatabase` defaults to `max: 10`, so on any profile other than AWS the unlock could land on a different pooled connection than the lock and strand it — no crash required at all. Removing connection affinity is the deeper win here, not just surviving a frozen Lambda.

**497 tests pass.** Worker suite run three times to confirm the concurrency paths are stable.

PR #327, opened 2026-08-30
Stop a vanished partition throwing out of the retention pass

### LINKED ISSUE #326: Concurrent retention passes can make the volume-cap partition probe throw, skipping the rest of maintenance
Found reviewing #325. The offending line is pre-existing from #283, but #325 makes it reachable, so it is worth fixing now rather than later.

## Problem

The volume-cap loop in `pruneRetention` lists partition names in one statement and then probes each in another:

```ts
const [{ pending }] = (await db.execute(
  sql.raw(`select exists (select 1 from only "${part}" where rolled_up_at is null) as pending`),
)) as unknown as [{ pending: boolean }];
```

If a concurrent pass drops `part` between the listing and the probe, this raises `undefined_table` and nothing catches it. The deliberate idempotence lives inside `click_events_drop_partition`, not here.

The throw escapes `pruneRetention`, and `runMaintenance` only wraps `ensureClickPartitions` in a `try` — so `pruneOutbox`, `pruneDeliveries` and `stuckProjections` are all skipped for that pass.

## Why it is newly reachable

Before #325, `pruneRetention` was guarded by a session-scoped advisory lock that never expired. That was the bug #323 fixed, and it had a side effect: two passes could never overlap. The failure mode was "nobody runs it", not "two run at once".

The lease deliberately allows a pass to be overtaken once its TTL lapses. That is the intended trade — it is what stops a dead holder blocking retention forever — but it turns this probe into a live path.

Both preconditions have to line up (more than `MAX_RETAINED_DAYS` partitions attached, plus an overrun), so probability is low. The blast radius is one skipped maintenance tail with a logged error, which is far better than the silent permanent stop it replaced. Still worth closing.

## Fix

- Guard the probe with `to_regclass`, or catch the `undefined_table` class and treat it as "already gone" the way `click_events_drop_partition` does.
- Wrap `pruneRetention` in `runMaintenance` the way `ensureClickPartitions` already is, so an unexpected throw cannot take the remaining jobs with it. That is worth doing independently of this bug — it is the same reasoning that isolated provisioning.

## Related

`#324` covers the sibling problem that contention is unobservable: `click_events_drop_partition` returns `false` for lock timeout, deadlock and already-dropped alike, so a pass losing every drop is indistinguishable from one with nothing to do.

DhananjayThomble:
## What does this PR do?

Stops a partition vanishing mid-pass from throwing out of `pruneRetention`, and isolates every maintenance job so one failing can no longer silently skip the rest.

Fixes #326

## Why

The volume-cap loop read partition names in one statement and probed each in the next:

```ts
sql.raw(`select exists (select 1 from only "${part}" where rolled_up_at is null) as pending`)
```

A concurrent pass dropping that partition in between made the probe raise `undefined_table`, and nothing caught it — unlike `click_events_drop_partition`, which has always treated a missing partition as somebody else having got there first. The throw escaped `pruneRetention` and took the rest of the maintenance pass with it.

**This only became reachable with #323.** Retention used to be guarded by a session-scoped advisory lock that never expired, so two passes could not overlap — the failure mode was "nobody runs it", not "two run at once". Replacing it with an expiring lease deliberately allows a lapsed pass to be overtaken, which is what turned this probe into a live path. Fixing the previous bug exposed this one.

## How

The pending check moves inside `click_events_drop_partition` (migration `0013`), and the raw probe is deleted.

A `to_regclass` guard was the obvious fix and it is not enough: the check and the drop would still be separate statements, so it would only narrow the window rather than close it. Inside the function the check runs **after the partition is locked**, so a row cannot arrive between deciding and dropping either.

**Behaviour is unchanged.** Returning `false` when rows are pending is exactly what both callers already did with a pending partition — skip it, leave it attached, don't count it as dropped. Telling "pending" apart from "lock timeout" and "already gone" is #324's job, and deliberately not half-done here.

## Second half: isolating the maintenance jobs

The issue also asked for `pruneRetention` to be isolated the way `ensureClickPartitions` already was. Rather than add a second bespoke `try`/`catch`, every job in the pass now runs through a small `step()` helper.

These jobs are independent — expiring salts has nothing to do with pruning webhook deliveries — so one throwing should cost that job, not the pass. Previously an unexpected error anywhere after provisioning skipped everything below it, including the privacy jobs, and skipped them **silently**: the pass never reached its own log line, so there was nothing to see.

Failures are collected and reported together, and the log message changes to `"maintenance pass completed with failures"` with a `failedJobs` array. A pass that half-worked now says so instead of looking like a pass with nothing to do.

## Also in here

A flake I shipped in #323. The TTL test used a one-second lease, and the window between claiming it and asserting a second caller is refused contains real database round trips — on a loaded machine the lease had already lapsed and the assertion failed. It failed on this branch before I touched anything related to it. Now three seconds, which is honest about what it measures at the cost of a couple of seconds' runtime.

I'd rather fix a known flake in the PR that surfaced it than leave it to be rediscovered as noise.

## Type of change

## How should this be tested?

No new environment variables. Migration `0013` replaces a function body — no schema change, so it applies cleanly to a populated database and is safe to re-run.

```
pnpm db:up && pnpm db:migrate
```

Two tests added against the drop function directly, now that it owns the check:

- **a partition that is already gone reports `false` rather than raising** — the #326 regression, covering the new `LOCK` and probe that run before the `DETACH`
- **one holding un-rolled-up clicks is refused, then becomes droppable once counted** — the second half matters: it proves the refusal was about the pending row rather than something incidental to that partition

The existing cap tests already asserted the pending invariant end to end, and they still pass unchanged, which is the useful signal that moving the check into SQL preserved behaviour.

Results: **499 tests pass** across all 8 packages. Worker suite run four times consecutively to confirm the timing fix holds.

## Notes for review

- `0013` is a `CREATE OR REPLACE FUNCTION` with no schema change, so `drizzle-kit generate` produces nothing and the journal entry and snapshot are hand-added. The snapshot is `0012`'s with a fresh `id` and its `prevId` pointing at `0012`, keeping drizzle's chain intact — verified all 27 tables carried over.
- `LOCK TABLE ... IN ACCESS EXCLUSIVE MODE` before the probe costs nothing extra: `DETACH` needs that lock on the parent regardless, and it is inside the existing bounded `lock_timeout` with the `undefined_table` handler now covering it too.
- `step()` returns a typed fallback per job rather than `undefined`, so the log line and the return shape are unchanged for callers.

## Checklist

The maintenance-pass log shape changed (`failedJobs`, and a different message on partial failure). That belongs with the operations notes in #290, which is already rewriting them.


DhananjayThomble:
## Review response — c5a8ca6

Design review came back NEEDS_CHANGES with one blocking finding. It was right, and it was mine.

### I introduced a new lock-order inversion

Moving the pending check inside `click_events_drop_partition` meant locking the partition so the answer could not change between deciding and dropping. But **`LOCK TABLE` on a partition takes no lock on the parent** — so the function went child→parent while every other route to a partition goes parent→child. The rollup's batch claim, its marker `UPDATE`, and retention's own row-level `DELETE` all take `RowShareLock`/`RowExclusiveLock` on `click_events` *and* on every partition, and `RowShareLock` conflicts with `ACCESS EXCLUSIVE` in both directions. A genuine cycle, verified empirically by the reviewer.

The nastier part is **which process dies**. The scanner starts waiting one step earlier, so its `deadlock_timeout` fires first and it takes the `40P01`. The visible symptom would have been a failed *rollup*, not a failed drop — and the `deadlock_detected` handler I'd written into the function would never have seen it. `rollupClicks` isn't wrapped by `step()`, so on the Lambda path that fails the frequent invocation.

Before my change the function's first request was the parent, so while it waited it held nothing anyone else needed: the same collision resolved as a bounded 2s wait, `lock_not_available`, and a `false` return. **I turned a self-reporting no-op on the low-priority job into an uncaught abort on the higher-priority one.**

Fixed by taking `LOCK TABLE ONLY "click_events" IN ACCESS EXCLUSIVE MODE` first. `ONLY` matters — without it the lock descends to every partition, which would block inserts routing to *today* while an ancient day is dropped. Holding the parent exclusively is also what actually makes the probe trustworthy, since all routing goes through the parent.

My migration comment said taking the partition lock "costs nothing extra, because DETACH needs the parent anyway". Wrong premise: no extra *strength*, but a changed *order*. Corrected along with the code.

### The regression test didn't test the regression

`"reports false instead of raising when the partition is already gone"` passes against `main` too — the old handler caught `undefined_table` from `DETACH` exactly as the new one catches it from `LOCK`. The bug was never in the function. It was in the **caller**, where the cap loop listed partition names in one statement and probed them in the next.

So there's now a test at that level. Reproducing it needs the world to change *between two of `pruneRetention`'s own statements*, which no fixture can arrange — so the database handle is proxied, and the moment the listing query returns, one of the partitions it just named is dropped out from under the loop. That's the exact interleaving a concurrent pass produces, made deterministic.

Verified by mutation:

```
undefined_table handler removed  ->  2 failed  (incl. the new caller-level test)
handler restored                 ->  83 passed
```

The old test is kept for the function's own tolerance, with its comment corrected to say plainly that it does not cover #326.

Second time in three PRs that a regression test of mine passed against the bug. The mutation check is the only thing that catches it, and it's cheap.

### `step()` made the invocation unable to fail

Swallowing every job error meant a pass where all seven failed — dead connection, wrong credentials — returned a tidy object of zeroes and reported success. Same failure shape #323 and #326 were both about: a pass that did nothing, indistinguishable from a pass with nothing to do.

It also meant the EventBridge rule's `retryAttempts: 2` had nothing to retry, so a transient outage waited an hour instead of getting two more chances inside the minute. Now throws when *every* job failed, handing the decision back to the scheduler. Per-job isolation is kept, which was the point.

### Also

Removed a stale comment block that still described provisioning-only isolation and was duplicated thirty lines below.

**500 tests pass**, all 7 CI checks green.

### Findings not addressed here

- **`0013_snapshot.json` is minified** where its eleven siblings are pretty-printed. Cosmetic — drizzle parses it either way and rewrites it on the next generate. It is the tell that it was hand-made, which it was, because a function-only migration produces no diff for `drizzle-kit` to work from.
- **A pending partition now briefly holds `ACCESS EXCLUSIVE`** where it previously read at `AccessShare` — and a partition with `rolled_up_at IS NULL` rows is exactly what the rollup wants. No cycle, since that path returns before requesting anything else, and it is bounded by the probe's duration. Worth knowing rather than fixing.
- **Nothing tests `step()` itself.** Asserting that a throwing job leaves the later ones running needs injection the worker doesn't have today. Noted rather than bodged.

PR #328, opened 2026-08-30
Report why a partition was not dropped, and order the maintenance locks consistently

### LINKED ISSUE #324: Partition maintenance: lock-order inversion, silent contention, and an over-long default lock hold

DhananjayThomble:
## What does this PR do?

Makes a retention pass able to say what it actually did, and gives the two partition-maintenance functions a consistent lock order.

Fixes #324

## Why

Three separate problems, all variations on the same theme.

### Contention was indistinguishable from having nothing to do

`click_events_drop_partition` returned `boolean`, and `false` covered four different outcomes: the partition still held un-rolled-up clicks, the lock could not be taken before the timeout, it deadlocked, or it was already gone. `pruneRetention` could only count successes.

So a pass losing **every** drop to write traffic reported `partitionsDropped: 0` — byte-identical to a pass with nothing to drop. Retention could stall indefinitely with no signal. That is the same silent-failure shape as #323 (a stranded lock stopping retention for ever) and #326 (a throw skipping the rest of the pass): the pass does nothing and looks exactly like a pass that had nothing to do.

It now returns `'dropped' | 'pinned' | 'contended' | 'missing'`, and only `contended` means anything needs attention — so that one gets its own `warn` line. `pinned` and `missing` are benign but reported when non-zero, so they cannot hide inside an idle-looking pass.

### The two maintenance functions locked in opposite orders

Since #326, the drop path goes parent → partition → default (`DETACH` needs the default too). Provisioning went the other way: it locked the default to make its drain safe, and only touched the parent when `ATTACH` did. Opposite orders on the same two objects is a deadlock cycle — and two overlapping maintenance passes stopped being exotic the moment a lapsed lease could be taken over.

Provisioning now takes the parent first, in **`SHARE UPDATE EXCLUSIVE`**. That is the mode `ATTACH` already needs, and it does not conflict with the `ROW EXCLUSIVE` an insert takes, so ordering this correctly costs the redirect path nothing. `ACCESS EXCLUSIVE` would also have fixed the ordering *and* blocked every insert for the duration — trading one problem for a worse one. Consistent order is what removes the cycle; matching strength is not required.

### The default's lock was held across the whole provisioning window

`ensureClickPartitions` ran the entire window as one `generate_series` statement with a lateral call. So the `ACCESS EXCLUSIVE` lock taken on the DEFAULT partition for the *first* day needing work was held until that whole statement committed — every day's attach, plus every day's drain.

Measured in the #325 review: a concurrent insert routed to the default blocked for **7.9s** while the `ATTACH` itself took 7ms. Only clicks routed to the default are affected, which is exactly the catch-up case where provisioning has fallen behind and the drain is large.

Now one statement per day, so the lock is released between days. Cost is 22 cheap catalogue lookups per hourly pass instead of one round trip.

## Type of change

## How should this be tested?

No new environment variables. Migration `0014` replaces two function bodies and changes one return type — no schema change, so it applies cleanly to a populated database.

```
pnpm db:up && pnpm db:migrate
```

Four tests added, and how each is driven matters:

- **contention is reported separately** — forced by holding `ACCESS EXCLUSIVE` on the parent from a second connection so the function's 2s timeout expires. That is the only way to observe `contended` deterministically; waiting for natural contention would be flaky and probably never fire.
- **pinned partitions are counted** — driven through the **cap** loop, not the age loop. Not incidental: `click_events_spent_partitions` filters pending partitions out before the age pass ever sees them, so it structurally cannot report a pinned day. The filter is worth keeping, because offering a candidate certain to be refused means taking `ACCESS EXCLUSIVE` on the parent for nothing.
- **per-day provisioning** — asserted by counting `execute` calls through a proxied handle (1 window query + 5 days), not by timing, which would be flaky.
- **`missing` and `dropped`** — the existing direct-function tests, updated for the new return type.

Verified by mutation: collapsing `contended` back into an undifferentiated outcome fails three tests.

```
contended collapsed  ->  3 failed | 17 passed
restored             ->  86 passed (worker), 503 total
```

Results: **503 tests pass** across all 8 packages.

## Notes for review

- I also fixed an existing assertion that compared `ensureClickPartitions`' result with `toBe`. Fine when it returned a number; reference equality now that it returns `{ ready, declined }`. Caught by the suite, not by inspection.
- `RetentionResult` gains three fields rather than replacing `partitionsDropped`, so existing log parsing and the `lambda.ts` passthrough are unaffected.
- The `warn` for contention is deliberately not an `error`: one contended pass is normal under load and the next will very likely succeed. It is the pattern over time that matters, which is what a metric filter on that line is for.
- `LOCK TABLE ONLY` in both functions — without `ONLY`, `LOCK TABLE` on a partitioned table descends to every partition, which would be a far heavier lock than either needs.

## Checklist

The maintenance log gained several fields across this PR and #326. Documenting what an operator should watch belongs with the operations notes in #290, which is rewriting them anyway.


DhananjayThomble:
## Review response — 59f8406

Review came back APPROVED with 9 findings, none blocking. Three fixed here, because two of them meant the counter this PR adds couldn't see the thing it was added for.

### The timeout was set one statement too late

`CREATE TABLE ... (LIKE "click_events" ...)` takes `ACCESS SHARE` on the parent — and that is provisioning's **first** parent lock, one statement earlier than the explicit `LOCK TABLE` this PR adds. It sat *above* `SET LOCAL lock_timeout`, so it waited without bound. A pass racing a drop that holds the parent exclusively blocked for as long as the drop took, then went on to succeed and report the day as provisioned. Reviewer measured **18.3s against a 20s holder, returning `ready`**.

Since `declined` exists precisely to make "provisioning is losing to write traffic" visible, that was the one collision it couldn't see. Default-partition contention was counted correctly; parent contention wasn't. The timeout now covers both.

### My header rationale was too strong

I wrote that the reorder closes a deadlock cycle between provisioning and dropping. **The cycle was already unreachable.** That same `LIKE "click_events"` holds `ACCESS SHARE` on the parent to end of transaction, so provisioning was parent-first already — implicitly, at a weaker mode. A drop could never take parent `ACCESS EXCLUSIVE` while a provisioner held `ACCESS SHARE`, so neither side could get one step into the other's ordering. Confirmed: 30k mixed ensure/drop/insert/rollup transactions, **zero deadlocks on this revision and on the one before it.**

The reorder is still worth having — the ordering is explicit rather than an accident of which statement happens to touch the parent first, and it takes the parent at the mode `ATTACH` actually needs, so nothing upgrades to a stronger lock mid-function, which is the shape that *would* reintroduce a cycle. The comment now says that, instead of claiming a fix it didn't make.

That's the third time in this run a comment of mine asserted more than the code did. Worth naming as a pattern rather than a one-off.

### Declined days now warn, like contended ones

`contended` got a dedicated `warn`; `declined` got a conditional field on an info line that otherwise reads as success. The argument for one applies to the other: a declined day has no partition, so its clicks land in the DEFAULT partition — which no query prunes and retention skips by design. One declined day is nothing; a provisioner that never wins quietly accumulates rows somewhere they'll sit for ever.

### Contention test made deterministic

It started a blocking transaction and slept 300ms, which had to cover a lazy TCP connect, auth, `BEGIN` and `LOCK TABLE`. Now polls `pg_locks` for the blocker's granted lock. Losing that race would have produced a red assertion rather than a false green, but a test that depends on winning a race is a test that will eventually fail for the wrong reason.

**503 tests pass**, worker suite run three times, all CI checks green.

### Filed rather than fixed

**A DEFAULT-routed insert can fail outright** when a concurrent `ATTACH` commits: `23514 new row for relation "click_events_default" violates partition constraint`, measured at **93/15,164 inserts** under load — and the same rate on `origin/main`, so not a regression here. But since the redirect path awaits its click writes, these are **lost clicks, not delayed ones**, and it qualifies my "ordering this correctly costs the write path nothing": true of blocking, not of failure. Filed separately.

### Acknowledged, not addressed

- **The cap loop can attempt every attached partition.** Both break conditions key on `outcomes.dropped`, so an all-pinned or all-contended pass tries every attached day. Pre-existing, and only reachable when `attached > MAX_RETAINED_DAYS` (default 1100). Breaking on *attempts* is the fix; it belongs with the cap logic rather than here.
- **`declined` and the lock ordering are untested.** Nothing drives `ensure` to return `NULL`, and nothing asserts the parent lock precedes the default's. The second needs `pg_locks` introspection mid-function, which is awkward to do honestly.
- **The already-gone test still passes against `origin/main`.** Only the literal changed, `false` → `'missing'`. Same gap #327's review flagged; kept because it does cover the function's own tolerance, just not the regression.
- **The pinned test assumes it owns the oldest partition.** Real order dependence; `fileParallelism: false` contains it for now.

PR #330, opened 2026-08-30
fix(clicks): retry a click insert once when a concurrent ATTACH invalidates the route

### LINKED ISSUE #329: A click routed to the DEFAULT partition can be lost when a concurrent ATTACH commits
Found while reviewing #328. **Pre-existing and not a regression** — measured at the same rate before and after that change — but it is the only path found so far that loses a click outright rather than delaying it.

## Problem

A click routed to the DEFAULT partition can fail with:

```
23514  new row for relation "click_events_default" violates partition constraint
```

when a concurrent `ATTACH PARTITION` for that click's day commits in between the router choosing the default and the partition constraint being evaluated against a freshly invalidated relcache.

Measured under load, 10 concurrent clients mixing provisioning, drops, inserts and rollup-shaped scans:

| | Failed inserts |
| --- | --- |
| `fix/retention-observability` | 93 of 15,164 (~0.6%) |
| `origin/main` | 124 of 16,101 (~0.8%) |

Same rate, so the ordering work in #324 neither caused nor worsened it.

## Why it matters more than it looks

Since #277 the redirect path **awaits** its click write. So this is not a delayed insert that lands late — it is an error on the write path, and the click is gone. `record()` catches and logs it, so a visitor still gets their redirect and nothing appears broken; the count is simply short.

It is also the exact case the DEFAULT partition exists to prevent. The whole argument for keeping a default is "an insert must never fail for want of a partition" — and this is an insert failing *because* a partition arrived.

## Why it is rare, and when it is not

Two things have to coincide: the click has to route to the DEFAULT partition, and an `ATTACH` for that same day has to commit in the same instant.

In steady state neither happens much — days are provisioned about a fortnight ahead, so clicks land in a real partition and the default stays empty. The window opens exactly when provisioning has fallen behind and is catching up, which is also when the default is receiving traffic. So the failure clusters precisely where it does the most damage.

## Direction

The tension is that #324 deliberately took `SHARE UPDATE EXCLUSIVE` on the parent during `ATTACH` rather than `ACCESS EXCLUSIVE`, so that provisioning does not block the redirect path. `ACCESS EXCLUSIVE` would close this window by serialising the insert against the attach — and would stall every insert for the duration of every attach, which is the trade #324 explicitly rejected.

Options worth weighing:

- **Retry the insert once on `23514`.** Cheap, local to `PostgresClickSink`, and correct: on the second attempt the day's partition exists and the row routes to it. The failure is genuinely transient, which is what makes a retry the honest fix rather than a patch over a race.
- **Retry inside the sink for any transient partition-routing error**, so the same guard covers a future variant.
- **Accept and measure.** Log it distinctly enough to count, and decide once there is a real rate from a real deployment rather than from a synthetic stress test.

I would take the first: one bounded retry on that specific SQLSTATE, with a test that forces the interleaving.

## Related

- #277 made click writes awaited, which is what turns this from a lost background write into a visible failed one.
- #324 chose the lock mode that leaves this window open, deliberately and for good reasons.

DhananjayThomble:
Closes #329

## The problem

`click_events` is partitioned by day, with a DEFAULT partition as the safety net for days that have not been provisioned yet. The worker attaches new day-partitions while the redirect path is inserting.

A row routed to the default can fail its own partition constraint, because that constraint is "none of the attached ranges" and the set of attached ranges just changed:

1. The tuple router looks at the catalogue, sees no partition for the row's day, picks `click_events_default`, and asks for `ROW EXCLUSIVE` on it.
2. `ATTACH` holds `ACCESS EXCLUSIVE` on the default while it drains, so the insert queues behind it.
3. The attach commits. The insert is released, processes the invalidation, and evaluates the default's constraint — which no longer covers that day. It cannot re-route, because its partition descriptor was pinned before the invalidation arrived.

```
SQLSTATE 23514: new row for relation "click_events_default" violates partition constraint
```

Postgres rechecks the default's constraint during routing precisely for this reason (commit `ef1e125`, backpatched to 12).

#329 measured **93 of 15164 inserts (~0.6%)** during a loaded provisioning pass. Same rate on `main`, so it is not a regression from a recent change.

It matters more than the rate suggests:

- Since #277 made click writes awaited, this is a **failed** write rather than a delayed one. On the Postgres sink the caller catches, logs, and the click is silently gone from analytics.
- It clusters exactly when provisioning has fallen behind, because that is when the default is receiving traffic at all. The failures arrive in a burst, on the busiest window, when the system is already degraded.
- It is the precise scenario the DEFAULT partition exists to prevent. The whole argument for keeping a default is that an insert must never fail for want of a partition — and here one fails *because* a partition arrived.

## The change

`insertClickEvents` retries once on that specific failure, and reports that it did.

```ts
try {
  await insertOnce(db, events);
} catch (err) {
  if (!isTransientPartitionRoutingError(err)) throw err;
  options?.onRetry?.(err);
  await insertOnce(db, events);
}
```

**In the shared insert, not in `PostgresClickSink`.** The worker's SQS consumer drains the queue back into the same table through the same function, so on the AWS profile the sink is not even on the path. Both writers need the retry; there is one place that is both of them.

**One retry, not a loop with backoff.** The failing statement had already pinned its partition descriptor, so it could not re-route. A fresh statement builds a fresh one and lands the row in the partition that now exists. A row that fails the same way twice is failing for some other reason and should surface rather than be retried into silence.

**Matched on `routine`, with the message as a fallback.** `23514` is `check_violation` — what every CHECK constraint raises. All four producers that reach this schema were provoked against Postgres 18.6:

| routine | message | retry |
| --- | --- | --- |
| `ExecPartitionCheckEmitError` | new row for relation "click_events_default" violates partition constraint | yes |
| `ExecFindPartition` | no partition of relation "x" found for row | yes |
| `ExecConstraints` | new row for relation "x" violates check constraint "y" | no |
| `ATRewriteTable` | updated partition constraint for default partition "x" would be violated by some row | no |

That last row is the interesting one: an `ATTACH` refused because the default already holds a row for the incoming range is a **permanent** failure, and its message mentions a partition constraint too. Matching on `partition constraint` would have swept it in.

`routine` is checked first because `errmsg()` output is translated through `lc_messages` and `routine` is not — a non-English server would otherwise stop retrying with no failing test and no log line, which is the same silence this fix exists to end. The message is kept as a fallback, OR'd rather than routine being authoritative, so a future Postgres renaming the function degrades to a text match instead of quietly never retrying.

**The retry is not invisible.** `onRetry` is optional and both writers log it at warn — the redirect through the Fastify logger, the worker through the shared pino instance. The case for this change is a rate, so there has to be something to count. Steady retries mean provisioning has stopped running and the DEFAULT partition is absorbing live traffic, which is the actual problem to go fix.

## What this deliberately does not do

It does not promote `ATTACH` to `ACCESS EXCLUSIVE` on the parent. That would close the window by serialising every insert against the attach, not just the ones routed to the default. #324 chose `SHARE UPDATE EXCLUSIVE` specifically so provisioning cannot stall the redirect path, and that trade still holds: a 0.6% retry costs one extra round trip on those rows, where the lock promotion costs latency on all of them.

## Two preconditions, documented because they are easy to break

1. **`db` must be a pool handle, not a transaction.** Inside an explicit transaction the 23514 aborts it and the retry fails with `25P02`, which this does not retry — the fix would quietly become a no-op. The signature asking for `Database` rather than the exported `Executor` union is what enforces it.
2. **Resending the batch is safe only because Drizzle emits one multi-row `INSERT ... VALUES (...), (...)`,** which is all-or-nothing in autocommit. There is no second line of defence if that changes: `id` is a server-side per-row `uuidv7()`, so a duplicated click would carry a fresh key, be indistinguishable from a real one, and be folded into the rollups.

## Refactor included

The Drizzle cause-chain walk (`postgresErrorCode`, `isUniqueViolation`, `isForeignKeyViolation`) moved from `apps/api/src/common/postgres-error.filter.ts` into a new `packages/database/src/postgres-errors.ts`, joined by `isTransientPartitionRoutingError`.

The detector belongs next to the statement it guards, and the walk it needs was in the wrong package. The filter re-exports all three names, so its existing importers (`links.service.ts`, `forms.service.ts`, `api.test.ts`) are untouched. Copying the ten-line walk instead would leave two versions to keep in step — and the bug it exists to prevent, a SQLSTATE sitting one level down being missed, is exactly the kind that drift reintroduces. It already turned a handled 409 into a 500 once.

## Testing

**24 tests added.**

Unit (`packages/database`, 12 + 10): the control flow around the insert and the branches of the detector, with a fake `db` that records the rows of every attempt. Retry-then-succeed; gives up after exactly two attempts; batch cardinality in and out; the retry resends the whole batch unchanged; `onRetry` fires once and only on a retry; a plain check violation is not retried; the `ATTACH`-refused wording is not retried; a non-English message still matches via `routine`; a missing `routine` still matches via the message; the walk finds the code several levels down and terminates on a circular cause chain; the surfaced error still carries the SQLSTATE.

Integration (`apps/worker/src/jobs/partitions.test.ts`, 2, live Postgres): the unit tests use fixtures the detector's author also wrote, so they cannot contradict it. These can.

- All four 23514s provoked against the real server, asserting the detector separates them. The routine names and message strings in the source are now checked against Postgres rather than against recollection.
- **The actual race, staged deterministically.** Hold the default's `ACCESS EXCLUSIVE` lock from a second connection, let an insert for an unprovisioned day queue behind it, attach that day, commit. The insert is released into exactly the window above. Asserts the retry fired *and* that the click ended up in the partition that arrived — not lost, not stranded in the default.

They live in `partitions.test.ts` rather than a new file so they serialise against the other partition-machinery tests instead of racing them for the same locks from a parallel vitest worker.

**Four mutations verified:**

| mutation | caught by |
| --- | --- |
| retry removed | 4 unit tests + the live race test |
| message needle widened to `"partition"` | 1 unit test + the live 23514 test |
| batch truncated to `events.slice(0, 1)` | 2 unit tests |
| `routine` match only, message fallback removed | 2 unit tests |

The "gives up after one retry" test asserts an exact attempt count of 2, so a change to the retry limit fails it by construction.

- `pnpm type-check` clean
- `pnpm test` — 527 passed, up from 503

## Reviewer's shortcut

The retry only fires for rows headed to the DEFAULT partition, so the honest way to make it unreachable is to keep provisioning ahead of traffic. This is a safety net under the safety net, and the `onRetry` warn is there to say when it is carrying more weight than it should.

### DIFF
--- a/apps/api/src/common/postgres-error.filter.ts
+++ b/apps/api/src/common/postgres-error.filter.ts
@@ -19,10 +19,10 @@ import { isForeignKeyViolation, isUniqueViolation, postgresErrorCode } from "@sn
 
 /* Re-exported, not redefined.
  *
- * These moved to @snapurl/database because the redirect and worker now need the
- * same cause-chain walk to recognise a transient partition-routing failure on the
- * click write. Two copies of it would be two things to keep in step, and the bug
- * this walk exists to prevent — a code sitting one level down being missed — is
+ * These moved to @snapurl/database because the shared click INSERT lives there
+ * and needs the same cause-chain walk to recognise a transient partition-routing
+ * failure. Two copies of it would be two things to keep in step, and the bug this
+ * walk exists to prevent — a code sitting one level down being missed — is
  * exactly the kind that drift reintroduces. Re-exporting keeps every existing
  * importer in this app working unchanged. */
 export { postgresErrorCode, isUniqueViolation, isForeignKeyViolation };
--- a/apps/api/src/common/postgres-error.filter.ts
+++ b/apps/api/src/common/postgres-error.filter.ts
@@ -1,5 +1,6 @@
 import { ArgumentsHost, Catch, ExceptionFilter, HttpException, Logger } from "@nestjs/common";
 import type { FastifyReply } from "fastify";
+import { isForeignKeyViolation, isUniqueViolation, postgresErrorCode } from "@snapurl/database";
 
 /* ============================================================
    A backstop so a database constraint can never reach a user as
@@ -16,24 +17,15 @@ import type { FastifyReply } from "fastify";
    message that does not leak schema details.
    ============================================================ */
 
-/** Walks the cause chain — the code can be several levels down. */
-export function postgresErrorCode(err: unknown): string | null {
-  let current: unknown = err;
-  for (let depth = 0; depth < 5 && current; depth++) {
-    const code = (current as { code?: unknown }).code;
-    if (typeof code === "string" && /^\d{5}$/.test(code)) return code;
-    current = (current as { cause?: unknown }).cause;
-  }
-  return null;
-}
-
-export function isUniqueViolation(err: unknown): boolean {
-  return postgresErrorCode(err) === "23505";
-}
-
-export function isForeignKeyViolation(err: unknown): boolean {
-  return postgresErrorCode(err) === "23503";
-}
+/* Re-exported, not redefined.
+ *
+ * These moved to @snapurl/database because the redirect and worker now need the
+ * same cause-chain walk to recognise a transient partition-routing failure on the
+ * click write. Two copies of it would be two things to keep in step, and the bug
+ * this walk exists to prevent — a code sitting one level down being missed — is
+ * exactly the kind that drift reintroduces. Re-exporting keeps every existing
+ * importer in this app working unchanged. */
+export { postgresErrorCode, isUniqueViolation, isForeignKeyViolation };
 
 const STATUS_FOR: Record<string, { status: number; message: string }> = {
   "23505": { status: 409, message: "That already exists." },
--- a/apps/redirect/src/click-sink.ts
+++ b/apps/redirect/src/click-sink.ts
@@ -27,10 +27,18 @@ export interface ClickSink {
 }
 
 export class PostgresClickSink implements ClickSink {
-  constructor(private readonly db: Database) {}
+  /** `onRetry` surfaces a click that had to be re-inserted because a concurrent
+   *  partition ATTACH invalidated its route (#329). Optional so a test can build
+   *  the sink with a bare handle; main.ts wires it to the app logger, because a
+   *  retry firing constantly means partition provisioning has stopped and the
+   *  DEFAULT partition is absorbing live traffic. */
+  constructor(
+    private readonly db: Database,
+    private readonly onRetry?: (err: unknown) => void,
+  ) {}
 
   async record(event: ClickEvent): Promise<void> {
-    await insertClickEvents(this.db, [event]);
+    await insertClickEvents(this.db, [event], { onRetry: this.onRetry });
   }
 }
 
--- a/apps/redirect/src/main.ts
+++ b/apps/redirect/src/main.ts
@@ -224,7 +224,13 @@ async function init(): Promise<void> {
       CLICK_QUEUE_URL,
     );
   } else {
-    clicks = new PostgresClickSink(database!.db);
+    /* Warn rather than error: a retry means the click was *saved*, so nothing is
+       broken. It is the rate that carries information — steady retries mean
+       partition provisioning has fallen behind and the DEFAULT partition is
+       taking live traffic, which is what a metric filter on this line is for. */
+    clicks = new PostgresClickSink(database!.db, (err) =>
+      app.log.warn({ err }, "click insert lost its partition route to a concurrent attach — retried"),
+    );
   }
 
   /* The salt source: the daily_salts table when a Postgres handle exists,
--- a/apps/worker/src/jobs/partitions.test.ts
+++ b/apps/worker/src/jobs/partitions.test.ts
@@ -1,5 +1,18 @@
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
-import { clickEvents, createDatabase, domains, eq, links, sql, workspaces, type Database } from "@snapurl/database";
+import {
+  clickEvents,
+  createDatabase,
+  domains,
+  eq,
+  insertClickEvents,
+  isTransientPartitionRoutingError,
+  links,
+  postgresErrorCode,
+  sql,
+  workspaces,
+  type ClickEvent,
+  type Database,
+} from "@snapurl/database";
 import { ensureClickPartitions, pruneRetention } from "./rollup.js";
 
 /* ============================================================
@@ -29,6 +42,17 @@ const DATABASE_URL = process.env.DATABASE_URL;
 const describeDb = DATABASE_URL ? describe : describe.skip;
 
 const PARENT = "public.click_events";
+
+/** Poll until a condition holds, rather than sleeping and hoping. Used for lock
+ *  state, where a fixed head start has to cover a lazy TCP connect plus auth and
+ *  losing that race makes a test assert the wrong thing. */
+async function waitUntil(what: string, holds: () => Promise<boolean>, tries = 200): Promise<void> {
+  for (let i = 0; i < tries; i++) {
+    if (await holds()) return;
+    await new Promise((resolve) => setTimeout(resolve, 50));
+  }
+  throw new Error(`timed out waiting for ${what}`);
+}
 
 describeDb("click_events partitioning", () => {
   let handle: ReturnType<typeof createDatabase>;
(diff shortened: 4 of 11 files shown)


DhananjayThomble:
## Review response

All 11 findings addressed in 05262f7. The two that changed the fix rather than the prose:

**The message needle matched a permanent failure.** I provoked all four 23514s this schema can raise against Postgres 18.6 rather than reasoning about them, and `ATTACH` refused because the default already holds a row for the incoming range comes back as:

```
routine: ATRewriteTable
message: updated partition constraint for default partition "click_events_default" would be violated by some row
```

`includes("partition constraint")` matched that. It is permanent — the fix is to drain the default, which is what `click_events_ensure_partition` does — so retrying it can only fail again. Needle tightened to `violates partition constraint`, and there is now a negative test carrying the real routine and wording.

**Matching translated text at all was the deeper problem.** As flagged: `errmsg()` goes through `lc_messages`, so a non-English server would stop retrying with no failing test and no log line — the same silence the fix exists to end. The check is now `23514` AND (`routine` is `ExecPartitionCheckEmitError` or `ExecFindPartition`, OR the message names a routing failure). Routine first because it is not translated; message kept as a fallback and OR'd rather than routine being authoritative, so a future Postgres renaming the function degrades to a text match instead of silently never retrying. Removing the fallback fails 2 tests.

### The rest

| # | finding | resolution |
| --- | --- | --- |
| 1 | "measured at roughly 0.6%" not derivable from anything committed | Attributed to #329 rather than stated as a property of this repo. |
| 2 | "plans against fresh catalogue state" is the wrong mechanism | Corrected in both files to the partition descriptor rebuilt from an invalidated relcache. Right conclusion, wrong pointer for whoever debugs this next. |
| 3 | "three different apps now need it" contradicted by the diff | Corrected. No app imports the detector; the shared insert does, and it lives in this package. Still the right move, different reason. |
| 4 | needle could widen with no test noticing | The `ATRewriteTable` negative case above. Mutation-verified: widening to `includes("partition")` now fails a unit test **and** the live 23514 test. |
| 5 | message match is locale-dependent, fails open to no-retry | Routine-first, as above. |
| 6 | batch contract entirely unasserted | Two tests: N in → N out in order, and the retry resends all N unchanged. `events.slice(0, 1)` now fails both. |
| 7 | atomicity is the only thing preventing double-counted clicks, undocumented | Stated in the JSDoc, including that `id` is a per-row server-side `uuidv7()` so a duplicate would be undetectable. |
| 8 | retry requires autocommit, nothing says so | Stated as an explicit precondition, with why `Database` rather than the exported `Executor` union is the signature. |
| 9 | "the click is simply gone" only true on one profile | Scoped to the Postgres sink; the AWS path's `batchItemFailures` redrive is noted as costing a redrive, not the click. |
| 10 | no integration test despite an existing harness | Two, in `partitions.test.ts` — see below. |
| 11 | no signal the retry ever fires | `onRetry` on the insert, wired to a warn in both writers. The bogus `describe("insertClickEvents retry logging")` and its stray `vi.restoreAllMocks()` are gone. |

Also corrected: the test comment calling `no partition of relation` transient because "a provisioning pass may have just created the day". As you note, it only fires with no DEFAULT partition, which 0007 makes permanent. It is still retried — the cost of being wrong is one round trip — but the comment now says why rather than inventing a mechanism.

### On the integration tests

Two, and they went in `apps/worker/src/jobs/partitions.test.ts` rather than a new file: a separate file runs in a parallel vitest worker and would race the existing tests for the same partition locks. Inside that file they serialise.

The first provokes each of the four 23514s and asserts the detector separates them, so the routine names and message strings in the source are checked against the server rather than against my recollection of it.

The second is the race itself, staged so it is deterministic instead of a load test. Your point about the window was the useful one — it is not "an insert ran concurrently with an attach", since `ATTACH` already holds `ACCESS EXCLUSIVE` on the default and serialises those. It is an insert *released* from that lock at the worst possible moment. So the test does exactly that: hold the default's lock from a second connection, poll `pg_locks` until an insert for an unprovisioned day is queued behind it (ungranted `RowExclusiveLock`), attach that day, commit. The insert wakes into precisely the #329 window. It asserts the retry fired and that the click ended up in the partition that arrived — not lost, not stranded in the default. Removing the retry fails it.

### One suggestion not taken

Linking the first error as `cause` when the second attempt fails differently. `onRetry` now receives the first error and both writers log it, so it is no longer discarded — which was the actual gap. Rewrapping the second error to carry the first seemed like it would make the thing a caller catches harder to reason about for no extra information.

### Verification

- `pnpm type-check` clean, `pnpm test` 527 passed (was 503).
- `@snapurl/database` 65 (was 58), worker 88 (was 86).
- Four mutations verified: retry removed (4 unit + 1 live), needle widened (1 unit + 1 live), batch truncated (2 unit), message fallback removed (2 unit). The retry-limit mutation is caught by construction — that test asserts an exact attempt count of 2.

Issue #352, opened 2026-09-06
[test-infra] Add accessible names to ambiguous controls so E2E selectors can be stable


DhananjayThomble:
## Problem

The frontend has **zero `data-testid` attributes**, and several pages render generic repeated labels ("Save", "Cancel", "Delete") across multiple cards, so those button names are not unique on the page.

This blocks reliable Playwright selectors. It also cannot be solved later by an automated test-writing agent, because adding attributes to components is a `src/` change outside the test-only allowlist.

## Fix

Add accessible names rather than test ids — this makes selectors unambiguous **and** is a genuine accessibility improvement:
- `aria-label` / `aria-labelledby` on icon-only and generically-labelled controls
- `role="region"` + `aria-labelledby` on repeated card sections so locators can be scoped

Existing good examples already in the codebase: the search input, the create-link drawer's close button, and the routing-rule move buttons all carry proper `aria-label`s. Follow that pattern.

Do **not** add `data-testid`.

## Acceptance criteria

- [ ] Every interactive control on the dashboard routes is reachable by a unique accessible name within its section
- [ ] No visual change
- [ ] No `data-testid` introduced

### Constraints for an automated agent
- Do **not** modify `.github/**`, `.kiro/**`, `package.json`, lockfiles, or any `*.config.*` unless this issue explicitly says to.
- Never weaken, skip, or delete an existing assertion to make a build pass.
- Import request/response shapes from `@snapurl/contract`; do not hand-roll payload literals.
- DB-backed tests use the existing `const describeDb = DATABASE_URL ? describe : describe.skip` idiom.
- If the task cannot be completed without touching a forbidden path, stop and comment on this issue instead of widening scope.

PR #370, opened 2026-09-09
test(web): set up vitest + first lib/utils tests (#351)

### LINKED ISSUE #351: [test-infra] Set up vitest for web/ (currently zero tests)
## Problem

`web/` has **no test files, no vitest config, and no `test` script**. The root `pnpm -r test` therefore skips the entire frontend silently — "all tests pass" currently says nothing about the dashboard.

## Fix

- `web/vitest.config.ts` with the `happy-dom` environment. **`happy-dom` is already in the lockfile** via `apps/extension`, so check before adding a dependency.
- `"test": "vitest run"` in `web/package.json`.
- Add `@testing-library/react` for component tests.
- One representative smoke test proving the harness works.

Match the conventions already used by the other packages' `vitest.config.ts` files: colocated `*.test.ts(x)`, `src/**` include glob.

## Acceptance criteria

- [ ] `pnpm --filter snapurl-web test` runs and passes
- [ ] The root `pnpm test` now includes web
- [ ] `pnpm type-check` and `pnpm build` still pass

## Note

This issue is explicitly allowed to touch `web/package.json` and add a config file. Later test-only issues are not.

DhananjayThomble:
## Set up vitest for web/ + first tests (closes #351)

`web/` had **zero tests** and no test tooling. This bootstraps it and adds a real first suite.

### Changes
- `web/vitest.config.ts` — node environment, `src/**/*.test.ts` (matches the monorepo's convention; DOM/component tests can add jsdom later without changing this).
- `web/package.json` — `test` (`vitest run`) + `test:watch` scripts, and `vitest ^3.2.4` devDep (matched to the rest of the monorepo). Lockfile updated.
- `web/src/lib/utils.test.ts` — **11 tests** covering the pure formatting helpers the dashboard renders every number/money/link with: `compact`, `inr` (lakh/crore), `pct`, `shortUrl`, `faviconFor` (incl. malformed-URL fallback), and `relativeDate`/`formatDate` (with faked time + the unparseable-input passthrough).

### Why this is enough to close it
The CI `verify` job runs `pnpm -r test`, which now recurses into `snapurl-web` automatically — verified locally (`utils.test.ts` runs under the root recursive command). So web tests now run on every PR, and the file gives a real, non-trivial baseline to grow (not a placeholder).

Local: web suite 11/11 pass; `pnpm install --frozen-lockfile` clean.

*QA program — phase 2 test-infra. First of the web-testing chain (#352/#353 build on this).*

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -410,6 +410,9 @@ importers:
       typescript:
         specifier: ^5.7.3
         version: 5.9.3
+      vitest:
+        specifier: ^3.2.4
+        version: 3.2.7(@types/node@22.20.1)(happy-dom@15.11.7)(jiti@2.7.0)(lightningcss@1.32.0)(supports-color@8.1.1)(terser@5.51.0)(tsx@4.23.12)(yaml@2.9.0)
 
 packages:
 
--- a/web/package.json
+++ b/web/package.json
@@ -9,7 +9,9 @@
     "vercel-build": "pnpm --filter @snapurl/contract build && pnpm --filter @snapurl/domain build && next build",
     "start": "next start --port 3000",
     "lint": "eslint .",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "test": "vitest run",
+    "test:watch": "vitest"
   },
   "dependencies": {
     "@hookform/resolvers": "^5.2.2",
@@ -37,6 +39,7 @@
     "@types/react": "^19",
     "@types/react-dom": "^19",
     "tailwindcss": "^4.1.13",
-    "typescript": "^5.7.3"
+    "typescript": "^5.7.3",
+    "vitest": "^3.2.4"
   }
 }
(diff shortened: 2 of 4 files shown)


DhananjayThomble (commented):

**Correctness:** the 11 assertions match utils.ts behavior, including edge cases (negative magnitudes, malformed-URL fallback in faviconFor, unparseable-date passthrough, faked time for relativeDate). Real coverage, not a placeholder.

**Scope/config:** additive only — vitest config (node env, `src/**/*.test.ts`, matching the monorepo convention), a `test` script so `pnpm -r test` recurses into web, and the vitest devDep pinned to the repo's ^3.2.4. No app code touched; lockfile committed so `--frozen-lockfile` holds.

**Non-blocking observation (NOT changed here — out of #351's scope):** in `src/lib/utils.ts`, `pct()` has a no-op ternary `n > 0 ? "" : ""` (both branches empty), so it never prefixes a sign. If a leading `+` on positive deltas was intended, that's a separate one-line source fix worth its own change; left untouched to keep this PR strictly test-infra.

CI gate green; heavy jobs correctly skipped. Approving.

PR #372, opened 2026-09-09
test(e2e): Playwright harness against fixtures mode + first journey (#353)

### LINKED ISSUE #353: [test-infra] Playwright E2E harness running against fixtures mode
## Problem

There is no browser E2E testing of any kind. Nothing verifies what a user actually does.

## Fix

Add an `e2e/` workspace package with Playwright, targeting the frontend's **fixtures mode** for per-PR runs.

Key facts that make this cheap:
- `NEXT_PUBLIC_USE_FIXTURES=true` activates a stateful in-memory fake backend in `web/src/lib/api/fixtures.ts`, so no database or API is needed.
- `next.config.ts` refuses a production build with fixtures on unless `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true` is also set. CI must set both.
- Auth is plain `localStorage` (`snapurl.accessToken` / `snapurl.refreshToken`) with client-side route protection, so a test can seed a session via `page.addInitScript` instead of driving the login form every time.
- Build and serve with `next start`, **not** `next dev` — dev's lazy per-route compilation is a flake source.

Fixture state is module-level: it survives client-side navigation but resets on a full page load. Encode that in the helpers.

## Scope for this issue

Harness + helpers + one journey (register or login through to creating a link). Further journeys are separate issues.

## Acceptance criteria

- [ ] `e2e/` added to `pnpm-workspace.yaml` with a `playwright.config.ts`
- [ ] Session-seeding and fixture-reset helpers in `e2e/support/`
- [ ] One passing end-to-end journey
- [ ] Chromium only; browser download cached by version
- [ ] A new CI job runs it on PRs

DhananjayThomble:
## Playwright E2E harness against fixtures mode (closes #353)

First browser E2E in the repo. A new `e2e/` workspace package drives the dashboard in **fixtures mode** (`NEXT_PUBLIC_USE_FIXTURES=true`) — no API, no database — so it's cheap enough to run per-PR.

### What's added
- **`e2e/` workspace package** (added to `pnpm-workspace.yaml`): `@playwright/test@1.63.0`, Chromium only.
- **`e2e/playwright.config.ts`** — `webServer` builds web with `NEXT_PUBLIC_USE_FIXTURES=true` + `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true` (both required by `next.config.ts`) and serves it with `next start` (not `next dev` — dev's lazy per-route compile is a documented flake source). Chromium project, `trace: on-first-retry`, 1 retry + 1 worker in CI.
- **`e2e/support/session.ts`** — `seedSession` (`addInitScript` sets the two `localStorage` token keys the fixture backend hands out, so the client-side route guard sees us logged in without driving the login form) and `resetFixtures` (documents the reality: fixture state is module-level and a full page load *is* the reset — there's no exported reset fn).
- **`e2e/tests/create-link.spec.ts`** — one journey: seeded session → `/links` → open the create drawer → fill destination + a unique back-half → submit → assert the new link appears. Selectors are **accessible-name only** (`getByRole`/`getByLabel`), which #352 made unique and stable — the new link is located by its copy button's `aria-label` `Copy short link <domain>/<slug>`.
- **CI job `Browser E2E (fixtures)`** in `verify.yml`: gated on `changes.code` (skips docs-only PRs), caches the Playwright browser keyed on version, installs Chromium, runs the suite, uploads the HTML report on failure. Added to the **`CI gate`** `needs` list so the gate cannot pass while E2E is red or pending.

*(This issue explicitly authorizes editing `pnpm-workspace.yaml`, adding a `playwright.config.ts`, and adding a CI job.)*

### Verified locally
Built web under fixtures, launched Chromium, ran the journey headless — **1/1 passed** (repeatably). `pnpm --filter snapurl-e2e type-check` clean; lockfile regenerated + committed (`--frozen-lockfile` holds); YAML validated.

### Watch on first CI run
Playwright-in-Actions can't be fully validated in a local sandbox. Things to watch in the CI log: (1) the browser cache key / first-run Chromium download, (2) the `next build` under fixtures on the runner, (3) `webServer` boot timing (180s budget). If any flakes, the fix is scoped to the config/CI job, not the app.

*QA program — phase 2 test-infra. Built on #352's accessible names.*

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -753,6 +753,72 @@ jobs:
           echo "migration bookkeeping recovered: $restored_migrations rows"
           echo "backup/restore round-trip verified"
 
+  # Browser end-to-end tests (issue #353). Drives the dashboard in fixtures mode
+  # (NEXT_PUBLIC_USE_FIXTURES=true) — no API or database — so it is cheap enough
+  # to run per-PR. Chromium only; the browser binary is cached by Playwright
+  # version so most runs skip the download. Gated on `code` like the other jobs
+  # so a docs-only PR skips it (CI gate then treats the skip as acceptable).
+  e2e:
+    name: Browser E2E (fixtures)
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v4
+        with:
+          run_install: false
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+
+      - name: Install
+        run: pnpm install --frozen-lockfile
+
+      # Build the workspace packages web depends on (@snapurl/contract, domain).
+      - name: Build packages
+        run: pnpm build:packages
+
+      # Cache the Playwright browser download, keyed on the installed version so a
+      # version bump busts the cache and everything else reuses it.
+      - name: Resolve Playwright version
+        id: pw
+        run: echo "version=$(node -p "require('@playwright/test/package.json').version")" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Playwright browsers
+        id: pw-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ steps.pw.outputs.version }}
+
+      - name: Install Chromium
+        if: steps.pw-cache.outputs.cache-hit != 'true'
+        run: pnpm --filter snapurl-e2e exec playwright install --with-deps chromium
+
+      # On a cache hit the browser binary is present but its OS deps may not be.
+      - name: Install Chromium system deps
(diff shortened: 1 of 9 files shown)


DhananjayThomble (commented):

**CI evidence:** the new `Browser E2E (fixtures)` job passed in Actions on the first run — the three risk areas (Playwright browser cache/download, `next build` under fixtures, `webServer` boot within the 180s budget) all held. CI gate green; heavy jobs correctly skipped/passed.

**Design:** fixtures mode (no API/DB) keeps it cheap per-PR; `next start` not `next dev` avoids lazy-compile flake; session seeded via the two real localStorage token keys so the route guard passes without driving login each time; the one journey asserts on #352's accessible names (the copy button's `Copy short link <domain>/<slug>`), which is stable and unique.

**Scope:** the issue explicitly authorizes `pnpm-workspace.yaml`, `playwright.config.ts`, and a CI job — all added. The e2e job is in the `CI gate` `needs` list, so a future E2E failure can't slip past the gate. No existing assertion weakened; lockfile regenerated + committed.

**Verified locally too:** built web under fixtures, ran Chromium headless, 1/1 pass (repeatably); e2e type-check clean. Approving.

PR #373, opened 2026-09-09
test(web): fixture-parity test — fail when fixtures drift from the API (#354)

### LINKED ISSUE #354: [test-infra] fixture-parity test: fail the build when frontend fixtures drift from the API
## Problem

`web/src/lib/api/fixtures.ts` is a hand-maintained fake backend. When a hook is added without a matching fixture branch, the UI **appears to work and is entirely fake**. Nothing currently detects this.

The repo already has the two halves needed to catch it automatically: `buildOpenApiDocument()` in `apps/api/src/openapi/document.ts` enumerates all registered routes, and the fixtures file dispatches on path + method.

## Fix

A plain vitest test that builds the OpenAPI document and asserts every registered path + method has a corresponding branch in the fixtures router. Runs in milliseconds inside the existing `pnpm test` — no new CI job.

Where the mismatch is intentional (an endpoint no UI calls), the test should require an explicit, commented allowlist entry rather than silently passing.

## Acceptance criteria

- [ ] Test fails if a route is added to the API with no fixture branch
- [ ] Test fails if a fixture branch exists for a route that no longer exists
- [ ] Intentional gaps require an explicit allowlist entry
- [ ] Runs inside the existing test suite

DhananjayThomble:
## Fixture-parity test: fail the build when web fixtures drift from the API (closes #354)

`web/src/lib/api/fixtures.ts` is a hand-maintained fake backend. When a route is added to the API with no fixture branch, the UI appears to work and is entirely fake. This adds a test that fails on drift in **both** directions, running inside the existing `pnpm test` — no new CI job.

### How the two halves connect (without breaking the boundary)
The route list lives in `apps/api` and the fixtures live in `web`, and the steering rules forbid `web` importing from `apps/*`. So the API's route surface crosses the boundary as **serialized data, not a source import**:
- `apps/api/scripts/emit-openapi.mjs` calls the existing static `buildOpenApiDocument()` and writes the `{path → methods}` map to a committed `web/src/lib/api/openapi.generated.json` (41 paths / 56 method+path pairs). Convenience: `pnpm --filter @snapurl/api emit:openapi`.
- The web test reads that JSON as data. The web bundle still depends only on `@snapurl/contract` and `@snapurl/domain`.

### The test — `web/src/lib/api/fixture-parity.test.ts`
- **Direction 1 (route with no branch):** for each live API route, probe the **real** `fixtureRequest` dispatcher with a concrete path; the dispatcher's `No fixture for <METHOD> <path>` sentinel means no branch. Exact, and derived from the actual dispatcher (not source-scraping).
- **Direction 2 (stale branch for a dead route):** `fixtures.ts` now exports `FIXTURE_ROUTE_PATTERNS` (the recognizer `{methods, pattern}` list); the test asserts each declared branch still matches a live API route. A third block re-probes each declared pattern against the real dispatcher so the list can't drift from the if/else chain.
- **Allowlist (`fixture-parity.allowlist.ts`):** intentional API-only routes (`/auth/refresh`, `/links/export`, `/forms/{id}/responses.csv`, `/health`) each require a non-empty `reason` — the test rejects a reason-less entry, so a gap can't be waved through silently. Suppresses both directions.

### Verified locally
- 157 parity assertions pass; full web suite 168, api suite 172, all green.
- **Proved it catches drift:** injecting a fake API route fails direction 1; the sentinel match is tightened so a branch's own not-found error (`No fixture form …`) isn't misread as a missing branch.
- No new dependency (lockfile unchanged); no `.github` change; no existing assertion weakened.

### Judgment calls (flagging per the issue's constraints)
- Added an `emit:openapi` script to `apps/api/package.json` (a convenience wrapper for the raw node command). Minor, non-behavioral; happy to drop it if you'd rather keep `package.json` untouched.
- The staleness risk (committed JSON going out of date) is mitigated by CI building the API before `pnpm test`; a hard freshness gate could be a follow-up.

*QA program — phase 2 test-infra.*

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -10,6 +10,7 @@
     "start": "node --env-file-if-exists=.env dist/main.js",
     "type-check": "tsc -p tsconfig.json --noEmit",
     "test": "vitest run",
+    "emit:openapi": "pnpm build && node scripts/emit-openapi.mjs",
     "dev:nest": "nest start --watch"
   },
   "dependencies": {
--- a/apps/api/scripts/emit-openapi.mjs
+++ b/apps/api/scripts/emit-openapi.mjs
@@ -1,0 +1,33 @@
+/*
+ * Emit the API's route surface (OpenAPI paths) as a committed JSON artifact that
+ * web's fixture-parity test consumes as DATA (issue #354). The route list crosses
+ * the web<->apps boundary as serialized JSON, never as a source import, so the web
+ * bundle still depends only on @snapurl/contract and @snapurl/domain.
+ *
+ * Runs on the COMPILED output (dist/), so it needs `pnpm --filter @snapurl/api build`
+ * first and then plain node — no extra TS runner dependency. buildOpenApiDocument()
+ * is static (no running server): it reads the registry populated by the path imports.
+ *
+ * Usage:  pnpm --filter @snapurl/api build && node apps/api/scripts/emit-openapi.mjs
+ */
+import { writeFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, resolve } from "node:path";
+import { buildOpenApiDocument } from "../dist/openapi/document.js";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const OUT = resolve(here, "../../../web/src/lib/api/openapi.generated.json");
+
+const doc = buildOpenApiDocument("api");
+
+const paths = {};
+for (const [p, ops] of Object.entries(doc.paths ?? {})) {
+  paths[p] = Object.keys(ops)
+    .filter((k) => ["get", "post", "put", "patch", "delete"].includes(k.toLowerCase()))
+    .map((m) => m.toUpperCase())
+    .sort();
+}
+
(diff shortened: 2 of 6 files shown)


DhananjayThomble (commented):

**CI:** CI gate green — the verify job (which now runs the 157 parity assertions inside `pnpm test`) passed, alongside E2E, integration and dynamo-smoke. No new CI job, as the issue required.

**Boundary:** the API route surface crosses web↔apps as committed JSON (`openapi.generated.json`, emitted by `apps/api/scripts/emit-openapi.mjs`), never a source import — web still depends only on contract/domain. Steering invariant preserved.

**Both directions covered:** dir-1 probes the real `fixtureRequest` dispatcher per live route (exact — not source-scraping); dir-2 checks each declared branch still maps to a live route; a third probe keeps the declared list honest against the if/else chain. Allowlist entries require a non-empty reason.

**Correctness proof:** I injected a fake API route locally and confirmed dir-1 fails; and tightened the sentinel match so a branch's own not-found error (`No fixture form …`) isn't misread as a missing branch. web 168 + api 172 green.

**Flagged judgment call:** added an `emit:openapi` convenience script to `apps/api/package.json` (non-behavioral wrapper). Happy to drop if you'd rather keep package.json untouched. Approving.

Issue #355, opened 2026-09-06
[test-infra] Add coverage tooling and a coverage-delta check


DhananjayThomble:
## Problem

There is no coverage measurement, so there is no automated way to tell a test that asserts real behaviour from one that asserts nothing. This matters specifically because agent-authored tests will be merged with limited human review.

## Fix

- Add `@vitest/coverage-v8` and enable coverage reporting.
- Publish `coverage-summary.json` as a CI **artifact** from the nightly run on `main`.
- Add a PR job that downloads the baseline artifact and fails when a test-only PR increases covered lines in **no** file.

Store the baseline as an artifact, not a committed file: a committed baseline would itself be a config file that test-only PRs are forbidden to update, which would deadlock the lane.

## Acceptance criteria

- [ ] `pnpm test` can emit coverage
- [ ] Nightly run publishes the baseline artifact
- [ ] A PR adding a tautological test fails the delta check
- [ ] A PR adding a real test passes it
- [ ] PRs touching only `e2e/**` are exempt (no vitest coverage there)

### Constraints for an automated agent
- Do **not** modify `.github/**`, `.kiro/**`, `package.json`, lockfiles, or any `*.config.*` unless this issue explicitly says to.
- Never weaken, skip, or delete an existing assertion to make a build pass.
- Import request/response shapes from `@snapurl/contract`; do not hand-roll payload literals.
- DB-backed tests use the existing `const describeDb = DATABASE_URL ? describe : describe.skip` idiom.
- If the task cannot be completed without touching a forbidden path, stop and comment on this issue instead of widening scope.

PR #375, opened 2026-09-09
test: fast-check property tests for the contract schemas (#356)

### LINKED ISSUE #356: [test-infra] Add fast-check and property tests for the contract schemas
## Problem

`packages/contract` is the single source of truth for every request and response shape, and it encodes genuinely subtle rules — but it is covered by only two test files.

The highest-value target is `HttpUrl` in `packages/contract/src/http-url.ts`, which refines a URL schema to reject private, loopback, and link-local hosts. **This is SSRF-adjacent**: it is the control preventing a user-supplied destination from pointing at internal infrastructure or cloud metadata.

## Fix

Add `fast-check` and property tests covering:
- `HttpUrl`: generators biased toward the denylist boundaries (each private IPv4 range and its edges, CGNAT, link-local/metadata, IPv6 loopback/ULA/link-local, IPv4-mapped IPv6, alternate IP encodings, trailing dots, userinfo tricks). Assert rejection.
- Tri-state fields: for each nullable/optional field, that `undefined`, `null`, and a value are each handled as intended. These three states have materially different service semantics and hand-written tests routinely miss one.
- Round-trip stability on response schemas.
- Discriminated unions: no input parses as two variants.

Note that slug **shape** is in the zod schema but reserved words, length, and extension rules live in `packages/domain/src/slug.ts` — a generator built only from the schema will produce slugs the contract accepts and the service still rejects. Test both layers.

## Acceptance criteria

- [ ] `fast-check` added as a dev dependency
- [ ] Property tests for the four areas above
- [ ] Tests are deterministic in CI (seeded)
- [ ] Any genuine bug found is reported as a separate issue, not silently accommodated

DhananjayThomble:
## fast-check property tests for the contract schemas (closes #356)

Adds `fast-check` and property tests over the four areas the issue calls out, biased toward the boundaries hand-written tests miss. Seeded for deterministic CI.

### Coverage
**`packages/contract/src/schemas.property.test.ts`** (10 properties):
- **`HttpUrl` — the SSRF guard** (highest-value target): generators biased at each denylist boundary — every private/loopback/link-local/CGNAT IPv4 range and its edges, IPv6 loopback/link-local/ULA, IPv4-mapped-IPv6 metadata (`::ffff:169.254.169.254`), `localhost`/`*.localhost` case-folding — all asserted **rejected**. Non-http(s) schemes (`javascript:`/`data:`/`file:`/`ftp:`) rejected. Public hosts accepted (no false positives). **No SSRF gap found** under 500 runs — the guard holds.
- **Tri-state fields**: `expiresAt`/`clickLimit` (`.nullable().optional()` on `CreateLinkInput`) preserve `undefined` / `null` / value as distinct. Documented an intentional asymmetry surfaced while writing this: `CreateLinkInput.comment` is optional-**only** while `Link.comment` is nullable — not a bug, but exactly the kind of subtlety property tests exist to pin.
- **Discriminated union**: `BulkLinkOutcome` parses as exactly one variant keyed on `ok`; a wrong-shape-for-its-discriminant object is rejected.
- **Round-trip**: `Link` re-parses stably.

**`packages/domain/src/slug.property.test.ts`** (4 properties): the slug **rules** live in the domain layer, not the contract (the schema pins only the shape). This test lives in `@snapurl/domain` deliberately — **contract must not import domain (a cycle)**, so the "test both layers" note is honored on the domain side. Reserved words, file-extension slugs, and over-length slugs are all shape-valid but rule-rejected; a clean slug is accepted.

### Notes
- `fast-check@^4.9.0` added as a devDep to both `packages/contract` and `packages/domain`; lockfile committed (`--frozen-lockfile` holds).
- Tests are **deterministic** (fixed `seed`).
- **No genuine bug found**, so nothing to file separately (per the acceptance criterion). The only "failure" while developing was a test asserting the wrong contract for `comment`, corrected.
- One generator initially **hung** (a `fc.string().filter(shapeRegex)` starved because random strings rarely match the slug shape) — rewritten to generate directly from the slug alphabet; both files run in <100ms.

*QA program — phase 2 test-infra. The last actionable phase-2 issue; #267 epic stays open.*

### DIFF
--- a/packages/contract/package.json
+++ b/packages/contract/package.json
@@ -6,9 +6,14 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "exports": {
-    ".": { "types": "./dist/index.d.ts", "default": "./dist/index.js" }
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
   },
-  "files": ["dist"],
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "tsc -p tsconfig.json",
     "dev": "tsc -p tsconfig.json --watch --preserveWatchOutput",
@@ -21,6 +26,7 @@
   "devDependencies": {
     "@types/node": "^22.20.1",
     "typescript": "^5.7.3",
-    "vitest": "^3.2.4"
+    "vitest": "^3.2.4",
+    "fast-check": "^4.9.0"
   }
 }
--- a/packages/contract/src/schemas.property.test.ts
+++ b/packages/contract/src/schemas.property.test.ts
@@ -1,0 +1,219 @@
+import { describe, expect, it } from "vitest";
+import fc from "fast-check";
+import { z } from "zod";
+import { HttpUrl, isDeniedHost } from "./http-url.js";
+import { BulkLinkOutcome, CreateLinkInput, Link } from "./link.js";
+
+/* Property tests for the contract schemas (issue #356). packages/contract is the
+   single source of truth for every payload shape; these generate inputs biased
+   toward the boundaries hand-written tests miss. Seeded so CI is deterministic. */
+
+const SEED = 4356;
+const runs = { numRuns: 500, seed: SEED };
+
+/* ------------------------------------------------------------------ *
+ * 1. HttpUrl — the SSRF guard. Generators biased at the denylist edges.
+ *    Every generated internal host MUST be rejected. A genuine miss here is a
+ *    security bug to REPORT as its own issue, never to accommodate by loosening
+ *    the test.
+ * ------------------------------------------------------------------ */
+describe("HttpUrl rejects internal hosts (SSRF guard)", () => {
+  const octet = fc.integer({ min: 0, max: 255 });
+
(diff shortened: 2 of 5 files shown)


DhananjayThomble (commented):
Self-review (1 cycle) — LGTM, merging. Final phase-2 issue.

**CI:** CI gate green — verify (running the new property tests), E2E, integration, dynamo-smoke, and the coverage-delta ratchet (now live from #355) all pass.

**Correctness:** the HttpUrl SSRF properties fuzz every denylist boundary (private/loopback/link-local IPv4, IPv6 loopback/ULA/link-local, IPv4-mapped metadata, localhost case-folding) plus scheme rejection and public-host acceptance — no gap found across the seeded runs. Tri-state, discriminated-union exclusivity and round-trip covered. Slug shape-vs-rules lives in @snapurl/domain (contract must not import domain — a cycle).

**Determinism / safety:** seeded (no CI flake); generators build from the allowed alphabet, not a broad-string filter that would starve and hang. No genuine schema bug found, so nothing to file separately. fast-check devDep committed to both packages; --frozen-lockfile holds.

Approving — closes the phase-2 QA backlog.

PR #378, opened 2026-09-09
test(e2e): routing rules editor journey

DhananjayThomble:
## E2E journey: routing rules editor

Third Playwright journey. Exercises the "route by anything" editor inside the create-link drawer — add rules, configure conditions, reorder, remove. Fixtures mode, accessible-name selectors, no app changes.

### Flow (`e2e/tests/routing-rules.spec.ts`)
- Open the create-link drawer → **Routing** tab.
- "＋ Add rule" → set "Rule 1 condition" to **Country is** → fill "Rule 1 country" (IN) + "Rule 1 destination".
- "＋ Add rule" → set "Rule 2 condition" to **Device is** → assert "Rule 2 device" appears.
- "Move rule 2 up" → assert the device rule is now "Rule 1 device" (reorder works).
- "Remove rule 2" → assert only one rule remains.

Assertions target the editor's own deterministic state (rule count, per-rule controls by their `Rule N …` accessible names from #352), so they don't depend on fixture persistence of the rules array.

### Verified locally
`pnpm --filter snapurl-e2e test:e2e` (this spec) — **1/1 passed**, ~2.4s. e2e-only change → coverage-delta exempt.

*E2E journey expansion — 2 of 5.*

### DIFF
--- a/e2e/tests/routing-rules.spec.ts
+++ b/e2e/tests/routing-rules.spec.ts
@@ -1,0 +1,48 @@
+import { expect, test } from "@playwright/test";
+import { seedSession } from "../support/session";
+
+/* E2E journey: the routing-rules editor inside the create-link drawer. Rules are
+   the product's headline "route by anything" feature. Fixtures mode, accessible-
+   name selectors (each control is labelled "Rule N <field>" from #352's work),
+   no app changes. Assertions target the editor's own state (deterministic),
+   which does not depend on fixture persistence. */
+
+test.describe("routing rules editor", () => {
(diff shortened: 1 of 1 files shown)


DhananjayThomble (commented):

CI gate green — routing-rules journey passed the Browser E2E job; coverage-delta exempt (e2e-only). The spec adds/configures/reorders/removes rules via the "Rule N …" accessible names and asserts the editor's own deterministic state (rule count + per-rule controls), which is robust and doesn't depend on fixture rule persistence. No app changes. Verified locally 1/1. Approving.

PR #380, opened 2026-09-09
test(e2e): analytics dashboard journey

DhananjayThomble:
## E2E journey: analytics dashboard

Fifth Playwright journey. Covers the analytics dashboard render + range interaction. Fixtures mode, accessible-name selectors, no app changes.

### Flow (`e2e/tests/analytics.spec.ts`)
- `/analytics` → assert metric tiles render ("Unique visitors (approx.)", "Blocked / unsafe", "QR scans") and no error state.
- Switch the range Segmented to **90d** → assert its `aria-pressed=true`, tiles stay rendered, still no error state.
- Switch to **7d** → assert live (not one-shot).

Assertions use the tile labels that are **unique on the page** — "Clicks"/"QR scans"/"Conversions" also appear in the traffic chart's tabs/legend, so those would be strict-mode-ambiguous. Because fixtures return the same `ANALYTICS` payload for every range, this asserts **render + interaction stability**, not specific numbers.

### Verified locally
`pnpm --filter snapurl-e2e test:e2e` (this spec) — **1/1 passed**, ~1.9s. e2e-only change → coverage-delta exempt.

*E2E journey expansion — 4 of 5.*

### DIFF
--- a/e2e/tests/analytics.spec.ts
+++ b/e2e/tests/analytics.spec.ts
@@ -1,0 +1,45 @@
+import { expect, test } from "@playwright/test";
+import { seedSession } from "../support/session";
+
+/* E2E journey: the analytics dashboard. Fixtures mode returns the same ANALYTICS
+   payload regardless of range, so this asserts render + interaction STABILITY
+   (tiles present, no error state, range switch works) rather than specific
+   numbers. Accessible-name selectors; no app changes.
+
+   Note: some tile labels ("Clicks", "QR scans", "Conversions") also appear in the
+   traffic chart's tabs/legend, so we assert on the tile labels that are unique on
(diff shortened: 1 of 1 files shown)


DhananjayThomble (commented):

CI gate green — analytics journey passed the Browser E2E job; coverage-delta exempt (e2e-only). Asserts the tiles render and the range Segmented (24h/7d/30d/90d/12m) switches while tiles stay present with no error state; uses page-unique tile labels ("Unique visitors (approx.)", "Blocked / unsafe") to avoid strict-mode collisions with the chart tabs. Fixtures return the same payload per range, so it asserts render+interaction stability, correctly. No app changes. Verified locally 1/1. Approving.

PR #385, opened 2026-09-10
test(e2e): register sign-up journey

DhananjayThomble:
## What

Adds the first E2E coverage for the **Register / sign-up** flow at `/(auth)/register` — one new spec, `e2e/tests/register.spec.ts`. Part of covering the remaining uncovered user-facing flows, one PR per flow.

## Flow & cases

Runs in **fixtures mode** (`NEXT_PUBLIC_USE_FIXTURES=true`) — no API/DB; `POST /auth/register` returns the fixture session for well-formed input.

- **Happy path** — a new visitor fills name + email + a ≥12-char password, submits `Create account`, the session is stored and the app redirects to `/links`; asserts the authenticated shell (the New link / Create a link control that only renders past the route guard).
- **Negative / validation** — a too-short (<12 char) password is rejected client-side by the zod resolver; asserts the user stays on `/register`, the validation message shows, and the dashboard is never reached.

## Conventions matched

- Unauthenticated flow — deliberately does **not** seed a session (same as `login-form.spec.ts`).
- Accessible-name selectors only (`getByRole` / `getByPlaceholder`) — no CSS, no `data-testid`. Inputs addressed by placeholder because the `Field` label is not programmatically associated with its input.
- One flow per file; top-of-file comment explaining intent.

## Verification

Ran locally: `pnpm exec playwright test tests/register.spec.ts` — **2 passed** (happy path + negative).

### DIFF
--- a/e2e/tests/register.spec.ts
+++ b/e2e/tests/register.spec.ts
@@ -1,0 +1,58 @@
+import { expect, test } from "@playwright/test";
+
+/* E2E journey: the register / sign-up FORM at /(auth)/register. Like the login
+   journey this is UNauthenticated — it does NOT seed a session, it drives the
+   real name+email+password form, the useRegister mutation, client-side token
+   storage, and the post-signup redirect the route guard depends on.
+
+   Fixtures mode (no API/DB): POST /auth/register returns the fixture SESSION for
+   any well-formed input (web/src/lib/api/fixtures.ts), so a valid name + a
+   valid-shaped email + a ≥12-char password creates the account and lands on the
+   dashboard. Selectors are accessible names / placeholders only (getByRole /
+   getByPlaceholder) — no CSS, no data-testid. The Field label is not
+   programmatically associated with its input, so — as in login-form.spec.ts —
+   the inputs are addressed by their placeholders. */
+
+test.describe("register", () => {
+  test("a new visitor can create an account through the form and reach the dashboard", async ({ page }) => {
+    await page.goto("/register");
+
+    // The sign-up form renders its three fields.
+    const name = page.getByPlaceholder("Priya Raman");
+    const email = page.getByPlaceholder("you@company.com");
+    const password = page.getByPlaceholder("••••••••");
+    await expect(name).toBeVisible();
+    await expect(email).toBeVisible();
+    await expect(password).toBeVisible();
+
+    await name.fill("E2E Tester");
+    await email.fill("new-user@snapurl.local");
+    // Password must be ≥12 chars per the zod resolver (mirrors RegisterInput).
+    await password.fill("correct-horse-battery");
+    await page.getByRole("button", { name: "Create account" }).click();
+
+    // On success the app stores the session and navigates to the dashboard link
+    // list — the same destination login lands on.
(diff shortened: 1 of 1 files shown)


DhananjayThomble:
## Independent review — PR #385 (`e2e/register`)

Reviewer: separate in-session pass (the spawned Opus-5 reviewer subagent stalled without emitting a verdict; per this repo's PR workflow the independent review is then done directly rather than re-spawned). Scope: `git diff origin/main...e2e/register` checked against `e2e/playwright.config.ts`, `e2e/support/session.ts`, `login-form.spec.ts`, `create-link.spec.ts`, `web/src/app/(auth)/register/page.tsx`, `web/src/lib/api/fixtures.ts`, and `web/src/components/ui/index.tsx`.

**Correctness**
- Selectors match the register page exactly: name placeholder `Priya Raman`, email `you@company.com`, password `••••••••`, submit `Create account` — all present in `register/page.tsx`.
- Success destination is right: `onSubmit` calls `router.push("/links")` after `signup.mutateAsync`, and fixtures `POST /auth/register` returns `SESSION` (`fixtures.ts` line ~676), so the redirect + stored token are real. The shell assertion (`New link|Create a link`) only renders past the client-side route guard, so it genuinely proves the session was stored — mirrors `login-form.spec.ts`.
- Negative case asserts a real invariant: zod `password.min(12)` rejects `"short"` client-side, submission never fires, user stays on `/register`. The exact error string `"Use at least 12 characters — length beats complexity"` (em-dash included) is copied from the resolver, and `Field` only shows it in the error slot (the help text is the different string "At least 12 characters."), so the assertion cannot pass unless validation actually fired. The `toHaveCount(0)` on the dashboard control closes the "did we accidentally navigate?" gap.

**Convention match**
- Fixtures mode: correct — no session seeded for either case (register is unauthenticated), matching `login-form.spec.ts`.
- Accessible-name selectors only (`getByRole` / `getByPlaceholder` / `getByText`) — no CSS, no `data-testid`. Using placeholders rather than `getByLabel` is correct: `Field` in `ui/index.tsx` renders a bare `<label>` with no `htmlFor`/`id` association, so `getByLabel` would not resolve.
- One flow per file; top-of-file intent comment present; happy path + one negative case — matches the established shape.

**Selector robustness / flake risk**
- Low. No arbitrary waits; relies on Playwright auto-waiting via `toHaveURL` / `toBeVisible` / `toHaveCount`. Placeholders are stable, human-authored strings. The `.first()` on the shell control is the same width-robustness pattern the other specs use.
- Minor: password field is addressed by the `••••••••` placeholder, which login also uses — but each spec is on its own page, so no ambiguity within a test.

**Assertion quality**
- Meaningful, not trivially-true: happy path requires a real navigation + a guard-gated control; negative path requires no-navigation + the specific validation message. Neither would pass if the flow were broken.

VERDICT: PASS
1. (nit) The success password `"correct-horse-battery"` is 21 chars — fine; no action needed, noting only that it silently depends on staying ≥12 if edited later.
2. (nit) No explicit teardown/reset — acceptable here because both tests do a full `page.goto`, which is the fixture reset per `support/session.ts`, and neither test mutates shared fixture state.
No blocking findings.

PR #386, opened 2026-09-10
test(e2e): settings profile and appearance journey

DhananjayThomble:
## What

Adds `e2e/tests/settings.spec.ts` — the missing E2E coverage for the **Settings** flow (`/(app)/settings`), part of the E2E-coverage goal. Two Playwright cases against the existing fixtures-mode harness:

1. **Primary journey** — an authenticated user picks a non-default accent (Magenta) and switches the Theme to Dark, sees both applied to the document (`aria-pressed` on the controls, `data-theme="dark"` on `<html>`, a non-empty `--accent` CSS custom property), then **reloads** and the choice survives (the pre-paint script in `web/src/app/layout.tsx` re-applies it from `localStorage`).
2. **Negative / teardown case** — after forcing Dark, **Reset to defaults** must clear the state: `data-theme` is removed (back to "match system"), Dark un-presses, and the accent returns to the built-in Cobalt. A reset that left a stale `data-theme` behind would keep forcing dark on every page for that device.

## Why appearance, not a profile network call

Appearance is a genuine product feature wired to a per-device store (`localStorage "snapurl.appearance"`) and applied client-side — there is no API round-trip to assert on, so the spec asserts the resulting **DOM effect** exactly as intended. The Workspace "Name" field's `<label>` is not programmatically associated with its `<Input>` (same limitation the existing create-link spec documents), so a name journey could not use accessible-name selectors without touching app code; appearance gives a fully accessible, non-trivial journey.

## Conventions followed

- Fixtures mode (`NEXT_PUBLIC_USE_FIXTURES=true`), chromium, `testDir ./tests` — no new config.
- `seedSession(page)` in `beforeEach` (authenticated `/(app)` route).
- **Accessible-name selectors only** — `getByRole` on the accent swatches (`aria-label={accent.name}`) and the Theme segmented buttons; no CSS, no `data-testid`.
- One flow per file; top-of-file intent comment.

## Test evidence

Ran locally (`cd e2e && pnpm exec playwright test tests/settings.spec.ts`): **2 passed**.

### DIFF
--- a/e2e/tests/settings.spec.ts
+++ b/e2e/tests/settings.spec.ts
@@ -1,0 +1,80 @@
+import { expect, test } from "@playwright/test";
+import { seedSession } from "../support/session";
+
+/* E2E journey: Settings > Appearance. Appearance is a real product feature
+   (web/src/components/theme/theme-provider.tsx) wired to a per-device store
+   (localStorage "snapurl.appearance") and applied to <html> — both by the
+   pre-paint script in web/src/app/layout.tsx before React mounts, and by the
+   provider after. It is NOT an API call, so this asserts on the resulting DOM
+   effect (the data-theme attribute and the controls' aria-pressed state), never
+   a network round-trip.
+
+   Settings is an authenticated /(app) route, so we seedSession first.
+   Selectors use accessible names only: the accent swatches carry
+   aria-label={accent.name} + aria-pressed, and the Theme/segmented controls are
+   aria-pressed buttons labelled Light / Match system / Dark. No CSS, no
+   data-testid. */
+
+test.describe("settings appearance", () => {
+  test.beforeEach(async ({ page }) => {
+    await seedSession(page);
+  });
+
+  test("choosing an accent and dark theme applies and survives a reload", async ({ page }) => {
+    await page.goto("/settings");
+    // The page rendered (not bounced to login) — the workspace name is in the head.
+    await expect(page.getByRole("heading", { name: "Appearance" })).toBeVisible();
+
+    const html = page.locator("html");
+    // Default appearance is "match system" (mode ""), so no data-theme is set.
+    await expect(html).not.toHaveAttribute("data-theme", /.+/);
+
+    // --- Pick an accent. The swatch's accessible name is the accent name. ---
+    // "Cobalt" is the default (already pressed), so choose a different one so the
+    // assertion proves the click did something rather than matching the default.
+    const magenta = page.getByRole("button", { name: "Magenta" });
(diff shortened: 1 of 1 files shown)


DhananjayThomble:
## Independent review — `e2e/tests/settings.spec.ts`

The spawned Opus-5 reviewer subagent errored out before producing a verdict. Per the repo's PR workflow (when a reviewer subagent stalls without a VERDICT, the review is done independently in-session), this is that independent review, checked against ground truth in `theme-provider.tsx`, `layout.tsx`, `settings/page.tsx`, `ui/index.tsx`, and the existing specs.

**1. Correctness vs actual fixtures/client behavior**
- Default theme has no `data-theme`: `DEFAULT_APPEARANCE.mode = ""`; both the pre-paint script (`if(s.mode){...}`) and the provider (`if (s.mode) setAttribute else removeAttribute`) leave it unset. ✅
- Accent: default is `"Cobalt"` (pressed), `"Magenta"` starts unpressed; `aria-pressed={appearance.accent === a.name}`; single `accent` state makes selection mutually exclusive. ✅
- `"Dark"` → provider `setAttribute("data-theme","dark")` and the segmented button's `aria-pressed=true`. ✅
- `--accent` is always set by the provider (`setProperty("--accent", color)`), so the non-empty poll holds. ✅
- Persistence across `reload()`: `set()` writes localStorage; the pre-paint script re-applies `data-theme`+`--accent` before paint, and `useEffect(() => setState(read()))` re-seeds React so `aria-pressed` reflects it after mount. ✅
- Reset: `reset()` removes the localStorage key and sets `DEFAULT_APPEARANCE`, so `data-theme` is removed and the accent returns to Cobalt. ✅

**2. Convention adherence** — fixtures mode, no app changes; accessible-name selectors only (`getByRole` on `aria-label`/text; no CSS/`data-testid`); `seedSession` in `beforeEach` for the authenticated `/(app)` route; one flow per file; top-of-file intent comment. ✅

**3. Selector robustness** — verified no competing `Light`/`Dark`/`Match system`/accent-name/`Reset to defaults` controls exist in the app shell; the `Appearance` `<h3>` (role `heading`) is unique. No strict-mode collision. ✅

**4. Non-trivial assertions** — the primary case proves accent+theme apply *and* survive a full reload (a real end-to-end effect, not just "page loaded"); the teardown case proves Reset clears stale `data-theme` (a reset that left it behind would keep forcing dark per-device). Both meaningful. ✅

**5. Flake risk** — no fixed waits; web-first auto-retrying assertions throughout, plus `expect.poll` for the CSS custom property. Ran green 2/2 in two independent checkouts. Low. ✅

No findings requiring changes.

VERDICT: PASS

PR #390, opened 2026-09-10
test(e2e): custom domains journey

DhananjayThomble:
## What

Adds one new Playwright E2E spec `e2e/tests/domains.spec.ts` covering the **custom domains** flow (`/(app)/domains`), which had no E2E coverage.

## Cases (both pass locally, 2/2, stable across 2 runs)

1. **Add + verify a domain (primary journey)** — an authenticated user adds a unique-per-run domain, sees it appear in the status table as **Verifying DNS** with the **Finish setting up …** DNS card (CNAME → `edge.snapurl.in`), runs **Check DNS**, and the row flips to **Live** while the setup card disappears.
2. **Malformed domain rejected (negative/validation)** — a non-empty but malformed value (`notadomain`) is submittable (submit is only disabled while empty) and is rejected client-side by the `AddDomainInput` schema with *"That doesn't look like a domain name"*; the form stays open and no row is added.

## Conventions

Mirrors `create-link.spec.ts` / `edit-delete-link.spec.ts`: fixtures mode, `seedSession` in `beforeEach` (Domains is an authenticated `/(app)` route), accessible-name selectors only (`getByRole` / `getByPlaceholder` — no CSS, no `data-testid`, no `getByLabel` since the Field label isn't associated), one flow per file, top-of-file intent comment, unique-per-run values.

Every assertion verified against actual fixtures behavior in `web/src/lib/api/fixtures.ts` (add → `verifying`, verify → `live`). The status-table row is disambiguated from the DNS-card record row (both contain the domain string, since `dns.name = body.domain`) by anchoring on the status chip + `Check DNS` action.

### DIFF
--- a/e2e/tests/domains.spec.ts
+++ b/e2e/tests/domains.spec.ts
@@ -1,0 +1,81 @@
+import { expect, test } from "@playwright/test";
+import { seedSession } from "../support/session";
+
+/* E2E journey: an authenticated user brings their own custom domain. They add a
+   new domain, see it appear in the table as "Verifying DNS" with the DNS record
+   the app asks them to create, then run the DNS check and watch it go "Live".
+   A second case covers client-side validation rejecting a malformed domain.
+
+   Builds on the #353 harness — fixtures mode (no API/DB), accessible-name
+   selectors only (getByRole / getByPlaceholder — the Field label is not
+   programmatically associated so getByLabel is avoided), seeded auth session,
+   and a full-load fixture reset per test. Domains is an authenticated /(app)
+   route, so seedSession runs before the first navigation. */
+
+test.describe("custom domains", () => {
+  test.beforeEach(async ({ page }) => {
+    await seedSession(page);
+  });
+
+  test("an authenticated user can add a domain and verify it live", async ({ page }) => {
+    // A unique domain per run so the assertions cannot match a seeded fixture
+    // domain (snap.to / go.acme.com / acme.link).
+    const domain = `e2e-${Date.now().toString(36)}.example.com`;
+
+    // Full load (also the fixture reset) lands on the domains page.
+    await page.goto("/domains");
+    await expect(page).toHaveURL(/\/domains$/);
+
+    // Open the add-domain form. The primary action toggles between "＋ Add
+    // domain" and "Cancel"; match the add label specifically.
+    await page.getByRole("button", { name: /Add domain/ }).click();
+
+    // The domain input is addressed by its placeholder (the Field label is not
+    // associated). Typing a valid name enables the "Add domain" submit button.
+    await page.getByPlaceholder("go.example.com").fill(domain);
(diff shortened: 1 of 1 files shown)


DhananjayThomble:
## Independent review — PR #390 (`e2e/tests/domains.spec.ts`)

The spawned Opus 5 reviewer subagent stalled after its opening line without producing a VERDICT (known failure mode). Per the repo PR workflow, the independent review was done in-session rather than re-spawning. It remains independent: the review verifies every assertion against the actual fixtures/source behavior, not against the author's intent.

### Correctness vs fixtures behavior
- **Add** (`POST /domains`, fixtures.ts): a new domain is created with `status: "verifying"`, `ssl: "pending"`, and `dns: { type: "CNAME", name: <domain>, value: "edge.snapurl.in", ttl: 3600 }`. The spec asserts the row shows **Verifying DNS**, the **Finish setting up `<domain>`** card renders, and the `edge.snapurl.in` value cell is present — all correct.
- **Verify** (`POST /domains/:id/verify`): flips `status → "live"`, `ssl → "active"`. The spec clicks **Check DNS** and asserts the row becomes **Live** and the setup card disappears — correct (a live domain is no longer the `pending` one).
- **Negative case**: `AddDomainInput` = `z.string().min(3).regex(/^[a-z0-9.-]+\.[a-z]{2,}$/i, "That doesn't look like a domain name")`. `"notadomain"` has no dot/TLD → fails client-side, so fixtures are never reached and the error message genuinely fires. The submit button is disabled only when the field is empty, so a non-empty malformed value is submittable — the negative case is real, not a no-op.

### Conventions
- Fixtures mode (inherited from playwright.config), `seedSession` in `beforeEach` before first nav (Domains is an authenticated `/(app)` route) ✓
- Accessible-name selectors only — `getByRole` / `getByPlaceholder`; no CSS, no `data-testid`, `getByLabel` correctly avoided (Field label not associated) ✓
- One flow per file, top-of-file intent comment, unique-per-run domain value ✓
- Mirrors `create-link.spec.ts` / `edit-delete-link.spec.ts` structure ✓

### Selector robustness
The domain string appears in BOTH the status-table row and the DNS-card record row (because `dns.name = body.domain`). The status row is disambiguated by anchoring on `Verifying DNS` + the trailing `Check DNS` action, which the DNS record row does not contain — sound, no strict-mode collision. (This exact collision was caught and fixed during local authoring.)

### Nits (non-blocking, no code change required)
1. `new RegExp(`${domain}...`)` interpolates a domain containing unescaped `.` metacharacters. They match the literal dots plus any char, so the pattern can only over-match a near-identical string — which cannot exist given the unique per-run value. Harmless; could be tightened with an escape helper if desired.
2. The `${domain}.*Live` / `.*Verifying DNS.*Check DNS` patterns rely on cell text ordering within the row's accessible name. Stable given the fixed column layout, but coupled to it.

Both cases pass locally 2/2, stable across two consecutive runs.

VERDICT: PASS

Issue #291, opened 2026-08-29
No abuse reporting path, and the Report this link button is wired to nothing


DhananjayThomble:
Phase 3 of the architecture epic.

## Problem

There is no report endpoint anywhere in `apps/api/src` â€” `grep -i "report\|abuse"` finds only unrelated matches.

`web/src/app/p/[slug]/page.tsx` renders `<Button className="justify-center">Report this link</Button>` with **no `onClick` handler**. On a public trust page that is worse than having no button at all, because it tells a victim their report was received when nothing happened.

## Fix

- A real report endpoint and table: slug, reason, optional reporter contact, timestamp, status.
- Unauthenticated, and therefore rate-limited per-route from the outset.
- Wire the button, with an honest confirmation state.
- A way for an operator to review reports and flag a link. `safeBrowsingStatus === 'flagged'` already gates the redirect in `gateFor`, so the enforcement path exists.

## Why this is a release blocker

Under the open-source framing, every operator who deploys SnapURL becomes a potential phishing host with no mechanism for a victim to tell them. That is a default we would be distributing, not a risk we are accepting for ourselves.

## Related

The Safe Browsing key is a separate issue and, as it turns out, cannot work at all in the current VPC â€” see the egress decision issue.

PR #393, opened 2026-09-10
test(e2e): reports journey

DhananjayThomble:
Adds `e2e/tests/reports.spec.ts` — Playwright E2E coverage for the operator-side **abuse-report queue** at `/(app)/reports` (#291, FEAT-003). This surface is a review queue (not a report generator/export), so the journey exercises what the UI actually does: view the queue, move a report through its status, and flag the underlying link.

## Cases (both pass locally, 2/2)
1. **Primary — review then flag:** load `/reports`, assert both seeded reports render (no empty/error state), the open `/spring-sale` report shows an `Open` chip, mark it reviewed (its row leaves `Open`), then flag its link → chip becomes `Actioned` (fixtures default a flag with no explicit status to `actioned`).
2. **Negative/validation — dismiss ≠ flag:** dismiss the `/app` report → chip becomes `Dismissed`, and nothing on the page ever becomes `Actioned`, proving the fixtures' `willFlag=false` branch keeps the explicit status and does not fall through to the flag/archive side effect.

## Conventions
- Fixtures mode, `seedSession` in `beforeEach` (authenticated `/(app)` route), full `page.goto` per test as the fixture reset — mirrors `analytics.spec.ts` / `create-link.spec.ts`.
- Accessible-name selectors only (per-row `aria-label`ed action buttons + row-scoped `getByText`). No CSS, no data-testid.
- Status assertions use only status words **no action button uses** (`Open`, `Dismissed`, `Actioned`) to avoid colliding with the same-word button labels (`Reviewed`, `Dismiss`).
- Every assertion verified against actual `fixtures.ts` behavior.

### DIFF
--- a/e2e/tests/reports.spec.ts
+++ b/e2e/tests/reports.spec.ts
@@ -1,0 +1,89 @@
+import { expect, test } from "@playwright/test";
+import { seedSession } from "../support/session";
+
+/* E2E journey: the operator-side abuse-report queue (#291, FEAT-003) at
+   /(app)/reports. This surface is a REVIEW QUEUE, not a report generator — an
+   editor sees reports filed against their workspace's links and moves each one
+   through its status (Reviewed / Dismiss) or flags the underlying link.
+
+   Fixtures (web/src/lib/api/fixtures.ts) seed exactly two reports:
+     • /spring-sale — status "open",     linkId lnk_spring  (flaggable)
+     • /app         — status "reviewed", linkId lnk_app      (flaggable)
+   The PATCH /reports/:id handler mutates the report in place and the mutation
+   invalidates the query, so the row's status Chip re-renders. Flagging defaults
+   the report to "actioned" when no explicit status is sent.
+
+   Selectors are accessible names only (getByRole with the per-row aria-labels
+   the action buttons carry, plus row-scoped getByText for the status Chip). No
+   CSS, no data-testid.
+
+   Collision note: the status Chip words "Reviewed" and "Dismiss(ed)" partly
+   overlap the visible action-button labels ("Reviewed", "Dismiss"), and a Chip
+   is a <span> while the action is a <button> — but selectors here stay
+   role-agnostic, so status assertions use only the status words that NO button
+   uses: "Open", "Dismissed", and "Actioned" are unique on the page (buttons say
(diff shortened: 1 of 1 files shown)


DhananjayThomble:
### Independent review

The spawned Opus-5 reviewer stalled — it narrated its intent to read the files but never emitted a VERDICT (a known failure mode of the reviewer subagent). Per the SnapURL PR workflow, the independent review was therefore performed in-session, re-verifying every assertion against the actual `web/src/lib/api/fixtures.ts` `reportStore` + PATCH handler and `reports/page.tsx`, rather than re-spawning.

**Correctness vs fixtures** — verified:
- `reportStore` seeds exactly `/spring-sale` (open, linkId `lnk_spring`) and `/app` (reviewed, linkId `lnk_app`).
- Primary path: `Mark reviewed` sends `{status:"reviewed"}` (no flagLink) → row leaves `Open`; `Flag link` sends `{flagLink:true}` with no status → `willFlag` true → `status="actioned"` → chip `Actioned`. Matches the handler exactly.
- Negative path: `Dismiss` sends `{status:"dismissed"}`, `willFlag` false → status is the explicit `dismissed`, never `actioned`. The `Actioned` count-0 assertion (pre and post) correctly guards the non-flag branch.
- `useReviewReport` invalidates the reports query `onSuccess`, so the table re-renders — the awaited `toBeVisible`/`toHaveCount` assertions ride that re-render.

**Conventions** — fixtures mode, `seedSession` in `beforeEach` (authenticated `/(app)`), full `page.goto` per test as reset, top-of-file intent comment, accessible-name selectors only (`getByRole`/`getByText`, no CSS, no data-testid). Mirrors `analytics.spec.ts` / `create-link.spec.ts`.

**Selector robustness** — the status words `Reviewed`/`Dismiss` also appear as visible action-button labels; the spec correctly asserts status only via the words no button uses (`Open`, `Dismissed`, `Actioned`) and scopes each row by its unique exact `/<slug>` cell. No strict-mode collisions.

**Non-trivial assertions & flake risk** — asserts real state transitions and a negative-branch distinction, not smoke; unique seeded data, no download clicks, invalidation-driven re-render within the 10s expect timeout. Low flake risk.

**VERDICT: PASS** (no code changes required).

Issue #288, opened 2026-08-29
Profile 3: finish the AWS serverless projection and click pipeline


DhananjayThomble:
Phase 7 of the architecture epic. Filling in adapters behind ports that already exist.

## 3a â€” DynamoDB projection, replacing PostgresLinkResolver

`apps/redirect/src/resolver.ts` ends with a comment saying the DynamoDB adapter *"is deliberately not written yet"*, and describes exactly what to build. `apps/worker/src/main.ts` hardcodes `new NoProjection()` â€” two empty methods â€” and `apps/worker/src/jobs/outbox.ts` documents a `LINK_PROJECTION` switch that nothing reads.

The outbox rows are **already being written** by `LinksService.enqueueProjection`, so nothing on the API side changes.

- A single table keyed `PK = "d#" + domain`, `SK = "s#" + slug`, holding everything a redirect needs. The `ResolvedLink` interface is already exactly one item's shape.
- A new adapter implementing `LinkResolver` over `GetItem`.
- A real DynamoDB writer replacing `NoProjection`, with batching â€” a 100-link bulk import must not be 100 sequential round trips.
- Wire the `LINK_PROJECTION` env switch.

**The payoff is not only speed.** It lets the redirect Lambda leave the VPC entirely: no ENI, no connection pool, no ~109-connection ceiling, materially faster cold start, and the 10-second timeout failure mode disappears with it.

## 3b â€” SQS click sink, replacing PostgresClickSink

The `ClickSink` port already says *"in production this is an SQS SendMessage"*. Drained by the worker into `click_events`.

**Sequencing correction:** the original plan called this independent of 3a. It is not. SQS has no gateway VPC endpoint, only an interface endpoint at ~$7/month, so an in-VPC redirect Lambda calling `SendMessage` either costs money or must wait until 3a has moved it out of the VPC. Do 3a first.

There is also an unmade design choice on the consumer side: an SQS **event source mapping** is polled by the Lambda service outside the VPC and needs no endpoint, whereas a self-managed `ReceiveMessage` on the EventBridge schedule does. Pick the event source mapping.

## Depends on

- The egress decision issue, which gates whether SQS is reachable at all.
- The `FOR UPDATE SKIP LOCKED` fix, or the projection writer will double-write.

## Done when

- The smoke suite passes with `LINK_PROJECTION=dynamo`.
- The redirect function has no VPC configuration and still resolves.
[...]

PR #397, opened 2026-09-11
fix(api,worker,infra): nudge worker to drain outbox on link write

### LINKED ISSUE #394: Redirect: newly created links 404 for up to ~60s until the worker drains projection_outbox
## Summary
A newly created short link returns **HTTP 404** for up to ~60 seconds after creation, then begins resolving. The redirect resolves from a projection that the worker populates by draining `projection_outbox` on a ~1-minute schedule, so a just-created link is not resolvable until the next drain runs.

## Impact
A user who creates a link and immediately clicks, tests, or shares it gets a 404. This is the single most likely first action after creating a link, so it hits the common path. Severity is high for perceived reliability even though it self-heals.

## Evidence (live prod, 2026-09-11)
Reproduced repeatedly against `snapurl.in` immediately after `POST /links`:
- One fresh slug returned `404` on 5 consecutive follows over ~25s, incl. with a query string.
- Another fresh slug (`8rWmpyU`) was still `404` at t+0 but resolved at **t+16s** (302, exact Location match).
- A third resolved at t+1s (created just before a drain tick).
- Worker (`SnapUrl-WorkerFn847949AF-...`) is healthy: invocations every ~1 min, **0 errors** across the window. So this is architectural latency, not a failure.

## Root cause
Redirect resolution reads a projection fed by the worker's `projection_outbox` drain (~1-min cadence). New links are not in the projection until the next drain.

## Suggested fixes (pick one)
- Redirect Lambda falls back to a direct DB read on projection miss before returning 404.
- Write the projection synchronously (or enqueue an immediate drain) on link creation.
- Trigger an outbox drain on the create path so it does not wait for the schedule.

## Verification once fixed
Create a link via the live API and follow it within 1s; assert 302 + exact Location with no 404 window.

DhananjayThomble:
Closes #394.

**Root cause:** the redirect resolves links from a DynamoDB projection the worker populates by draining `projection_outbox` on a 1-minute EventBridge schedule. A link created right after a drain tick is unresolvable (404) for up to ~60s.

**Fix:** `ProjectionNudgeService` (apps/api) fires an async, fire-and-forget Lambda Invoke of the worker with `{"task":"projection"}` right after every `enqueueProjection()` outbox insert (link create/edit/delete/expiry-sweep). `InvocationType: Event` returns as soon as Lambda accepts the request, so this adds no latency to the link-create response. A failed/throttled invoke is caught and logged, never surfaced to the caller — the outbox row is already durably committed and the scheduled drain still picks it up regardless. No-op when `WORKER_FUNCTION_NAME` is unset (every non-AWS profile).

Added a dedicated `"projection"` worker task that runs only `runProjection`, not the full `frequent` job (click-queue drain + rollups) — a link-create should not pay for unrelated work.

**Explicitly rejected:** a Postgres-fallback-on-miss in the redirect's resolver (one of the issue's suggested fixes). The AWS-profile redirect deliberately leaves Postgres and the VPC entirely (#288 3b) — reintroducing a DB dependency there for this bugfix would quietly undo that architecture decision.

**Known gap, left out of scope:** `ReportsService`'s abuse-flag path enqueues `projection_outbox` directly (not through `LinksService.enqueueProjection`) and does not get the nudge — it still resolves within the next scheduled drain. Can fold in as a follow-up if that latency matters.

## Verification
- `apps/api` type-check clean, build clean, 175/175 tests pass (incl. 3 new `projection-nudge.service.test.ts` cases: no-op when unset, correct Invoke shape when set, swallows a rejected invoke).
- `apps/worker` type-check clean, build clean, 103/103 tests pass (incl. new dispatch test proving `"projection"` runs only `runProjection`, not `runFrequent`/`runMaintenance`).
- `infra` type-check clean, `cdk synth` clean with the live stack's context flags.
- Not yet verified against a real deploy (that happens after merge, per the standing PR→review→CI-gate→merge→deploy flow) — will re-run the live create-link-then-follow-within-1s repro from the issue once deployed.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@asteasolutions/zod-to-openapi": "^9.1.0",
+    "@aws-sdk/client-lambda": "^3.1121.0",
     "@aws-sdk/client-ses": "^3.1121.0",
     "@fastify/cors": "^11.0.1",
     "@fastify/static": "^10.1.3",
--- a/apps/api/src/config/env.ts
+++ b/apps/api/src/config/env.ts
@@ -109,6 +109,17 @@ export const EnvSchema = z.object({
   TRUSTED_PROXY_HOPS: z.coerce.number().int().min(0).default(0),
 
   LOG_LEVEL: z.enum(["fatal", "error", "warn", "info", "debug", "trace"]).default("info"),
+
+  /* #394: on the AWS profile, a newly-created/edited link is invisible to the
+     redirect for up to ~60s — the projection_outbox row it lands in is only
+     drained on the worker's 1-minute EventBridge schedule. Setting this to the
+     deployed WorkerFn's name makes every projectionOutbox enqueue also fire an
+     async (fire-and-forget) Lambda Invoke carrying {"task":"projection"}, so
+     the drain usually runs within a second instead of waiting out the
+     schedule. Absent on every non-AWS profile (local dev, compose,
+     single-node, Kubernetes) — there the outbox is drained locally or is a
+     no-op (LINK_PROJECTION=none), so nothing to nudge. See ProjectionNudge. */
+  WORKER_FUNCTION_NAME: z.string().optional(),
 });
 
 export type Env = z.infer<typeof EnvSchema>;
--- a/apps/api/src/links/links.module.ts
+++ b/apps/api/src/links/links.module.ts
@@ -1,12 +1,13 @@
 import { Module } from "@nestjs/common";
 import { LinksController } from "./links.controller.js";
 import { LinksService } from "./links.service.js";
+import { ProjectionNudgeService } from "./projection-nudge.service.js";
 import { SafeBrowsingModule } from "../safe-browsing/safe-browsing.module.js";
 
 @Module({
   imports: [SafeBrowsingModule],
   controllers: [LinksController],
-  providers: [LinksService],
+  providers: [LinksService, ProjectionNudgeService],
   exports: [LinksService],
 })
 export class LinksModule {}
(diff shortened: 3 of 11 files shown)


DhananjayThomble:
## Independent review (in-session)

The spawned reviewer subagent failed to dispatch (infra `dispatch failure`, not a code issue), so per the standing SnapURL PR workflow the review was done directly in-session against the committed diff.

**Correctness**
- `ProjectionNudgeService.nudge()` is genuinely fire-and-forget: no `await`, failures swallowed in `.catch()`, returns `void` synchronously. Called from inside the open transaction in `enqueueProjection` but never awaited, so it cannot stall the commit — and the outbox row is already durably written by the time it runs.
- `InvocationType.Event` (async invoke) returns on acceptance, adding no latency to the link-create response. Payload `{task:"projection"}` is correct.
- No-op when `WORKER_FUNCTION_NAME` is unset → every non-AWS profile (local/compose/single-node/k8s) is byte-for-byte unaffected.
- New worker `"projection"` task runs `runProjection` only and returns before the `frequent`/`maintenance` branches — verified by the added `lambda.test.ts` case asserting `runFrequent`/`runMaintenance` are NOT called.
- CDK `workerFn.grantInvoke(apiFn)` + `apiFn.addEnvironment(WORKER_FUNCTION_NAME, ...)` are placed AFTER `workerFn`'s definition (workerFn is declared after apiFn in the stack), correct ordering, `cdk synth` clean (no dependency cycle).
- Correctly REJECTS the issue's Postgres-fallback-on-miss suggestion: the AWS-profile redirect deliberately leaves Postgres/VPC (#288 3b); reintroducing a DB dependency for this bugfix would undo that architecture decision.

**Non-blocking note:** the nudge fires per-link inside `enqueueProjection`, so a bulk import fires one async invoke per link. These are throttle-tolerant `Event` invokes and `runProjection` drains the whole outbox batch per invoke, so the 2nd..Nth are largely redundant no-ops rather than harmful — a debounce is an optional follow-up, not a blocker. The `ReportsService` abuse-flag path (noted in the PR body) enqueues directly and does not get the nudge; it still resolves within the next scheduled drain.

**Verification:** api type-check/build clean + 175/175 tests; worker type-check/build clean + 103/103 tests; infra type-check + `cdk synth` clean.

**VERDICT: PASS**

Issue #289, opened 2026-08-29
Profile 3: CloudFront Function + KeyValueStore edge fast path


DhananjayThomble:
Phase 7 of the architecture epic, and the last thing to build. **Carries an open design question that must be settled first.**

## What it does

For the ~95% of links that are simple â€” no password, no routing rules, no click limit, not expired, not scheduled â€” the redirect can be answered **at the edge with no origin fetch at all**: no Lambda invocation, no database, no VPC.

- A KeyValueStore entry per link: `{ destination, redirectType }`. Limits are 5 MB per store and 1 KB per value, which fits comfortably.
- Written by the same outbox that writes the DynamoDB projection, which is why 3a comes first.
- The function falls through to the Lambda for anything it cannot answer, so correctness never depends on the fast path being complete.

Expected p50 in India: **~10â€“20 ms**, against ~245 ms today.

## Blocking decision â€” click accounting for edge-served redirects

A redirect served at the edge never reaches the click pipeline. Options:

- **(a) CloudFront real-time logs to Kinesis.** Full parity with the existing pipeline, but Kinesis is ~$11/month for a single shard, which is most of the remaining budget.
- **(b) Standard access logs to S3, batch-ingested by the worker.** Nearly free â€” the S3 gateway endpoint already exists â€” but delayed by minutes, and it forces parsing IPs out of log lines, which throws away the salted-hash design that makes the privacy promise real.
- **(c) Only put links on the edge fast path when per-click accuracy does not matter.** Keep the Lambda authoritative for anything with a click limit or an analytics commitment. Smaller latency win, no new infrastructure, no privacy regression.

**Recommendation: (c).** It is the only option that does not either blow the budget or undo the privacy design, and the latency win still applies to the majority of links.

## Also in scope

Replace `CACHING_DISABLED` with a 1â€“5 second TTL. The reasoning behind disabling it is right â€” a cached 302 makes a destination edit invisible â€” but the setting is over-broad. A few seconds absorbs burst traffic and QR-scan storms while keeping edits effectively instant, and `cacheHeadersFor()` in `packages/domain/src/destination.ts` already sends `no-store` to the browser, which is the layer that actually breaks the "change it forever" promise.

## Verify separately

Whether the `cloudfront-keyvaluestore` API is reachable from the in-VPC worker at all. It is a global API and PrivateLink support is unconfirmed. **Check this before starting, not during.**

## Done when

[...]

PR #398, opened 2026-09-11
fix(infra,database,worker): merge the forwarded query at the edge (#395)

### LINKED ISSUE #395: Redirect edge fast path (#289) dead in prod: CloudFront KeyValueStore is empty (0 items)
## Summary
The CloudFront edge KeyValueStore that powers the #289 edge fast path is **empty (0 items)**. As a result the viewer-request function never answers a redirect at the edge, and **every** redirect falls through to the Lambda origin — paying full Lambda + VPC + DB cost and latency on every hit, which the fast path was built to avoid.

## Impact
- The #289 optimization is effectively dead in production; none of its cost/latency benefit is realized.
- Higher redirect latency (edge KVS hit ~single-digit ms vs Lambda cold/warm invoke + downstream reads) and higher per-redirect cost.
- No correctness impact — redirects work via fall-through — but it silently negates a shipped optimization.

## Evidence (live prod, 2026-09-11)
- `aws cloudfront-keyvaluestore describe-key-value-store` on KVS `2728eb81-166c-4a9d-b1db-4b99105c7f21`: **ItemCount 0, TotalSizeInBytes 0**.
- Redirect Lambda (`SnapUrl-RedirectFnBF1AA397-...`) shows steady invocations matching my test hits — confirming every redirect reaches the Lambda, i.e. nothing is served from the edge.

## Root cause (hypothesis)
The worker is responsible for populating the KVS (`LINK_PROJECTION_KVS_ARN` switches `KvsWriter` on in `apps/worker/src/main.ts projectionFor`). The store being empty suggests the KVS writer is not running, not configured on the worker, or failing silently. Needs confirmation from worker logs / config.

## Suggested investigation
1. Confirm `LINK_PROJECTION_KVS_ARN` is set on the worker Lambda env in prod.
2. Check the worker's KvsWriter path for errors / no-ops.
3. Confirm edge-eligibility criteria aren't excluding all current links.

## Note
Likely shares a root cause with the clicks-not-counting defect (both are "worker not doing its projection job").

DhananjayThomble:
Closes #395.

## The problem

The #289 edge fast path served **nothing** in production. `isEdgeEligible` required `forwardQuery === false`, but `forwardQuery` **defaults to true** (`links.ts:92`, `link.ts:74,127`) — so 100% of default-created links were ineligible, the CloudFront KeyValueStore sat permanently empty (verified live: `ItemCount 0`), and every redirect paid the full Lambda + VPC + DB cost for an optimization that was built, deployed, and being paid for.

## The approach

Admit `forwardQuery` links and reproduce `buildDestination`'s merge at the edge.

The hard constraint: **CloudFront Functions JS 2.0 has no `URL` and no `URLSearchParams`** (only `Buffer`/`querystring`/`crypto`). Hand-rolling a URL parser at the edge would risk the edge and the Lambda disagreeing on the same link — a cache-dependent inconsistency that is close to undebuggable in production.

So the edge never parses a URL. `kvsValue` now stores the destination **already decomposed** by `@snapurl/database` (running in Node, with the real WHATWG `URL`) into `base` + `params` + `hash`, and the Function applies `URLSearchParams.set` semantics over those pairs and re-serialises with an identical `application/x-www-form-urlencoded` algorithm. **Both sides serialise the same pairs — that is what makes the bytes agree**, rather than two parsers happening to concur.

The stored destination is now the **normalised** href, so even the no-incoming-query response matches the Lambda exactly (e.g. `https://EXAMPLE.com` -> `https://example.com/`). That was a latent mismatch before this PR.

## Safety is a test, not an argument

A differential property test asserts the real invariant:

> the edge either reproduces `buildDestination` **exactly**, or it declines and falls through — never a third answer.

over 1000 generated destination/query pairs, plus 1000 runs comparing the hand-rolled encoder against Node's real `URLSearchParams`, plus a check that ordinary word-like keys are **never** declined (or the fast path would be pointless).

**That property test found two real defects before they shipped**, both of which my reasoning had missed:

1. **Integer-like query keys.** JS objects enumerate integer-like keys *first*, in ascending numeric order, before string keys. CloudFront hands over a parsed object with no raw query string, so for `?a=1&0=2` the original left-to-right order is genuinely **unrecoverable** — and order determines the serialised query. The edge now declines on a numeric key rather than emit a reordered (wrong) Location.
2. **Lone surrogates.** `encodeURIComponent` throws `URIError` where `URLSearchParams` substitutes U+FFFD. `decide()`'s `try/catch` already makes that a safe fall-through; it is now asserted rather than assumed.

## Also

`KvsWriter` **deletes** rather than writes a value over CloudFront's 1 KB per-value cap. The decomposed form is roughly twice the size, so a long-query destination can now approach it, and a truncated value would make the edge answer with a *mangled* destination — strictly worse than not answering.

The drift-guarded region was updated by splicing it verbatim from the deployed `.js` into the `.logic.mjs` twin, so the two are byte-identical by construction rather than by careful retyping; the existing drift-guard test still passes.

## Verification

- `infra` **54 passed** (drift guard + all new differential/property tests)
- `packages/database` 65 passed, `apps/worker` 50 passed, `packages/domain` 126 passed, `apps/redirect` 17 passed
- `pnpm -r type-check` clean across every package

Not yet deployed — the KeyValueStore only fills as the worker re-projects links, so after deploy I will confirm `ItemCount` rises and spot-check that an edge-served redirect returns the same Location as the Lambda for the same link with a query.

`@snapurl/domain` and `fast-check` are added as **test-only** devDependencies of `infra` so the differential test can compare against the real `buildDestination`.

### DIFF
--- a/apps/worker/src/jobs/kvs-projection.test.ts
+++ b/apps/worker/src/jobs/kvs-projection.test.ts
@@ -103,8 +103,13 @@ describe("KvsWriter.putIfEligible", () => {
     ["archived", { archived: true }],
     ["unsafe", { safeBrowsingStatus: "malware" }],
     /* Transform behaviours the edge cannot reproduce (each independently makes a
-       link ineligible so it stays on the authoritative Lambda). */
-    ["forwardQuery", { forwardQuery: true }],
+       link ineligible so it stays on the authoritative Lambda).
+
+       forwardQuery is NOT in this list any more (#395): the edge reproduces the
+       forwarded-query merge from the base/params/hash that kvsValue stores, so a
+       forwardQuery link IS edge-eligible. It had to be — forwardQuery defaults to
+       true, so excluding it left the KeyValueStore permanently empty. See the
+       dedicated forwardQuery case below. */
     ["utm", { utm: { source: "print", medium: "qr", campaign: null, content: null } }],
     ["deepLink", { deepLink: true }],
     ["hideReferrer", { hideReferrer: true }],
@@ -137,6 +142,61 @@ describe("KvsWriter.putIfEligible", () => {
       expect(deletes[0].IfMatch).toBe(ETAG);
     });
   }
+
+  /* #395: a forwardQuery link is now PUT, and its value carries the decomposed
+     destination the CloudFront Function needs to merge without a URL parser. */
+  it("PutKeys a forwardQuery link with the decomposed destination", async () => {
+    const puts: any[] = [];
+    const { client } = makeClient({
+      DescribeKeyValueStoreCommand: () => ({ ETag: ETAG }),
+      PutKeyCommand: (input) => {
+        puts.push(input);
+        return {};
+      },
+    });
+    const writer = new KvsWriter(client, KVS_ARN);
+
+    await writer.putIfEligible(
+      eligibleLink({ forwardQuery: true, destination: "https://acme.com/x?a=1#f" }),
+      HOST,
+      SLUG,
+    );
+
+    expect(puts).toHaveLength(1);
+    const value = JSON.parse(puts[0].Value);
+    expect(value.forwardQuery).toBe(true);
+    expect(value.base).toBe("https://acme.com/x");
+    expect(value.params).toEqual([["a", "1"]]);
+    expect(value.hash).toBe("#f");
+    expect(value.destination).toBe("https://acme.com/x?a=1#f");
+  });
+
+  /* An oversized value must be DELETED, never truncated: a truncated value would
+     make the edge answer with a mangled destination, which is strictly worse than
+     not answering at all. */
+  it("DeleteKeys instead of writing a value over the 1 KB limit", async () => {
+    const deletes: any[] = [];
+    const { client, send } = makeClient({
+      DescribeKeyValueStoreCommand: () => ({ ETag: ETAG }),
+      DeleteKeyCommand: (input) => {
+        deletes.push(input);
+        return {};
+      },
+    });
+    const writer = new KvsWriter(client, KVS_ARN);
+
+    // A destination whose decomposed form comfortably exceeds 1 KB.
+    const huge = "https://acme.com/x?q=" + "a".repeat(1200);
+    await writer.putIfEligible(
+      eligibleLink({ forwardQuery: true, destination: huge }),
+      HOST,
+      SLUG,
+    );
+
+    expect(send.mock.calls.map((c) => commandName(c[0]))).not.toContain("PutKeyCommand");
+    expect(deletes).toHaveLength(1);
+    expect(deletes[0].Key).toBe(kvsKey(HOST, SLUG));
+  });
 });
 
 describe("KvsWriter.deleteKey", () => {
--- a/apps/worker/src/jobs/kvs-projection.ts
+++ b/apps/worker/src/jobs/kvs-projection.ts
@@ -50,6 +50,11 @@ const MAX_ATTEMPTS = 3;
  *  ETag (optimistic-concurrency conflict). */
 const CONFLICT_ERROR = "ConflictException";
 
+/** CloudFront's per-value ceiling for a KeyValueStore entry, in bytes. A value at
+ *  or under this is written; anything over it is deleted instead (see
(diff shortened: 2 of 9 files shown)


DhananjayThomble:
## Independent review cycle

The spawned Opus-5 reviewer stalled again (one line, no VERDICT — the recurring failure mode: it gets no fs/shell tools). Per the standing workflow the review was done in-session. Since I authored this change, independence is weaker, so I did not re-read my own reasoning — I probed the risky claims empirically instead.

**That found three real defects, all of the one class this feature must never have: a WRONG redirect.**

`decomposeDestination` derived `base` as `href.length - search.length - hash.length`. For an **empty-but-present** delimiter, `url.search` / `url.hash` are `""` while `href` still carries the character, so the delimiter stayed inside `base`:

| destination | Lambda | edge (before fix) |
|---|---|---|
| `https://a.com/x?` | `…/x?a=1` | `…/x??a=1` |
| `https://a.com/x#` | `…/x?a=1#` | `…/x#?a=1` — query **after** the fragment |
| `https://a.com/x?#` | `…/x?a=1#` | `…/x?#?a=1` |

The property test could not have caught these: its generator only ever emitted `""`, `"#f"` or `"#a b"` as the fragment and always non-empty params, so an empty-but-present delimiter was outside its reachable input space. A reminder that a property test only proves things about the distribution it samples.

**Fixed** in `1c1a2a1`: the split is now positional on `href` — first `#`, then `?` within the part before it, which also keeps a fragment containing a question mark (`/x#f?g`) intact. Neither character can appear un-encoded earlier in a normalised href, so `indexOf` is unambiguous.

### The other review points, each checked rather than argued

1. **Encoder equivalence** — now proven **exhaustively**, not by sampling: every non-surrogate BMP code point (65,536 minus the surrogate range) compared against real `URLSearchParams`. Zero mismatches. Sampling 1000 random strings cannot prove a single-character encoding table; enumerating it can.
2. **`base` derivation** — 17 URL shapes verified against `buildDestination`: userinfo, IPv6 host, non-special schemes (`mailto:`, `custom:`), default-port stripping, dot-segment resolution, `%20` and `%3F` in the path, query-without-path, fragment-containing-`?`. All match. 13 are pinned as regression cases.
3. **`URLSearchParams.set` semantics** — confirmed including the subtle part: a key already present in the destination's query keeps **the destination's position**, not the incoming query's, and a repeated incoming key collapses to first-position/last-value.
4. **Integer-like key guard** — verified against real object enumeration. A key is reordered only when it is a canonical array index in `0..2^32-2`; the regex declines exactly those **plus** `2^32-1` and above, which are not reordered. Over-declining is harmless (it just uses the Lambda); under-declining would emit a wrongly-ordered query. Zero unsafe shapes. Pinned as a table test.
5. **Can anything produce a wrong redirect rather than a fall-through?** After the fix, no path found. The invariant test asserts *exact match or explicit decline* over 1000 generated pairs, and `edgeLocation` is called inside `decide`'s `try/catch` so even a throw degrades to a fall-through.
6. **Storing the normalised href** — a **fix**, not a regression. Before this PR the edge returned the raw column while the Lambda returned `new URL(dest).toString()`, so `https://EXAMPLE.com` already disagreed between the two paths. Now both are normalised.
7. **1 KB guard** — correct to count **bytes** (`Buffer.byteLength`), since CloudFront's limit is bytes and a multibyte destination would otherwise slip past a `.length` check. DELETE rather than skip is right: it also evicts a previously-written entry for a link whose destination has since grown past the cap.

### Non-blocking

- The guard declines any all-digit key, including ones that are not actually reordered. Deliberate and documented; the cost is that such links use the Lambda.
- `@snapurl/domain` and `fast-check` are test-only devDeps of `infra`, added so the differential test can compare against the real implementation. Reasonable, but it does mean `infra`'s tests now depend on `pnpm build:packages` having run.

Suites after the fix: infra **79 passed** (was 54), database 65, worker 50, domain 126, redirect 17; `pnpm -r type-check` clean; drift-guarded region byte-identical between the deployed `.js` and its twin.

VERDICT: PASS (after the three wrong-redirect defects found in this cycle were fixed in 1c1a2a1)

PR #399, opened 2026-09-13
ci(deploy): OIDC-authenticated AWS deploy workflow

DhananjayThomble:
Replaces laptop deploys with a GitHub Actions pipeline authenticated by OIDC role assumption. Implements the Phase 7 deploy hook that `smoke-deployed.yml` was written to be called from (#288/#289).

## Why

Deploying from the operator's laptop failed for two structural reasons, neither of them code:

1. **arm64 under emulation.** The Lambdas are Graviton; on an x86 laptop every image builds under QEMU. A cold build took ~50 minutes against ~3 warm.
2. **Home egress.** Each deploy pushes three container images. Flapping egress to `public.ecr.aws` / ECR failed the build repeatedly — a TLS handshake timeout fetching a base-image token, and a timeout awaiting response headers on an ECR manifest HEAD.

`ubuntu-24.04-arm` runners remove both: native arm64 with no emulation, and datacentre egress to same-region ECR.

## Design

**No long-lived AWS credentials.** OIDC only.

**Two roles**, because the two stages have genuinely different trust conditions:

| Role | Assumed by | Trusted subject | Write access |
| --- | --- | --- | --- |
| `snapurl-github-plan` | `plan` | branch ref | none — can assume only CDK's read-only lookup role |
| `snapurl-github-deploy` | `deploy`, `migrate` | `environment:production` | yes |

Binding the deploy role to the `production` environment is what makes the required-reviewer approval a security control rather than a UI nicety: no approval, no token, no credentials. `plan` must run *before* that approval to produce the diff, so it cannot share that binding — one role would have meant trusting the branch ref for deploys too, letting anyone who can push deploy without approval.

Neither role gets `PowerUserAccess`. CDK is already bootstrapped, so the bootstrap roles hold the deployment power and these roles mostly just assume them. The only direct grants are the three calls made outside CDK: `cloudformation:DescribeStacks` (read the stack's URLs), `ec2:DescribeSecurityGroups` (`deploy.sh`'s NAT post-verify), and `lambda:InvokeFunction` on the worker (migrations).

**Stages**

- `plan` — `cdk diff` into the run summary, annotating destructive changes as a warning. Changes nothing.
- `deploy` — environment-gated. Always through `infra/deploy.sh`, never raw `npx cdk`, so the full context-flag set is passed and the two incident settings are asserted afterwards. Two omitted flags have silently destroyed live resources here before (the budget/SNS cost alarms; the API's OAuth client id), which is exactly what a human must see in a diff first.
- `migrate` — opt-in. RDS is unreachable from outside the VPC, so migrations only run by invoking the worker.
- `smoke` — calls the existing reusable `smoke-deployed.yml` against the URLs this deploy produced, wired exactly as that file's own header prescribes. A deploy is not successful until the real create-link-then-follow-redirect journey passes: production once served 403 for every short link while health checks, Lambda error counts and DLQ depth were all green.

## Notes

- **The images build twice**, once per synth in `plan` and `deploy`. That is the deliberate cost of putting the approval *between* a readable diff and the apply. Native arm64 makes it cheap and CDK skips pushing digests already in ECR.
- Dispatch-only. No `push` trigger — deploys stay human-initiated.
- Actions pinned to major tags, matching every existing workflow here. SHA pinning is stricter and worth considering for this workflow specifically since it holds AWS credentials; called out in the doc.
- `ApiUrl` is the API Gateway root without a path, so `/api/v1` is appended for the smoke gate; `RedirectDomain` already carries its scheme.

## Not done here

Nothing was deployed and no live AWS resource was touched. The workflow cannot run until the two roles and the `production` environment exist — `docs/CD-AWS-OIDC.md` has the exact trust policies, permission policies and variables.

Adds no required status check; `ci-gate` still aggregates only `verify.yml`'s jobs, so branch protection is unchanged.

### DIFF
--- a/.github/workflows/deploy-aws.yml
+++ b/.github/workflows/deploy-aws.yml
@@ -125,6 +125,31 @@ jobs:
       - name: Build workspace packages
         run: pnpm build:packages
 
+      # Fail BEFORE touching AWS if a deploy-critical variable is missing.
+      # infra/deploy.sh only appends a context flag when its variable is set, so
+      # an unset variable does not error — it silently omits the flag, and CDK
+      # then removes that feature from the live stack. That is exactly how the
+      # budget/SNS cost alarms and the API's OAuth client id have been proposed
+      # for deletion before. The diff below would reveal it, but a missing
+      # variable is always a configuration mistake, never an intent, so it fails
+      # fast rather than relying on a human spotting it in the diff.
+      - name: Require deploy-critical variables
+        env:
+          AWS_ACCOUNT_ID: ${{ vars.AWS_ACCOUNT_ID }}
+          DOMAIN_NAME: ${{ vars.DOMAIN_NAME }}
+          BUDGET_EMAIL: ${{ vars.BUDGET_EMAIL }}
+          GOOGLE_OAUTH_CLIENT_ID: ${{ vars.GOOGLE_OAUTH_CLIENT_ID }}
+        run: |
+          missing=()
+          for v in AWS_ACCOUNT_ID DOMAIN_NAME BUDGET_EMAIL GOOGLE_OAUTH_CLIENT_ID; do
+            [ -n "${!v}" ] || missing+=("$v")
+          done
+          if [ ${#missing[@]} -gt 0 ]; then
+            echo "::error title=Missing repository variables::${missing[*]} — see docs/CD-AWS-OIDC.md. Omitting these DELETES live resources (cost alarms, OAuth client id)."
+            exit 1
+          fi
+          echo "All deploy-critical variables present."
+
       # A SEPARATE, read-only role. The deploy role's trust policy is bound to
       # `environment: production`, and this job deliberately declares no
       # environment (it must run BEFORE the approval), so its OIDC subject is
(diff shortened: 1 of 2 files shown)


DhananjayThomble:
## Independent review

The spawned reviewer has stalled without a VERDICT on every PR in this repo (no fs/shell tools), so this is the in-session review per the standing workflow. Reviewed against the real files rather than my own reasoning: `infra/deploy.sh`, `infra/lib/snapurl-stack.ts` outputs, `infra/cdk.json`, `Dockerfile` stages, and all four existing workflows.

### Findings — 2 real, both fixed in 2bc30a0

**1. Missing variables silently delete infrastructure (high).** `deploy.sh` appends each `-c` flag only when its env var is non-empty. An **unset** `BUDGET_EMAIL` or `GOOGLE_OAUTH_CLIENT_ID` therefore does not error — it omits the flag, and CDK removes the budget/SNS cost alarms or strips OAuth from the API Lambda. This exact class has been caught twice by a manual diff. The `plan` job annotates destructive diffs, but that depends on a human noticing. Since a missing variable is never an intent, both `plan` and `deploy` now hard-fail on any of the four being empty. Checked in **both** jobs because environment-scoped variables can resolve in `deploy` but not in `plan`.

**2. The plan job could not have authenticated (blocking).** My first draft pointed `plan` at the deploy role, whose trust policy is bound to `environment:production`. `plan` deliberately declares no environment — it must run before the approval — so its OIDC subject is the branch ref and the assume-role would have failed on the very first run. I had written this up as a caveat in the doc instead of fixing it, which was wrong. Resolved properly with a second read-only role (`snapurl-github-plan`) that can assume only CDK's read-only lookup role. This is also the better least-privilege answer than widening the deploy role's trusted subjects, which would have let anyone who can push a branch deploy without approval.

### Verified, not assumed

- Reusable-workflow contract: `smoke-deployed.yml` requires `redirect_base_url` + `api_base_url`; both are passed, no unknown extras (checked programmatically against its `workflow_call.inputs`).
- `ApiUrl` is `httpApi.apiEndpoint` — the API Gateway root with **no** path — so `/api/v1` is appended, which is what the smoke script's `API` expects. `RedirectDomain` already includes `https://`, so it is passed through unmodified.
- Required status check queried live: `["CI gate"]`. It aggregates only `verify.yml`'s jobs, so this PR adds no required check and branch protection needs no change.
- Setup steps match the repo convention exactly (`pnpm/action-setup` before `setup-node` with `cache: pnpm`, node 22, `pnpm install --frozen-lockfile`).
- The guard's `${!v}` indirect expansion was executed both ways rather than eyeballed: all-set passes, missing correctly reports `DOMAIN_NAME BUDGET_EMAIL`.
- Two stacks exist (`SnapUrl` in ap-south-1, `SnapUrlCert` in us-east-1, joined by a cross-region cert reference), so the role policy grants bootstrap-role assumption in **both** regions.

### Accepted trade-offs, called out rather than hidden

- **Images build twice** (one synth per job). This is the price of putting the approval between a readable diff and the apply; native arm64 makes it cheap and CDK skips pushing existing digests.
- **Two approval prompts** when `run_migrations` is ticked — environment protection is per job and `migrate` must declare the environment to get credentials at all. Documented.
- **Major-tag action pinning** (`@v4`) for consistency with the four existing workflows. SHA pinning is stricter for the one workflow that holds AWS credentials; flagged in the doc as a deliberate open choice rather than silently adopted.

### Not verifiable before merge

The workflow cannot execute until the two IAM roles and the `production` environment exist. Specifically unproven until a first real run: that `ubuntu-24.04-arm` is available to this repo, and that CDK's lookup role alone is sufficient for `cdk diff` (if it is not, the plan role needs `cloudformation:GetTemplate` added). Nothing was deployed and no live AWS resource was touched by this PR.

Issue #348, opened 2026-09-06
[infra] Pin CDK context flags so a deploy cannot silently revert live infrastructure


DhananjayThomble:
## Problem

`npx cdk deploy` for this stack takes its entire configuration from `-c` context flags, and **CDK does not remember them between invocations**. Omitting one does not error — it redeploys that feature to its default and tears down live infrastructure.

This has already broken production twice. In the most recent incident a deploy that omitted the flags carried by another branch reverted two fixes at once:
- the NAT instance security group's ingress rule went back to `[]`, killing **all** Lambda egress (`ETIMEDOUT` reaching Secrets Manager)
- `DATABASE_SSL_NO_VERIFY` disappeared, so the RDS TLS handshake failed with `SELF_SIGNED_CERT_IN_CHAIN`

Both are silent at deploy time; the stack reports success.

## Fix

Remove the class of error rather than documenting around it. Either:
- a committed `context` block in `infra/cdk.json` holding the non-secret defaults, or
- a checked-in `infra/deploy.sh` wrapper that always passes the full flag set and fails loudly if a required one is unset

Secrets and account-specific values must **not** be committed — those stay as env vars or CLI overrides.

Additionally add a post-deploy verification step that asserts the two settings above are actually present on the deployed resources, so a silent revert fails loudly.

## Acceptance criteria

- [ ] A deploy cannot be run without the required context being applied
- [ ] No secret or account identifier is committed to the repo
- [ ] `infra` type-check and `cdk synth` still pass
- [ ] `docs/DEPLOYMENT.md` updated to describe the single supported deploy command

### Constraints for an automated agent
- Do **not** modify `.github/**`, `.kiro/**`, `package.json`, lockfiles, or any `*.config.*` unless this issue explicitly says to.
- Never weaken, skip, or delete an existing assertion to make a build pass.
- Import request/response shapes from `@snapurl/contract`; do not hand-roll payload literals.
- DB-backed tests use the existing `const describeDb = DATABASE_URL ? describe : describe.skip` idiom.
- If the task cannot be completed without touching a forbidden path, stop and comment on this issue instead of widening scope.

DhananjayThomble:
## Decision: use a committed `context` block in `infra/cdk.json`

Chosen over a `deploy.sh` wrapper. Implement it that way.

### What goes in the committed context block
Only non-secret, environment-independent defaults, e.g. the NAT strategy and the
public domain name. These are the values whose omission silently changes
[...]

PR #400, opened 2026-09-13
ci(deploy): rollback, unconditional migrations, drift detection

DhananjayThomble:
Closes the recovery gaps the deploy pipeline in #399 shipped without. Everything here is workflow/docs/infra-output only — no application code, no topology change, **no new AWS spend**.

Sequenced ahead of a staging environment deliberately: staging needs a second RDS plus a NAT instance (`natStrategy=none` won't do, the Lambdas need Secrets Manager egress — that was the #348 incident), hits a real blocker (`budgetName: "SnapURL-monthly-cost"` is account-global and would collide), and needs product decisions. This costs nothing and closes the sharper risk.

## 1. Record the deployed commit

Nothing could answer *"which commit is production on?"* — which makes "redeploy the last good commit" unactionable. Adds an **optional** `deployedGitSha` prop threaded through `-c` context, surfaced as a `DeployedGitSha` output. CI passes the resolved SHA.

Verified additive rather than asserted: synthesizing with and without the flag yields **byte-identical `Resources` (80) and `Parameters` (10)**, the only delta being the new output — so omitting it reproduces the previous template exactly, and no logical ID moved.

## 2. Rollback

Nothing could undo a deploy. CloudFormation auto-rollback covers a *failed* update; it does nothing for one that **succeeds and is functionally broken** — which is exactly what the redirect-403 outage was (CFN reported success, health checks 200, zero Lambda errors, empty DLQ).

- New `git_ref` input. **Both** `plan` and `deploy` check out the same explicit ref, defaulting to `github.sha` rather than the branch tip, so main advancing mid-run cannot make the applied change differ from the reviewed diff.
- The deploy now resolves the actually-checked-out commit (`git rev-parse HEAD`) rather than reusing `github.sha`, which would be wrong during a rollback.
- **The image check is deliberately NOT a hard gate**, and I want to flag that I changed my own plan here. A CDK asset hash is derived from its build *context*, so an older ref reproduces the same hash, and if the image is absent from ECR, CDK **rebuilds it from source** rather than failing. A missing image cannot break a rollback — only slow it. So the preflight reports fast-path (cached) vs slow-path (rebuild), which is the genuinely useful thing mid-incident. Shipping it as a "safety gate" would have been theatre.

`docs/ROLLBACK.md` leads with what rollback does **not** undo: migrations are forward-only, so backing out code leaves the new schema under old code. There's a table of which migration shapes make a code rollback safe versus actively worse.

Rollback feasibility was verified, not assumed: the CDK assets ECR repo's lifecycle policy expires only **untagged** images after 365 days, and asset images are digest-tagged — all 15 current digests are retained.

## 3. Migrations always run

Removed the `run_migrations` checkbox. Correctness must not depend on an operator remembering to tick a box — forget it on a release needing a new column and production breaks in a way the smoke gate may not surface.

Safe because `packages/database/src/migrate.ts` uses Drizzle's migrator, which records applied migrations in its own table and applies only pending ones, so it no-ops. The workflow documents the consequence: the order is now "new code, then new schema", so migrations must stay expand/contract.

This also simplified the smoke gate's condition — `always()` with explicit result checks existed only to tolerate a skipped migrate job, which can no longer happen.

## 4. Weekly drift detection

New read-only `drift-detection.yml`. This account has carried infrastructure CloudFormation didn't know about: during the 403 incident a `lambda:InvokeFunction` grant was added by hand and existed outside the template until it was made permanent in CDK. Nothing would have told us.

Uses the **plan** role, not the deploy role — a scheduled job that could deploy is a standing risk with no upside. Reports drifted resources by name and fails the run; it does **not** attempt to correct drift, since deciding whether the template or reality is right is a judgement call.

## Verification

- **`actionlint` clean on all 6 workflows.** It earned its place: it caught a real YAML break where embedded multi-line Python starting at column 0 terminated a block scalar. Replaced with a one-line `jq` extraction, then tested against both the real 3-image manifest and the image-less cert manifest.
- `cdk synth` succeeds with **and** without `deployedGitSha`, templates diffed structurally.
- `pnpm -r type-check` clean; infra 79 tests pass.
- Every embedded bash block executed locally against **both** the pass and fail case — the drift poll (completes / never completes), and the drift branch (IN_SYNC exit 0 / DRIFTED exit 1). My first poll-loop test was itself wrong (subshell counter never incremented, reporting "0 polls"); I rewrote the harness rather than accepting the green.

## Not verified

Nothing was deployed and no live AWS resource was mutated. Untested until a real run: the ECR preflight against a genuine rollback, and drift detection end to end (needs the two new permissions on the plan role — `docs/CD-AWS-OIDC.md` is updated with `ecr:DescribeImages` and the four `cloudformation:*Drift*` actions).

Still open from the architecture review, in order: staging environment, then build-once-promote-many, then alarm-gated canary rollout. The canary gap is stated explicitly at the bottom of `ROLLBACK.md` — this shortens time-to-recovery, it does not stop users seeing the problem.

### DIFF
--- a/.github/workflows/deploy-aws.yml
+++ b/.github/workflows/deploy-aws.yml
@@ -60,11 +60,15 @@ on:
         required: true
         type: string
         default: SnapUrl
-      run_migrations:
-        description: Invoke the worker with {"task":"migrate"} after deploying.
+      git_ref:
+        description: >-
+          Commit, tag or branch to deploy. Blank = the ref this workflow was
+          dispatched from. Set it to an earlier commit to ROLL BACK — read
+          docs/ROLLBACK.md first, because redeploying old code does NOT undo a
+          database migration.
         required: false
-        type: boolean
-        default: false
+        type: string
+        default: ""
       skip_smoke:
         description: "Skip the post-deploy smoke gate. Leave false: unchecked means the deploy must prove itself."
         required: false
@@ -106,7 +110,13 @@ jobs:
       contents: read
       id-token: write # to assume the AWS role via OIDC
     steps:
+      # Pin to an explicit commit rather than the branch tip, and use the SAME
+      # one in the plan and deploy jobs. Two reasons: main advancing between the
+      # two jobs must not make the applied change differ from the reviewed diff,
+      # and `git_ref` needs a single place to take effect for a rollback.
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.git_ref || github.sha }}
 
       - uses: pnpm/action-setup@v4
         with:
@@ -202,6 +212,49 @@ jobs:
             echo "::warning title=Destructive change in plan::This diff removes or replaces resources. Confirm every one is intended before approving."
           fi
 
+      # Informational, NOT a gate — and the distinction matters, so it is stated
+      # rather than implied. A CDK image asset's hash is derived from its build
+      # CONTEXT, so an older commit reproduces the same hash; if that image is
+      # absent from ECR, CDK rebuilds it from the checked-out source rather than
+      # failing. So a missing image cannot break a rollback, it can only make it
+      # slow (a full emulation-free but still multi-minute arm64 build) instead
+      # of near-instant. During an incident, knowing which of those you are in
+      # for is worth a few seconds; pretending it is a safety check would not be.
+      - name: Rollback preflight — are the target images already in ECR?
+        env:
+          AWS_ACCOUNT_ID: ${{ vars.AWS_ACCOUNT_ID }}
+        run: |
+          set -uo pipefail
+          repo="cdk-hnb659fds-container-assets-${AWS_ACCOUNT_ID}-${AWS_REGION}"
+          manifest="infra/cdk.out/${{ inputs.stack }}.assets.json"
+          if [ ! -f "$manifest" ]; then
+            echo "No assets manifest at $manifest — skipping preflight."
+            exit 0
+          fi
+          tags="$(jq -r '(.dockerImages // {}) | to_entries[] | (.value.destinations | to_entries[0].value.imageTag)' "$manifest")"
+          present=0
+          absent=0
+          for tag in $tags; do
+            if aws ecr describe-images --repository-name "$repo" --image-ids "imageTag=$tag" >/dev/null 2>&1; then
+              echo "  cached in ECR : $tag"
+              present=$((present+1))
+            else
+              echo "  needs rebuild : $tag"
+              absent=$((absent+1))
+            fi
+          done
+          {
+            echo "### Rollback preflight"
+            echo
+            echo "Deploying commit \`$(git rev-parse HEAD)\`."
+            echo
+            if [ "$absent" -eq 0 ]; then
+              echo "All $present image(s) already in ECR — the deploy skips the image build (fast path)."
+            else
+              echo "$absent of $((present+absent)) image(s) must be rebuilt — expect several extra minutes."
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
       - uses: actions/upload-artifact@v4
         with:
           name: cdk-diff-${{ github.run_id }}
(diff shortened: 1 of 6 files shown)


DhananjayThomble:
## Independent review

In-session per the standing workflow (the spawned reviewer has stalled without a VERDICT on every PR in this repo). Reviewed against the real files, and I ran things rather than reasoning about them.

### Findings — 2 real, both fixed

**1. YAML break in an embedded script (blocking).** My first version of the rollback preflight embedded multi-line Python via `python3 -c "` inside a `run: |` block. The Python lines started at column 0, which **terminates a YAML block scalar** — the workflow was unparseable. `actionlint` caught it at `deploy-aws.yml:243`; my own `yaml.safe_load` check caught it too. Replaced with a one-line `jq` extraction. This is the concrete argument for installing actionlint rather than eyeballing workflow YAML, and it is worth adding to CI (noted below).

**2. Hand-constructed ECR repository name and implicit region (fixed in 082ed76).** The preflight built `cdk-hnb659fds-container-assets-<acct>-<region>` by hand and relied on the ambient region. CDK records `repositoryName` **and** `region` per asset destination, and an asset can legitimately target a different region than the stack — this app already spans ap-south-1 and us-east-1. A wrong name silently reports *every* image as "needs rebuild": a false signal, delivered exactly when someone is reading it during an incident. Now parsed from the manifest with the `${AWS::AccountId}` placeholder substituted and `--region` passed explicitly.

### A design decision I reversed, deliberately

The task I set specified a **hard gate** failing the deploy if a target image was absent from ECR. Implementing it, I checked the premise and it does not hold: a CDK image asset's hash is derived from its build **context**, so an older ref reproduces the same hash, and if the image is missing CDK **rebuilds it from source** rather than failing. A missing image therefore cannot break a rollback — only slow it.

So the gate would have been theatre: a check that never fires for the reason claimed. Shipped as an informational preflight instead, reporting fast-path (cached) vs slow-path (rebuild), with the reasoning written into the step's comment so the next reader does not "fix" it back into a gate.

### Verified by execution, not inspection

- **`DeployedGitSha` is genuinely additive:** synth with and without the flag gives byte-identical `Resources` (80) and `Parameters` (10); the sole delta is the new output. So a local deploy omitting it reproduces the prior template exactly and no logical ID moved.
- **Rollback is actually possible:** the CDK assets ECR repo's lifecycle policy expires only **untagged** images after 365 days; asset images are digest-tagged, and all 15 current digests are retained. Without this, `ROLLBACK.md` would be describing something that does not work.
- **Migrations are safe to always run:** read `packages/database/src/migrate.ts` — Drizzle's migrator tracks applied migrations in its own table and applies only pending ones. Confirmed before removing the checkbox, not assumed from the function name.
- **Every embedded bash block, both outcomes.** Drift poll: completes after N polls, and gives up rather than spinning when it never completes. Drift branch: `IN_SYNC` → exit 0, `DRIFTED` → exit 1 with annotation. Preflight: real 3-image manifest and the image-less cert manifest.
- My **first** poll-loop test was itself wrong — the counter lived in a subshell so it reported "0 polls" and the loop looked broken. I rewrote the harness rather than accepting or dismissing the green.
- `actionlint` clean on all 6 workflows; `pnpm -r type-check` clean; infra 79 tests pass.

### Judgement calls worth a second opinion

- **Two approval prompts per run.** Unchanged from #399, but migrations now always run, so every deploy *and every rollback* prompts twice. Structurally required (the deploy role trusts only `environment:production`, so `migrate` must declare it to get credentials). Mildly annoying during an incident; the alternative is a second environment without reviewers, which weakens the boundary. Kept, documented.
- **`always()` removed from the smoke gate.** It existed only to tolerate a skipped `migrate`, which can no longer happen, so the default "all needs succeeded" is now exactly right. A failed migration correctly blocks the gate rather than letting a half-migrated environment report success.
- **Weekly, not daily, drift detection.** Drift is not urgent — it is a slow-burn correctness problem — and the cron is deliberately off-the-hour because scheduled workflows bunch and get delayed.

### Not verifiable before merge

Nothing was deployed; no live AWS resource was mutated. Untested until a real run: the preflight against a genuine rollback, and drift detection end to end, which needs two new plan-role permissions (`ecr:DescribeImages`, four `cloudformation:*Drift*`) documented in `docs/CD-AWS-OIDC.md`. `DETECTION_FAILED` on specific resources would indicate a missing per-service read permission rather than a workflow bug.

### Recommended follow-up (not in this PR)

Add `actionlint` to `verify.yml`. It caught a workflow-breaking error here that no other check in this repo would have — `verify.yml` does not lint workflow YAML at all, so a malformed workflow can currently reach main.

Issue #357, opened 2026-09-06
[test-gap] CloudFront function tests pass while the deployed function cannot load


DhananjayThomble:
## What happened

The CloudFront viewer-request function ended with `export { handler };`. CloudFront Functions reject that:

```
SyntaxError: Illegal export statement in 130
```

**An invalid CloudFront Function does not fall through to the origin.** The distribution answers *every* request with a 503. One line took the whole redirect path down.

It went unnoticed for weeks because the redirect domain had no DNS record, so nothing ever reached the distribution. It surfaced within minutes of the record being added.

Fixed in the accompanying PR by deleting the export.

## Why the tests did not catch it

`infra/functions/redirect-viewer-request.test.ts` has 14 passing tests, and they all passed while the deployed function was fatally broken.

The drift guard asserts that the shared logic region in `redirect-viewer-request.js` is byte-for-byte identical to its twin in `redirect-viewer-request.logic.mjs`. That is a good invariant and it worked as designed — but it only covers the **logic region**. The module scaffolding around it (the import, the handler declaration, the export) is outside the guarded region and is never validated.

So the suite verifies the function computes the right answer, and never verifies the file is a **loadable CloudFront Function**.

## What to add

A test that validates the deployed artifact itself, not just its logic:

- [ ] Assert the file contains **no** `export` / `module.exports` statement
- [ ] Assert a global `async function handler(` is declared
- [ ] Assert the only `import` is the permitted `cloudfront` module
- [ ] Assert the file is within the CloudFront Function size limit

Consider also a CI step running `aws cloudfront test-function` against the synthesized code. That is the only check that would have caught this with certainty, since it exercises the real runtime — but it needs AWS credentials, so it belongs on a deploy-gated or nightly job, not on the per-PR path.

## Broader point

This is the same class as the `.env` / silent-revert issue in #348: **a check that passes while the real thing is broken.** Worth calling out in the QA program's rubric — a test that cannot fail when production is down is not covering production.

### Constraints for an automated agent
- Do **not** modify `.github/**`, `.kiro/**`, `package.json`, lockfiles, or `*.config.*` unless the issue says to.
- Never weaken, skip, or delete an existing assertion.
- Do not alter the DRIFT-GUARDED REGION or its twin; this issue is about the scaffolding around it.

PR #402, opened 2026-09-13
fix(infra): fit the edge function in CloudFront's 10 KB limit

DhananjayThomble:
**The #395 edge query merge has been undeployable since the day it merged.** Today's production deploy ([run 34758696282](https://github.com/DhananjayThomble/URL-Shortener-App/actions/runs/34758696282)) failed on it:

```
RedirectViewerRequest  AWS::CloudFront::Function  UPDATE_FAILED
Internal error reported from downstream service during operation
'(Service: CloudFront, Status Code: 413, ...)'
```

`413` is Payload Too Large. CloudFront caps a function's code at 10 KB (10240 bytes) and the file had reached **11,336**. CloudFormation auto-rolled back and production was unaffected — redirects verified 302 with query forwarding intact, API health ok — but the fix could never have shipped.

## The real defect is the missing check

79 unit tests, `cdk synth`, `cdk diff` and the CI gate all passed on an artifact AWS refuses to accept. The byte count is the symptom; the absence of any check is the defect, and it's the same class as #357 (an illegal `export` would break the function while the logic tests stayed green).

## The fix

Of the 11,336 bytes, only **3,837 were code** — **7,499 were comments**. So spend the budget on code:

| file | before | after |
|---|---|---|
| `redirect-viewer-request.js` | 11,336 | **7,690** (2,550 headroom) |
| `redirect-viewer-request.logic.mjs` | 8,852 | 7,088 |

The full rationale moves to a new `infra/functions/README.md` — why the encoder must match `URLSearchParams` byte-for-byte, the deliberate lone-surrogate divergence, why an integer-like query key declines, why there is no `export`, and the two-file drift-guard design. **None of it is lost.** Several of those points were found by a failing test or a broken deploy rather than by reading code, so they're worth more than the bytes they cost — just not at the edge. Terse pointers remain in the function.

## Verified, not assumed

- **The code is byte-identical to `origin/main` once comments are stripped** — 3,762 bytes on both sides, asserted mechanically. This is a comments-only change with no behaviour difference.
- **The drift-guarded region was re-spliced** from the `.js` into the `.mjs` with a script, so byte-identity holds by construction rather than by hand-editing both sides and hoping.
- **The new size guard was proven to fail.** I padded the file over the limit and confirmed it reports `redirect-viewer-request.js is 11098 bytes; CloudFront's limit is 10240. Over by 858. Move explanation to infra/functions/README.md.`, then restored and confirmed green. A guard nobody has watched fail isn't a guard.
- infra **80 tests** pass (79 + the guard).

It lives in the existing `#357` *"properties of the DEPLOYED file"* suite, which is exactly where a future reader would look for it.

## Follow-up worth noting

This is a concrete instance of the argument for a **staging environment** — a change that is correct, tested, reviewed and undeployable, discovered only by attempting production. It cost nothing this time because CloudFormation's auto-rollback worked. It won't always.

Also worth recording: a failed CDK deploy does **not** always mean "nothing was applied". That holds only for a failure at the image build or ECR push, before the change set runs. This one got past that, so the stack really did update and roll back (`UPDATE_ROLLBACK_COMPLETE`, `LastUpdatedTime` moved).


DhananjayThomble:
## Independent review

In-session per the standing workflow (the spawned reviewer has stalled without a VERDICT on every PR in this repo).

The risk in this PR is not the size reduction — it is that a "comments-only" edit to a **drift-guarded, unexported, untypechecked** file could silently alter behaviour, and the edge function is the one component where a one-byte divergence produces a cache-dependent, near-undebuggable inconsistency between the edge and the Lambda. So I checked exactly that rather than reading for plausibility.

### The load-bearing assertion

**Comments stripped, the code is byte-identical to `origin/main`: 3,762 bytes on both sides.** Done mechanically (strip block comments, whole-line and trailing `//`, collapse blank lines, compare) rather than by eyeballing the diff, because the diff is 133 changed lines and eyeballing it proves nothing. If that comparison had differed by a byte I would have treated this as a behaviour change requiring the full differential suite to be re-reasoned.

### Verified

- **Drift guard holds, and by construction.** The region was re-spliced from the `.js` into the `.mjs` with a script, then the test's own extraction was re-run to confirm `region(twin) === region(runtime)`. Hand-editing both sides would have been the obvious way to introduce exactly the drift this guard exists to catch.
- **The new size guard actually fails.** Padded the file to 11,098 bytes, confirmed the assertion fires with `Over by 858. Move explanation to infra/functions/README.md.`, restored, confirmed 80/80 green. I do not count an unwatched guard as a guard — that is the specific mistake that let 11,336 bytes reach `main`.
- **The `.mjs` still exports what the test imports** (`decide, edgeKey, edgeLocation, formEncode, serialiseParams`) — the splice touches only the region, but the export line sits outside it and would break every logic test if disturbed.
- Sizes: 7,690 bytes, **2,550 under the limit**. Not so tight that the next legitimate comment breaks the build.
- The remaining terse comments preserve every *actionable* warning (the six substitutions ARE the whole difference; the integer-key decline; the surrogate throw becoming a fall-through; no `export`), each pointing at the README for the reasoning.

### Judgement calls

- **README over stripping comments at synth time.** A build step that strips comments would keep the prose in the source and satisfy the limit, but it breaks the drift guard's premise that the file on disk is the file deployed, and adds machinery to the one place that must stay boring. Moving the prose is lower-risk and leaves the artifact honest.
- **Limit asserted at exactly 10 × 1024**, not a padded threshold. A guard should encode the real constraint; an artificial margin invites someone to raise it.
- **`infra/functions/README.md` rather than `docs/`.** It sits beside the two files it describes, which is where someone editing them will actually look.

### Not verified

The trimmed function has not been deployed — that is the point of the re-dispatch this unblocks. A deploy is the only thing that proves CloudFront accepts it, and while 7,690 < 10,240 makes a 413 very unlikely, I have not seen AWS accept this exact artifact. The verification that matters (KVS `ItemCount` rising above 0, and the edge-served `Location` matching `buildDestination` byte-for-byte with no redirect-Lambda invocation) still cannot run until it deploys.

PR #404, opened 2026-09-13
ci: lint workflow YAML with actionlint in the CI gate

DhananjayThomble:
## Lint the workflows in CI (last CI-hardening item)

The one gap the CD work exposed: nothing in the repo checked the GitHub Actions workflows themselves. The app jobs test the application, but a malformed workflow could reach `main` and break only at dispatch time. Concretely, while building the CD pipeline an embedded multi-line script starting at column 0 silently terminated a `run: |` block scalar — a real, workflow-breaking mistake that `yaml.safe_load`, every repo test, and eyeballing all passed, and only `actionlint` caught.

### What this adds
- A new `actionlint` job that parses every workflow, type-checks `${{ }}` expressions and `needs` graphs, and runs **shellcheck** over `run:` scripts.
- It uses the pinned `docker://rhysd/actionlint:1.7.12` container, which bundles actionlint + shellcheck + pyflakes — **no toolchain setup step** needed.
- A new `workflows` path filter on the `changes` job so it runs only when a workflow file changes (and always on push to `main`).
- Added to `ci-gate`'s `needs`, so a workflow lint failure now **blocks merge**. `ci-gate` already tolerates a skip, so unrelated PRs stay green.

### Verification
- Ran actionlint against all current workflows locally: **0 errors** (this new job included).
- Confirmed it catches a deliberately-broken workflow (duplicate `jobs:` key) — the guard is real, not decorative.
- The bundled shellcheck/pyflakes are disabled only in a bare-binary run; the container image ships them, so CI gets the stronger check than a local binary would.
- This PR itself touches a workflow, so the new `actionlint` job runs on it — a live green proof.

No behavior change to any existing job.


DhananjayThomble:
## Independent review (in-session; spawned reviewer path unavailable)

**VERDICT: PASS**

- **Wiring is correct.** The new `workflows` filter output is declared on `changes`, defined in the paths-filter, and the push-to-main override sets it alongside `code`, so the default branch always lints workflows. `actionlint` is added to `ci-gate`'s `needs`, so a lint failure blocks merge; `ci-gate`'s existing success-or-skipped logic means an unrelated PR (job skipped) stays green. No existing job changed.
- **Tool choice is sound.** The pinned `docker://rhysd/actionlint:1.7.12` container bundles shellcheck + pyflakes, so CI gets `run:`-script linting that a bare binary on the runner would silently skip (verified locally: shellcheck/pyflakes were disabled in the binary run). Pinned to an exact release, not a floating tag.
- **Proven, not asserted.** actionlint runs clean over all current workflows (0 errors, including this new job), and catches a deliberately-broken workflow. This PR touches a workflow, so the job runs on the PR itself — a live green.
- **Non-blocking note:** the container action pins by tag, not digest. Consistent with the repo's other `@v4` action pins, so not a regression; a future hardening pass could pin all actions by SHA.

PR #405, opened 2026-09-13
fix(mobile): reachable scroll affordance for wide tables (DC3)

DhananjayThomble:
## What & why (DC3, MAJOR — mobile responsiveness audit)

On `/team`, below the `sm` breakpoint the **Members** table and the **role permission matrix** are wider than a phone viewport (the shared `<Table/>` primitive carries a `min-width`). Their right-hand columns — Members' **Links** column and the matrix's **Admin / Owner** columns — were **silently clipped** off the right edge with no cue that more content existed.

`TableWrap` already wrapped these in `overflow-x-auto`, so the columns were technically scrollable but **undiscoverable** on a phone (no scrollbar, no affordance). That's the genuine defect.

## Approach — `overflow-x-auto` + a visible mobile affordance (not card-reflow)

I chose **keep the horizontal scroller, add a discoverability affordance** over card-reflow:

- Card-reflow would mean restructuring every table's markup across the **7 pages** that use `TableWrap` (team, conversions, developers, domains, reports, bio, forms) — high blast radius for a MAJOR-but-not-catastrophe finding.
- The scroller already exists; the only real gap is that users can't tell it scrolls. So I made `TableWrap` overflow-aware and added, **mobile-only**, a right-edge fade + a `Scroll →` text hint (colour-cue is paired with text for a11y), shown **only while there is still content to the right** (measured with a resize/scroll observer, hidden at scroll-end).

## Desktop is untouched

- The overlay is `sm:hidden` **and** only mounts when the table actually overflows. On desktop these tables fit, so nothing renders and behaviour is byte-identical.
- Shared-component change is gated to **mobile + real overflow**, so the other five `TableWrap` pages get the same benefit on phones and regress on none.

## Tests

- New `e2e/tests/team-mobile-tables.spec.ts` (390px iPhone-class viewport, fixtures mode, `test.use` viewport override so the global desktop project is unchanged): proves each wide-table wrapper is horizontally scrollable, the `Scroll →` affordance shows while columns are off-screen, and the right-most **Links / Admin / Owner** columns are **reachable** (scroll into view) — i.e. no data is hidden. **2/2 pass.**
- Existing `e2e/tests/team.spec.ts` re-run green (**3/3**) — no regression from the shared `TableWrap` change.
- `tsc --noEmit` clean; `next build` clean.

Files: `web/src/components/ui/index.tsx` (TableWrap), `e2e/tests/team-mobile-tables.spec.ts`.


DhananjayThomble (commented):
## Independent review (in-session, single cycle)

Reviewed the full diff of both files. The spawned-reviewer path stalls on this repo, so per the SnapURL PR workflow this is the one independent review done in-session.

### `web/src/components/ui/index.tsx` — TableWrap
- **Overflow math is correct.** `scrollWidth - clientWidth > 1` for "can scroll"; `scrollLeft >= scrollWidth - clientWidth - 1` for "at end", both with a 1px tolerance to avoid sub-pixel flapping. ✓
- **Re-measures on every relevant event:** initial synchronous `measure()` in the effect, a passive `scroll` listener, a `ResizeObserver` on the scroller, and a `window resize` listener; all cleaned up in the return. Deps `[children]` re-arm when the table's rows change. ✓
- **No SSR/hydration mismatch:** `overflow`/`atEnd` default to `false`, so the server render and the first client render both emit no overlay; the effect flips state only after mount. `ResizeObserver` is `typeof`-guarded. ✓
- **Desktop byte-identical:** the affordance is `sm:hidden` AND only mounts when the table genuinely overflows — on desktop these tables fit, so nothing renders. Verified empirically: `team.spec.ts` (desktop) still 3/3. ✓
- **a11y:** overlay nodes are `aria-hidden` + `pointer-events-none` (don't intercept touch, don't pollute the a11y tree); the colour fade is paired with a text `Scroll →` hint so the cue is not colour-only. ✓
- **Blast radius acknowledged and bounded:** shared component used by 7 pages; the change is gated to mobile + real overflow, so it's additive for phones and inert for the desktop layouts the other pages already ship.

### `e2e/tests/team-mobile-tables.spec.ts`
- 390px viewport via `test.use` (does not add a project / touch the global desktop config). Fixtures mode + seedSession match `team.spec.ts`. ✓
- Waits past the Members skeleton before asserting (avoids the known SnapURL table `isLoading` race). ✓
- Asserts the real defect and the real fix: wrapper is scrollable, `Scroll →` shows while off-screen, and Links/Admin/Owner are reachable via `scrollIntoViewIfNeeded` (start off-screen → become in-viewport). Role/accessible-name selectors only. ✓

### Nits (non-blocking)
- The fade uses `from-surface`; if a table ever sat on a non-`surface` background the fade seam could differ slightly. All current TableWrap usages sit inside `Card` (surface), so fine today.

**VERDICT: PASS.** No blocking issues. No changes requested.

PR #406, opened 2026-09-13
fix(web): mobile navigation drawer below lg (DC1) + de-emphasize mobile search stub (DC5)

DhananjayThomble:
## What & why

Mobile-responsiveness audit finding **DC1 (CATASTROPHE)**: the authenticated app shell hid **all** primary navigation below the Tailwind \`lg\` breakpoint (1024px). The sidebar is \`<aside className="hidden lg:flex …">\` and the top bar carried only \`＋\` / a search stub / the avatar, so a phone user who landed on \`/links\` could not reach any other section (Analytics, QR, Bio, Forms, Conversions, Domains, Reports, Developers, Team, Settings). Confirmed on 11/11 authenticated pages on both Pixel 7 and iPhone 15 during the audit.

## Change

- **DC1 fix** — add a \`lg:hidden\` hamburger button in the \`Topbar\` that opens a left drawer. The drawer **reuses the exact same \`NAV\` array** as the desktop sidebar (extracted into shared \`NavLinks\` + \`QuotaMeter\` sub-components), so the two can never drift. Drawer behaviour: closes on navigation, on the close (✕) button, on backdrop tap and on \`Escape\`; locks body scroll while open; 44×44 touch targets on the open/close controls.
- The desktop \`<aside className="hidden lg:flex">\` sidebar is **untouched** and renders byte-identically at \`lg\` and up. \`Topbar\` now takes a \`counts\` prop (threaded from the \`(app)\` layout) to populate the drawer's per-section counts, matching the sidebar.
- **DC5 fix (minor, folded in, low-risk)** — the non-functional search stub dominated the mobile top bar and invited dead taps. Gated to \`hidden sm:flex\` so it no longer shows on phones; desktop unchanged (\`sm:\` applies). Wiring real search is out of scope and intentionally left.

## Desktop preservation

All additions are gated at mobile breakpoints (\`lg:hidden\` for the whole mobile nav; \`hidden sm:flex\` for the search stub). At \`lg\`+ the mobile nav renders nothing and the sidebar takes over; the \`Topbar\` padding \`px-[14px] sm:px-[22px]\` resolves to \`px-[22px]\` at desktop = the previous value.

## Tests

New \`e2e/tests/mobile-nav.spec.ts\` (fixtures mode, accessible-name selectors, \`seedSession\`):
- **mobile 390×852** — hamburger opens a drawer that exposes **every** NAV destination (a contract list asserts the full set, not a subset);
- **mobile** — tapping a destination navigates and closes the drawer, proven repeatably across two hops;
- **mobile** — the close button dismisses without navigating (negative case);
- **desktop 1280×900** — the hamburger is hidden and the sidebar shows its links.

Verified locally: \`type-check\` clean, \`next build\` green, new spec **4/4 pass**, and existing shell-dependent specs (create-link, analytics, settings ×2) still pass — no regression.


DhananjayThomble (commented):
## Independent review — VERDICT: PASS

Reviewed the full diff for correctness, the desktop-preservation constraint, and a11y.

**Desktop byte-identity (hard constraint) — holds.**
- The `<aside className="hidden lg:flex …">` sidebar markup is unchanged; `NavLinks`/`QuotaMeter` were extracted but render identical DOM (same class strings, same structure; `onNavigate` is `undefined` for the sidebar so no attribute change).
- `Topbar` padding `px-[22px]` → `px-[14px] sm:px-[22px]` resolves to `px-[22px]` at ≥sm (desktop) — identical.
- Search stub `flex` → `hidden sm:flex` resolves to `flex` at ≥sm — identical on desktop; hidden only on phones (DC5).
- `MobileNav` is `lg:hidden`, so nothing new renders at lg+.
- The desktop e2e case confirms the sidebar shows and the hamburger is hidden at 1280px.

**Correctness.**
- The mobile drawer reuses the **same `NAV` array** as the sidebar via the shared `NavLinks`, so the two cannot drift — the spec's contract list asserts the drawer exposes all 11 destinations.
- Drawer closes on navigation (pathname effect), close button, backdrop tap, and Escape; body scroll is locked while open and restored on cleanup.
- `Topbar` now requires `counts`; the only call site (`(app)/layout.tsx`) is updated in this PR.
- No new dependencies, no API/backend changes.

**Verification.** type-check clean; `next build` green; new `mobile-nav.spec.ts` 4/4 pass; existing shell-dependent specs (create-link, analytics, settings ×2) still pass.

**Non-blocking nits (a11y polish, not required to merge):**
1. The drawer is `role="dialog" aria-modal="true"` but does not move initial focus into itself or trap Tab. It's fully usable (Escape/backdrop/all controls reachable), but a stricter modal would focus the close button on open and trap focus. Worth a follow-up if the team wants full modal semantics.
2. Consider a matching focus-return to the hamburger on close.

Neither nit blocks the DC1/DC5 fix. Approving.

PR #407, opened 2026-09-13
fix(mobile): raise interactive controls to >=44px touch targets (DC2, DC4)

DhananjayThomble:
## What

Mobile-responsiveness audit findings **DC2 (MAJOR)** and **DC4 (minor)**: primary interactive controls were well under the WCAG 2.5.8 (AA) 24px floor / 44px usability target below the `lg` breakpoint.

| Control | Before (mobile) | After (mobile) | Desktop |
|---|---|---|---|
| Topbar `＋` create button | ~13×20px | 44×44px | unchanged |
| Account-menu avatar | 29×29px | 44×44px tap area (visual 29px circle preserved) | unchanged |
| Links status filter pills | ~28-31px tall | ≥44px | unchanged |
| `Segmented` toggle options | ~26-30px tall | ≥44px | unchanged |
| Per-row actions (`Button size="sm"`: Change role / Remove / Edit) | ~28px tall | ≥44px | unchanged |
| Marketing header hamburger (DC4) | 38×38px | 44×44px | unchanged |

## How

- New shared `@utility tap-target` in `web/src/app/globals.css`: under `@media (max-width: 1023.98px)` (the same 1024px line where the app shell switches to its mobile layout) it enforces `min-width/min-height: 44px`. Applied to the create button, avatar, filter pills, `Segmented` options and the `Button` `sm` size (per-row actions).
- The globals `.flex > * / .grid > * { min-width: 0 }` reset outranks a single utility class, so the two **flex-child icon buttons** (topbar create, marketing hamburger) that assert width additionally carry explicit `max-lg:w-11 max-lg:h-11`, which beats the reset. All other controls only need height, which the utility covers.
- The avatar keeps its small visual circle: the 29px `bg-teal` bubble is now an inner `<span>`, the button is the 44px hit box.
- **Desktop is untouched** — every rule is gated below `lg`.

## Skeptic-discards left alone (per audit)

26×26 colour swatches (accepted dense-picker pattern) and 350-wide inputs at 39-40px (4px short) were explicitly discarded in the audit's skeptic pass and are **not** modified.

## Tests

- New `e2e/tests/mobile-touch-targets.spec.ts` — 390×852 mobile viewport (`isMobile`, `hasTouch`), fixtures mode, accessible-name selectors — asserts a rendered `boundingBox()` ≥44px for: the create button, avatar, a filter pill, a per-row team action, a `Segmented` option, and the marketing hamburger.
[...]

PR #409, opened 2026-09-14
feat(web): wire top-bar search as an accessible link combobox (#408)

### LINKED ISSUE #408: Wire the top-bar search (currently a de-emphasized stub)
## Context
Follow-up from the mobile-responsiveness audit (finding **DC5**, minor). The app-shell top bar shows a search input labelled *"Search links, slugs, destinations, tags"* that is a **non-functional stub** — it renders but has no handler/route behind it.

As part of PR #406 (DC1 mobile nav) the stub was **de-emphasized on mobile** (`hidden sm:flex`) so it no longer dominates the phone top bar and invites dead taps. It is still visible on `sm`+ / desktop, and still does nothing.

## Ask
Actually wire the search so it queries links/slugs/destinations/tags and navigates to results, OR remove it if search isn't on the near-term roadmap. This is a **feature**, not a bug — filing so the dead affordance is tracked rather than silently shipped.

## Pointers
- Component: `web/src/components/app-shell/index.tsx` (Topbar search input).
- There is no search route/handler today — wiring it means a new query path (fixtures + real API) and a results surface or inline dropdown.
- Keep the mobile de-emphasis (`hidden sm:flex`) unless the wired search is worth surfacing on phones too.

_Source: SnapURL mobile Design Critique, 2026-09-11. Sibling fixes already merged: DC1 #406, DC2/DC4 #407, DC3 #405._

DhananjayThomble:
## Summary
Wires the top-bar search input — previously a non-functional stub (placeholder text, no handler, no route) — into an accessible combobox that queries the workspace's links as you type and shows matches in a listbox directly under the input. DC5 follow-up.

Closes #408

## Fixtures-search decision
**Server-side search via the existing `useLinks({ search })` hook** (`GET /links?search=`). The contract already has `ListLinksQuery.search`, and both `LinksService.list` and the fixtures backend (`web/src/lib/api/fixtures.ts`, the `/links` handler filters slug/destination/title/comment by the term) already honor it — **verified before implementing**. So there is **no fixtures change and no client-side re-filtering**: the dropdown shows exactly what the links page would for the same query, just capped to 8 rows. This keeps fixtures↔real-API parity.

## Behavior
- Debounced ~250ms; only fetches when the query is non-empty.
- Each result row shows `domain/slug` + click count + destination. Navigates to the link's detail page `/links/:id` (same route the links list uses).
- Loading (`Searching…`), empty-state (`No matches`), and hidden-when-empty states.
- **A11y:** `role="combobox"` on the input with `aria-expanded` / `aria-controls` / `aria-activedescendant` / `aria-autocomplete="list"`; results are a `role="listbox"` of `role="option"` rows. **ArrowUp/Down** move the active option, **Enter** opens the highlighted one, **Esc** closes. Closes on **outside-click** and on **route change**. Click uses `mousedown` so the outside-close doesn't cancel the navigation.
- **Mobile:** keeps DC5's `hidden sm:flex` — search stays desktop/tablet only, not surfaced on phones. Desktop layout otherwise byte-identical (same `flex-1 max-w-[400px]` wrapper); the mobile nav drawer (#406) and tap-target sizing (#407) are untouched.

## Verification
- `pnpm run type-check` clean; `pnpm build` (fixtures mode) clean.
- New E2E `e2e/tests/topbar-search.spec.ts` (6 cases): matching query shows the result; click navigates to `/links/:id`; ArrowDown+Enter navigates; no-match shows empty state; clearing hides the dropdown; combobox absent on a 390px phone viewport.
- Full app-shell regression: `topbar-search` + `mobile-nav` + `mobile-touch-targets` = **19 passed**.

## Files
- `web/src/components/app-shell/index.tsx` — new `TopbarSearch` combobox, replaces the stub div; imports `useLinks`.
- `e2e/tests/topbar-search.spec.ts` — new spec.


DhananjayThomble:
## Independent review (in-session, single cycle)

Reviewer note: the repo's spawned-reviewer path stalls on this repo, so per the established workflow this is a rigorous in-session independent review of the diff.

**Correctness**
- Fixtures-search claim verified against source: `web/src/lib/api/fixtures.ts` `/links` handler reads `params.get("search")` and filters `[slug, destination, title, comment]` (lines ~864–875). `ListLinksQuery.search` exists in `packages/contract/src/link.ts`. So server-side search is real; no fixtures change needed. ✅
- Detail route confirmed as `/links/:id` — matches `LinkRow` (`NextLink href={\`/links/${link.id}\`}`). ✅
- Debounce uses a `setTimeout` cleared on each keystroke → one query per pause, not per keypress. ✅

**Blocker found & fixed in this cycle**
- The empty-query case called `useLinks(undefined)`, which still fires `GET /links` (full list) even with the dropdown closed — a superfluous fetch on non-`/links` pages. Fixed by adding an optional `{ enabled }` to `useLinks` (default `true`, every existing call site unchanged) and gating the top-bar call on a non-empty query. Commit `9d08640`. Re-verified: type-check clean, 6/6 search E2E green, 168/168 web unit tests (incl. 157 fixture-parity) green.

**A11y**
- `role="combobox"` + `aria-expanded`/`aria-controls`/`aria-activedescendant`/`aria-autocomplete="list"`; results `role="listbox"` of `role="option"`. ArrowUp/Down/Enter/Esc + outside-click + route-change close all handled. Click uses `mousedown` + `preventDefault` so the outside-close doesn't cancel navigation. ✅

**Regression**
- Desktop layout preserved (same `flex-1 max-w-[400px] hidden sm:flex` wrapper). Mobile stays `hidden sm:flex` (DC5). `mobile-nav` (#406) + `mobile-touch-targets` (#407) specs re-run green (19 passed together). ✅

**Non-blocking nits (not addressed, low value)**
- Loading label is `Searching…` shown only in the (rare) window before the debounced query resolves; acceptable.
- `⌘K` kbd hint is decorative — no global `⌘K` handler is wired (out of scope for #408, which is about wiring the input itself).

**VERDICT: PASS**

Issue #430, opened 2026-09-15
[test-gap] Bug-injection gate: suite catches 6/7 injected bugs; three blind spots found


DhananjayThomble:
Part of #418. This is the `.kiro/steering/qa-oracles.md` §4 gate — the check that says
whether a green suite means anything. Seven known bugs were injected one at a time, the
relevant suite run, then each reverted (`git diff` verified clean between every step;
nothing committed).

## Result: the suite is genuinely sensitive — 6 of 7 caught

| # | mutation | suite | result |
|---|---|---|---|
| 1 | `routing.ts` `matches()` — country compare `!==` → `===` | domain | **CAUGHT** (4 tests) |
| 2 | `routing.ts` `evaluateRouting()` returns fallback unconditionally | domain | **CAUGHT** (6 tests) |
| 3 | `slug.ts` `generateSlug()` returns a constant | domain | **partial** (1 test, incidental) |
| 4 | `hll.ts` register rank off-by-one | domain | **CAUGHT** (3 tests) |
| 5 | `jobs/rollup.ts` click rollup off-by-one | worker (real PG) | **CAUGHT** (8 tests) |
| 6 | `links.service.ts` — drop workspace-ownership clause on `getFrom()` | api | **MISSED** |
| 7 | `contract/link.ts` — remove `.pipe(HttpUrl)` from `destination` | contract | **CAUGHT** (1 test) |

That most injections fail loudly is the reassuring half, and it retires the "provisional"
caveat I attached to earlier results: these suites can go red.

## Blind spot 1 — `LinksService.get()` has no test at all (most serious)

Removing the workspace-ownership clause from `getFrom()` — the query behind
`GET /links/:id` — leaves **all 169 tests passing**.

The current code is correct (`links.service.ts:193` has
`.where(and(eq(links.id, id), eq(links.workspaceId, workspaceId)))`). The problem is
purely that nothing tests it: `service.get(` appears **zero times** in
`links.service.integration.test.ts`. The file's only cross-workspace assertion
("never returns another workspace's links", line 304) exercises `service.list()`.

So `GET /links/:id` — one of the four `:id` routes the controller test enumerates — has
no happy path, no not-found case, and no cross-workspace case. A tenancy regression
there would ship green.

- [ ] Add integration coverage for `LinksService.get()`: found, not-found, and
      another workspace's id → not returned
- [ ] Audit the other `:id` service methods (`clone`, `update`, `remove`) for the same
      shape of gap

[...]

PR #432, opened 2026-09-15
test: cover LinksService :id methods and generateSlug randomness

DhananjayThomble:
## What does this PR do?

Closes two of the three blind spots in #430. Test files only — **no production code
changed**.

The bug-injection gate proved both gaps by mutation rather than by guesswork:

**1. `LinksService` `:id` methods had no cross-workspace coverage.** Dropping the
`eq(links.workspaceId, workspaceId)` clause from `getFrom()` — the query behind
`GET /links/:id` — left all 169 API tests passing. `service.get(` appeared zero times in
the integration test; the file's only cross-workspace assertion exercised `list()`.

Added found / not-found / cross-workspace coverage for `get`, and cross-workspace
assertions for `clone` and `update` as well.

**2. `generateSlug()` had no randomness or uniqueness oracle.** Making it return a
constant failed only the incidental "generates the requested length" assertion. Added
`packages/domain/src/slug.generate.property.test.ts` (fast-check, seeded): N successive
calls are distinct, output is alphabet-legal and never contains look-alikes (`0O1lI`),
and the length invariant holds swept across 1..64.

## Mutation-proof — each new test was shown to go RED against its target bug

| mutation | result |
|---|---|
| drop `eq(links.workspaceId, workspaceId)` from `getFrom()` | `get`/`clone`/`update` cross-workspace tests **FAILED** (3 failed / 22 passed). Reverted → 25/25 green. |
| `generateSlug()` → `return "aaaaaaa"` | distinctness + length-sweep tests **FAILED** (3 failed / 45 passed). Reverted → 129/129 green. |

`remove` correctly stayed green under the first mutation — it has its own separate guard
select the mutation didn't touch.

A test that cannot fail against the bug it targets isn't a regression test, so this
proof is the point of the PR as much as the tests are.

## Type of change

- [x] Chore (refactoring code, workflow improvements)

## How should this be tested?

```bash
# API integration tests need a database
DATABASE_URL=postgres://snapurl:snapurl@localhost:5435/snapurl \
  pnpm --filter @snapurl/api test      # 17 files, 181 tests
pnpm --filter @snapurl/domain test     # 8 files, 129 tests
pnpm type-check                        # 12 projects
```

Uses the existing `describeDb = DATABASE_URL ? describe : describe.skip` idiom, so the
suite still skips cleanly without a database.

## Not covered (deliberate)

- `bulkCreate()` also calls `getFrom()`, but its ids come from its own insert — no
  caller-supplied id, so no cross-workspace input surface to test.
- The slug collision-retry path in `insert()`/`bulkCreate()` is different behaviour from
  `generateSlug`'s own randomness and needs DB setup; not a pure property test.
- No uniformity/chi-square test on character distribution — a flakier property that the
  constant-return mutation doesn't motivate.

Blind spot 3 from #430 (making the HttpUrl field check exhaustive) is **not** in this PR.


DhananjayThomble:
CI note: `coverage-delta` fails here, and it is not this PR's tests that are at fault.

Both coverage jobs run `pnpm test:coverage` with no `DATABASE_URL`, so the DB-gated integration suites skip during coverage measurement. The new `LinksService.get()` coverage therefore never runs, contributes no covered lines, and the test-only guard reports "did not raise covered lines in ANY file".

#433 fixes that by giving both `coverage-baseline` and `coverage-delta` a postgres service and migrations. Once #433 lands I'll rebase this PR and re-run — blocking on it rather than weakening the gate, per `.kiro/steering/testing.md` ("Do not weaken a gate to make a PR green — fix the code").

PR #454, opened 2026-09-16
fix(api): a conversion may only reference a link in the caller's workspace

DhananjayThomble:
## What does this PR do?

`ConversionsService.record` resolved a `slug` under a workspace-scoped query, but took a
caller-supplied `linkId` **on trust** and wrote it straight into the row:

```ts
let linkId = input.linkId ?? null;
if (!linkId && input.slug) {
  ...where(and(eq(links.workspaceId, workspaceId), eq(links.slug, input.slug)))  // scoped
}
// linkId path: inserted with no workspace check at all
```

So one workspace could book a conversion against another's link. That is more than a stray
row — the report `innerJoin`s `links` on `conversions.linkId`, so the attacker's own
report would surface the victim link's slug.

The two ways into the same field disagreed about who owns the link. Now they agree: a
supplied `linkId` is resolved under the same scope, and 400s when it does not.

Also constrains `RecordConversionInput.linkId` to a **uuid**. The value reaches a uuid
column, so a malformed one surfaced as a 500 from Postgres rather than a 400 naming the
field.

## Verification

**Mutation-proven.** Disabling the ownership check fails exactly the two tests that should
fail — *"refuses a linkId belonging to another workspace"* and *"refuses a linkId that does
not exist at all"* — while the own-workspace happy path stays green.

The test seeds two real workspaces with their own links and drives the service against a
real Postgres, so it exercises the actual query rather than a mock.

`pnpm --filter @snapurl/api test` — 200/200 · `pnpm type-check` clean.

## Note on the oracle

The correct behaviour was not invented for this fix — the `slug` branch has always been
workspace-scoped. The bug was that the second path into the same field did not follow the
rule the first one already established.

## Type of change


DhananjayThomble:
## Review — approved

Verified locally in a clean worktree at `c5e522a`, against a real Postgres 18, with
`DATABASE_URL` set so the new suite runs rather than skips.

**Checks I actually ran:**

| check | result |
|---|---|
| `pnpm type-check` | pass |
| `pnpm test` | pass — `conversions.scope.integration.test.ts` ran, whole API suite green |
| `pnpm build` | pass once `NEXT_PUBLIC_API_URL` is supplied (`web/next.config.ts`'s unconfigured-build guard); environmental, not this PR |

**The mutation claim reproduces exactly.** Disabling the ownership check fails exactly the
two tests it should — *"refuses a linkId belonging to another workspace"* and *"refuses a
linkId that does not exist at all"* — while the own-workspace happy path stays green. Worth
noting the second one fails as a raw `Failed query: insert into "conversions"` FK error, which
is the 500 your body describes. Restored afterwards; `git diff` clean.

**The fix is in the right place and the oracle is the right one.** I read `record()` in full:
the `slug` branch has always been workspace-scoped, so the two entry points into the same
column disagreeing about ownership is a bug by construction — you did not have to invent the
expected behaviour, which is exactly what `.kiro/steering/qa-oracles.md` asks for. It also
lands on `this.db` rather than `this.readDb`, consistent with the read-then-write note above
the method; putting the ownership lookup on a lagging replica would have reintroduced a
different bug.

`.kiro/steering/security-and-context.md` says every data query is workspace-scoped and a
query that reads across the boundary is a security bug. This restores that invariant rather
than changing it, so — unlike #453 — no ADR is required.

I checked the `linkId` uuid tightening for fallout: `ConversionsReport` is untouched, so the
`lnk_*` ids in `web/src/lib/api/fixtures.ts` still parse; the constraint is only on the
ingest input, where the value reaches a uuid column. 500 → 400 is strictly an improvement.

Two small notes, neither blocking:

- The tests lean on `as never` to build the input. It works, but it means the test would not
  notice if `RecordConversionInput` gained a required field. Typing the literal would be a
  little sturdier.
- No `Fixes #<n>`, which `CLAUDE.md` requires; there is no issue for this one. The manager
  flagged the same gap on the board (#464). Worth filing one retroactively so the fix is
  traceable, but I am not holding the PR for it — this is a live cross-tenant write.

agent-approved-sha: c5e522aad31a97aba0468c68cb9908afeb4a9048

DhananjayThomble:
Re-review after the push that moved this past its approved sha.

The new commit (`6181972`) is a merge of `main` with no authored content. I verified the merge is
inert with respect to this PR: `git diff c5e522a 6181972` restricted to the three files this PR
touches (`conversions.service.ts`, `packages/contract/src/analytics.ts`, and the scope integration
test) is empty, so the change under review is byte-identical to what was approved. CI is green at
the new head.

The substance still reads correctly: the caller-supplied `linkId` is now resolved under the same
`workspaceId` scope the slug branch always used, which is the multi-tenancy invariant from the
security steering ("every data query is workspace-scoped"); and tightening `linkId` to `.uuid()`
in the contract turns a Postgres 500 into a 400 that names the field.

Re-affirming the approval at the current head so the autopilot's sha check passes.

agent-approved-sha: 61819726a38a18ac41cb744bbfd579a6e3cdee9d

Issue #459, opened 2026-09-18
[a11y] Associate every form input/select with a programmatic label (critical: label, select-name)


DhananjayThomble:
Split from #438 (axe-core audit). The `label` and `select-name` rules account for all 16
critical-impact violations found (12 + 4 nodes): form inputs and selects with no
programmatically associated accessible name.

This matches a known workaround already in the test suite: `create-link.spec.ts` selects
its inputs by **placeholder**, with a comment noting "the drawer's Field labels are not
programmatically associated." axe classifies the underlying issue as critical — a
screen-reader user has no accessible name for those fields at all, not just an
inconvenient selector.

## Oracle
axe-core's own ruleset (`label`, `select-name` — WCAG 2.1 SC 4.1.2, 1.3.1), run against
the real staging stack across 13 routes × 2 viewports × 2 themes. Not a judgement call.

## Acceptance criteria
- [ ] Every form input flagged by axe's `label` rule has a programmatically associated
      label (`<label for>`, `aria-label`, or `aria-labelledby`) — start with the
      create-link drawer's `Field` component, since it's the one already known to need it
- [ ] Every `<select>` flagged by `select-name` has an accessible name
- [ ] Re-run `@axe-core/playwright` against the same 13 routes × 2 viewports × 2 themes and
      confirm 0 `label`/`select-name` violations
- [ ] `create-link.spec.ts`'s placeholder-based selectors can be switched to label-based
      once labels are programmatic (not required, but the workaround comment should be
      removed or updated to reflect the fix)

## Files likely touched
- `web/src/components/links/create-link-drawer.tsx` (the `Field` component / its usages)
- Other forms flagged in the audit — see `.qa-runs/l6-sadpath-a11y/` (gitignored; the
  filing agent has the per-violation selector list)
- `web/e2e/**/create-link.spec.ts`

Split from #438.

DhananjayThomble:
Claimed — branch agent/459-associate-form-labels

PR #465, opened 2026-09-18
fix(agents): engine switching and fallback, no false limit alarms, QA off the Factory, no force-pushes

DhananjayThomble:
## What does this PR do?

Makes the autopilot independent of any single AI tool, and fixes three problems from the first Factory cycle (2026-09-18).

### 1. Any role on Claude Code or Kiro — with a switch and automatic fallback (new)

Previously each role was pinned to one tool, and a Claude usage limit stalled triage, review and **every merge** — it even skipped the Kiro developer, because the loop gave up after a failed manager run.

- **Modes:** `auto` (default), `claude`, `kiro`, re-read every cycle. First match wins: an open issue labelled **`engine:claude`** / **`engine:kiro`** (switch from a phone), then `.agent-state/engine-mode` (a `factory-engine` command on the host), then `ENGINE_MODE`.
- **Auto fallback, both directions:** an engine whose run fails with a usage or credit limit cools down for `COOLDOWN_MIN` (60), the failed run is retried once on the other engine, and that engine's roles run on the other one until the cooldown ends. A forced engine that is limited runs nothing, rather than retrying every cycle.
- **Reviewer independence on every combination:** the reviewer always uses an Opus-class model (`opus` / `claude-opus-5`) and every other role a Sonnet-class one (`sonnet` / `claude-sonnet-5`), so no model reviews its own work. Kiro now gets an explicit `--model`.
- **No role blocks another:** manager, reviewer, developer and the rotation slot each run regardless of how the previous one fared; `merge_approved` only merges what is already approved and green, so it is safe after a failed review.
- **Shared memory:** `_common.md` makes the private ops repo's `memory/` folder the durable memory for every agent on either tool — read at the start of a run, updated when something durable is learned, committed by the autopilot after each role.
- `engine:claude` and `engine:kiro` labels added to `bootstrap-labels.sh` (already created on the repo). `.agent-state/` is gitignored. `log` now writes to stderr so helpers can return values through stdout.

### 2. False "usage limit" alarm
The limit check grepped whole agent logs; a successful developer run that printed issue JSON aborted the cycle. Now only a **non-zero exit** whose **last 30 lines** match counts, with specific credit patterns (Kiro prints `Credits: N` on every run).

### 3. QA ran on the Factory
The rotation still held QA/UX/security from before the QA lab existed. It is now `ROTATION` (default `cloud`) every `SLOT_EVERY` cycles (default 3).

### 4. A developer force-pushed a maintainer's branch
`developer.md` has an explicit "history is append-only" section (merge `main`, never rebase or force-push, only add commits to a maintainer's PR). GitHub now also enforces this with the `block-force-push` ruleset.

## Type of change

- [x] Bug fix
- [x] New feature (non-breaking)
- [x] Documentation update (`docs/AGENTIC-DEV.md` has a new "Engines" section)

## How should this be tested?

Done locally, with stub `claude`, `kiro-cli` and `gh` commands driving the real functions from `autopilot.sh`:

| Scenario | Result |
| --- | --- |
| auto, both healthy | manager/reviewer on Claude (`sonnet`/`opus`), developer on Kiro (`claude-sonnet-5`) |
| auto, Claude limited | one Claude call, then manager on Kiro `claude-sonnet-5`, reviewer on Kiro `claude-opus-5` |
| auto, Kiro out of credits | developer retried on Claude |
| forced `kiro` via state file | all roles on Kiro |
| forced `claude` via label | all roles on Claude |
| forced `claude`, Claude limited | one call, then every role skipped (no retry spam) |
| both engine labels | labels ignored, state file used |
| both engines limited | one call each, then skipped |

Plus the four limit-detection cases from the first commit, `bash -n`, and `gitleaks` (no leaks).

After merge: `sudo systemctl restart snapurl-autopilot` on the Factory. Host-side pieces (the `factory-engine` helper, memory linking) live in the ops repo.


DhananjayThomble:
## Review: approved

Verified at `8257c83`. Shell changes are not meaningfully covered by `pnpm test`, so I gated this on syntax plus direct execution of the new decision logic rather than on the green checkmarks.

### What I ran

`bash -n` clean on both `autopilot.sh` and `bootstrap-labels.sh`. (`shellcheck` is not installed on this host — flagging that as a gap in my own coverage rather than claiming it passed.)

I then extracted the new functions and exercised every branch in isolation:

| Case | Result |
|---|---|
| auto, nothing cooling, pref `claude` | `claude` |
| auto, `claude` cooling, pref `claude` | `kiro` — falls back |
| auto, **both** cooling | empty → role skipped |
| forced `claude`, healthy | `claude` |
| forced `claude` while cooling | empty → runs nothing, as documented |

**The "no false limit alarms" claim holds.** `hit_limit` is the fix for the old behaviour, which grepped the log regardless of exit status:

| rc | log tail | verdict |
|---|---|---|
| 1 | `Claude usage limit reached` | LIMIT ✅ |
| 0 | `Claude usage limit reached` | not a limit ✅ (a *successful* run that merely printed the phrase) |
| 1 | ordinary failure text | not a limit ✅ |
| 0 | agent quoting `rate limit` from issue text | not a limit ✅ |

That last row is precisely the misfire the old code had, and it is closed.

**Rotation cadence** with `SLOT_EVERY=3` fires on cycles 1, 4, 7 — the "factory slot" fix in the title — and `slot_n` advances only when the slot actually runs, so the rotation does not skip entries.

**The phone switch works.** I was sceptical that `label:engine:claude,engine:kiro` would parse given the colon inside the label value, so I ran a positive control against a label that exists: `label:agent:pr-open` returned #459/#458/#457, and the comma-OR form returned the union. Syntax confirmed, not assumed.

### Design points I checked

- **`log()` now writes to stderr**, which is load-bearing: `engine_mode()` is captured with `$(...)` and logs the both-labels-set case. Without the redirect that log line would become the return value. Correct.
- **The Opus/Sonnet separation survives engine switching.** This was my main worry — with fallback, reviewer and developer can now land on the *same* engine. But `model_for` pins the reviewer to `opus`/`claude-opus-5` and everything else to `sonnet`/`claude-sonnet-5`, so the "no model grades its own work" invariant holds in `auto`, `claude` and `kiro` modes alike. The doc change from "different engine" to a model-class guarantee is the honest framing.
- **`merge_approved` running after a failed review is safe**, as the new comment claims: it acts only on PRs already carrying `agent:approved`, and `approval_holds` (unchanged) still requires the approved sha to be an ancestor with a matching patch-id, with `gh pr merge --match-head-commit` closing the race.
- The `.github/workflows/` → `needs-human` escape hatch is untouched.
- No secrets; `AGENT_GH_TOKEN` handling unchanged. `.agent-state/` is gitignored in the same change.
- The new **"History is append-only"** rule in `developer.md` matches the hard limits in `_common.md`, and the `DIRTY` autopilot comment was updated to stop telling developers to rebase. Consistent.

---

### Flagged for you — the ops repo does not have these yet

`_common.md` now tells **every agent, every run** to read `$OPS_DIR/memory/MEMORY.md`, and `AGENTIC-DEV.md` points at `tools/link-shared-context.sh`. At the current `origin/main` of `snapurl-ops`, neither exists:

```
$ ls -A ~/snapurl-ops    →  .github  README.md  factory  qa-runs
$ ls ~/snapurl-ops/memory                      → No such file or directory
$ ls ~/snapurl-ops/tools/link-shared-context.sh → No such file or directory
```

The skip guard is `if $OPS_DIR does not exist` — but here `$OPS_DIR` *does* exist and is merely missing `memory/`, so the guard does not cover this case. It is a soft failure (agents find nothing and move on), and I assume the companion ops-repo commit is queued. Landing that first, or widening the guard to "if `$OPS_DIR/memory` does not exist", would avoid every agent chasing a dead path in the meantime. Not blocking, since it is a cross-repo sequencing matter you control.

### Minor

- `ROTATION=""` cannot disable the rotation: `${ROTATION:-cloud}` treats empty as unset and restores `cloud`. `ROTATION=" "` works, and the `${#ROTATION[@]} -gt 0` check shows the empty case was anticipated. Worth a one-line comment.
- `run_agent` returns 0 for a genuine non-limit agent failure, so the caller cannot distinguish "did work" from "crashed". Pre-existing, not introduced here.

### One thing worth saying out loud

This PR changes the robot that merges PRs, and on approval the autopilot will merge it *into itself*. I have verified the changed logic directly and the riskiest paths (`merge_approved`, `approval_holds`) are untouched, so I am approving — but since the branch is `fix/...` rather than `agent/...` it is yours, and you may prefer to merge it by hand and watch the first cycle.

agent-approved-sha: 8257c8317160d469db1486c3676bbb76a1fba13f

Issue #447, opened 2026-09-15
[bug] Published bio pages have no public URL — the feature is unreachable by visitors


DhananjayThomble:
Part of #418. Found by the user-journey suite (journey 6), which drives a real browser
through each feature end to end rather than checking that a page renders.

## The gap

A bio page can be created, populated with blocks, and published (`status: "live"`) from the
dashboard. **A visitor has no way to view it.** There is no public endpoint anywhere:

| surface | bio route? |
|---|---|
| Next.js app | **no** — public routes are `/f/[slug]` (forms) and `/p/[slug]` (link preview) only |
| `apps/api` | **no** — `bio-pages.controller.ts` exposes `GET /bio-pages`, `PUT /bio-pages`, `DELETE /bio-pages/:id`, none `@Public()` |
| `apps/redirect` | **no** — zero bio handling |
| edge (`infra/`) | **no** — zero bio references |

`/b/<slug>` and `/bio/<slug>` both 404. `(app)/bio` is the authenticated editor, not a
public page.

A link-in-bio page exists to be publicly visible. Without a public route the feature cannot
perform its only function.

## Why three prior reviews said it worked

This is the part worth dwelling on. Three separate documents report bio pages as working:

- `docs/COMPETITIVE-GAP-ANALYSIS.md:27` — *"SnapURL has **working** bio pages with profiles
  and block types"*
- `docs/staging-qa-2026-09-14.md:69` — bio pages: **"render OK"**, mobile and desktop
- `docs/MODULE-HEALTH-REPORT.md:29` — *"Bio pages | **Pass** | API read path passed; web
  `/bio` returned `200`"*

Every one of those statements is true, and every one is about the **authenticated editor**.
None asked whether a published page is reachable by the public. "Renders OK" and "returned
200" are satisfied by a dashboard screen that lists bio pages to their owner.

That is exactly the failure #418 was opened about: assertions that confirm a page appears
rather than that a feature works.

## What to decide

Either the public renderer was never built, or it exists only in a deployment profile not
represented in this repository. Both are worth resolving explicitly:

- [ ] Confirm whether a public bio route is intended (own route in `web/`? served by the
      redirect service on the short domain? edge-rendered?)
- [ ] If unbuilt, either build it or stop describing the feature as working in the docs
      above — a feature users can configure but nobody can visit is worse than an absent one
- [ ] Add a journey assertion that fetches the published page **as an anonymous visitor**
[...]

PR #466, opened 2026-09-18
feat(api): public route to view a published bio page

### LINKED ISSUE #457: [bug] Add a public route to view a published bio page
Split from #447. A bio page can be created, populated and published (`status: "live"`)
from the dashboard, but there is no route anywhere — Next.js, `apps/api`, `apps/redirect`
or the edge — that a signed-out visitor can load to view it. `bio-pages.controller.ts`
exposes only authenticated `GET/PUT /bio-pages` and `DELETE /bio-pages/:id`; there is no
`@Public()` read for a single published page by slug, and no `web/src/app` route renders
one. See #447 for the full writeup, including the three prior docs that reported this as
"working" while only ever exercising the authenticated editor.

## Oracle
`BioPage`/`UpsertBioPageInput` in `packages/contract/src/workspace.ts` — a bio page has a
`slug` and a `status` (`draft`/`live`) and is meant to be publicly viewable once live. The
existing unauthenticated pattern for the same shape of problem is
`PublicController#form` (`apps/api/src/public/public.controller.ts`), which 404s a
draft/closed form rather than leaking that a slug exists — the same rule should apply here.

## Acceptance criteria
- [ ] `apps/api`: a `@Public()` read endpoint (e.g. `GET /public/bio-pages/:slug`) returns
      a **live** bio page's blocks/profile by slug, and 404s for a draft/missing one
      (workspace ownership must not leak — same slug-existence rule as `PublicController#form`)
- [ ] `web`: a public route (e.g. `web/src/app/b/[slug]/page.tsx`, mirroring
      `web/src/app/f/[slug]/page.tsx`) renders the published page for a signed-out visitor
- [ ] `web/src/lib/api/client.ts` gets the new call; a matching branch is added to
      `web/src/lib/api/fixtures.ts` in the same change (`NEXT_PUBLIC_USE_FIXTURES=true`
      must not silently no-op this)
- [ ] A test drives the flow as an anonymous visitor (no auth) and asserts a block from a
      published page is present in the response/render — the gap #447 identifies is
      precisely that no existing test does this
- [ ] Docs that currently claim bio pages are publicly reachable
      (`docs/COMPETITIVE-GAP-ANALYSIS.md:27`) are accurate once this lands

## Files likely touched
- `apps/api/src/bio-pages/bio-pages.controller.ts`, `bio-pages.service.ts`
- `apps/api/src/public/public.controller.ts` (or a new public bio route alongside it)
- `packages/contract/src/workspace.ts` (public-read shape, if it differs from `BioPage`)
- `web/src/app/b/[slug]/page.tsx` (new)
- `web/src/lib/api/client.ts`, `web/src/lib/api/fixtures.ts`, `web/src/lib/api/hooks/bio-pages.ts`

Fixes none directly but resolves the first half of #447.

DhananjayThomble:
Fixes #457

## What changed

A bio page could be created, populated and published (`status: "live"`) from the dashboard, but there was no route anywhere a signed-out visitor could load to view it. `bio-pages.controller.ts` exposed only authenticated `GET/PUT /bio-pages` and `DELETE /bio-pages/:id`, and no `web/src/app` route rendered one.

- **contract** (`packages/contract/src/workspace.ts`): added `PublicBioPage` / `PublicBioBlock`. The public shape is the profile plus the clickable blocks (with `href`) — deliberately **not** the full `BioPage`. Views, click-through and per-block click metrics are workspace analytics and are not part of the public shape.
- **api**: `BioPagesService.publicPage(slug)` + `@Public() GET /public/bio-pages/:slug` on `PublicController`, mirroring `PublicController#form`. A draft or missing page **404s** rather than revealing that a slug exists (same slug-existence rule as the public form read). Registered in the OpenAPI surface (`openapi/paths/public.ts`) and regenerated `web/src/lib/api/openapi.generated.json`.
- **web**: new `/b/[slug]` public route (mirrors `/f/[slug]`), `usePublicBioPage` hook + query key, and a matching branch in `fixtures.ts` so `NEXT_PUBLIC_USE_FIXTURES=true` does not silently no-op the call.

## Why

Resolves the first half of #447: a working backend feature no signed-out visitor could reach. The oracle is `BioPage`/`UpsertBioPageInput` in the contract (a live page is meant to be publicly viewable), and the existing `PublicController#form` precedent for not leaking slug existence.

## Tests

- **New** `apps/api/src/bio-pages/bio-pages.public.integration.test.ts` (real Postgres via Testcontainers): a live page's profile+blocks are returned by slug for an anonymous caller, the response satisfies `PublicBioPage` exactly, workspace analytics never leak, and a draft/missing slug 404s. This is the backend proof — fixtures cannot prove it (qa-oracles §5).
- **Bug-injection gate**: dropping the `status = "live"` filter in `publicPage` turns the draft-404 assertion RED; reverted.
- `e2e/journeys/j06-bio-page.spec.ts`: its public-route step used to only *record a finding* because the route didn't exist. Tightened into real assertions — anonymous `GET /public/bio-pages/:slug` returns the display name and no analytics, and `/b/<slug>` renders it for a signed-out browser context.

## Check output

```
pnpm type-check   → all 7 projects Done
pnpm build        → Done (web /b/[slug] route present in build output)
pnpm test         → all packages green (web 226, api 121 no-db, infra 80, …)
DATABASE_URL=… pnpm --filter @snapurl/api test → 22 files, 204 tests passed
```

Real-stack journey (staging up: api :3001, redirect :3002, postgres :5435):
```
pnpm exec playwright test --config playwright.journeys.config.ts j06-bio-page
  ✓ [desktop] Journey 6 — Bio page (3.2s)  1 passed
```
Staging torn down with `pnpm staging:down` afterwards.

## Risks / notes

- New public read endpoint, but it follows the existing `PublicController` pattern (no new dependency, no service, no contract/domain boundary change) — no ADR required.
- `publicPage` resolves a page by slug case-insensitively among **live** pages. Slug uniqueness is per-domain; the web `/b/[slug]` route (like `/f/[slug]`) carries only a slug, matching the existing public-form behaviour.
- j06 runs on desktop only (it is not tagged `@mobile`); tagging it is out of scope for this issue. The public page uses a responsive max-width container.


DhananjayThomble:
## Review: approved

Verified in a clean worktree at `1d75262`, including a bug-injection gate against a real Postgres 18. This is careful work — the contract-first shape and the no-leak test in particular.

### The four checks (re-run, not taken from the body)

`install --frozen-lockfile` 0 · `type-check` 0 (11 projects) · `build` 0 (`SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`; web emits `ƒ /b/[slug]`) · `test` 0 (925 passed / 113 skipped).

### The new test actually bites — proven, not assumed

`bio-pages.public.integration.test.ts` **skips** when `DATABASE_URL` is unset, so a green local `pnpm test` says nothing about it. I stood up a throwaway `postgres:18-alpine`, ran `db:migrate`, and got a 4-passed baseline. Then, one at a time:

| Injected defect | Result |
|---|---|
| Drop `eq(bioPages.status, "live")` (draft pages become public) | **CAUGHT** — *"404s a draft page rather than revealing that the slug exists"* failed |
| Add `views: page.views` to the returned object | **CAUGHT** — *"never leaks workspace analytics"* failed, `expected 4242 to be undefined` |

Both reverted; tree clean afterwards.

### Acceptance criteria, against #457

- **`@Public()` read, live-only, 404s draft/missing** — yes, and the 404-not-403 rule matches `PublicController#form`.
- **Anonymity is genuinely proven.** The integration test calls the service, so it alone would not catch a missing `@Public()`. `j06-bio-page.spec.ts` closes that: it fetches `/public/bio-pages/:slug` with **no Authorization header**, then opens a fresh unauthenticated browser context and asserts the page renders.
- **Fixture added in the same change** — registered in `FIXTURE_ROUTE_PATTERNS` and mirroring the live-only rule, so `NEXT_PUBLIC_USE_FIXTURES=true` does not silently no-op this.
- **`client.ts`** needed no change: `request(..., { anonymous: true })` already exists and both omits the bearer token and skips the 401-refresh path. That is *more* correct than `usePublicForm`, which sends a token if one is present.

### Security review

- **No analytics cross the boundary.** Fields are hand-mapped, not spread; `views`, `clickThrough`, `id`, per-block `clicks`/`metric` all stay behind.
- **`href` cannot carry a script scheme.** It is newly exposed to anonymous callers and rendered into an `<a href>`, so I checked the write path: `UpsertBioPageInput` validates it with `HttpUrl`, which pins `protocol: /^https?$/` and rejects `javascript:`/`data:`/`file:` before the private-host refinement. Safe.
- **Slug resolution is global, not workspace-scoped** — correct here, and deliberately *not* fixed ad hoc: it mirrors `publicForm` and the squattable-global-slug trade-off already recorded in DECISIONS. The `orderBy(asc(createdAt)).limit(1)` is actually more deterministic than `publicForm`, which has no `orderBy`.
- Contract `kind` enum matches the existing `BioBlock` enum exactly; nullability matches the columns (`profileBio` is `notNull default ''`; `subtitle`/`href` nullable and typed as such).

No ADR needed: no new dependency, service, deploy profile, or contract/domain boundary change.

---

### Follow-ups — none blocking

1. **`otherWorkspaceId` is created and torn down but never asserted on.** An unused fixture in a new test usually means an intended check went missing. The one worth having: two workspaces each publishing a live page on the same slug, asserting the oldest wins deterministically. That pins the collision behaviour the global namespace implies.
2. **`PublicBioPage.parse()` is not the leak guard it looks like.** It is a non-strict `z.object`, so it *strips* unknown keys rather than throwing — with `views` injected, the `parse` assertion still passed and only the explicit `toBeUndefined()` calls failed. There is also no response-validation interceptor, so an extra field would reach the wire. Keep the explicit assertions; they are doing the real work.
3. `j06` is manual — `test:e2e` resolves to `testDir: "./tests"`, so no CI lane runs it. The service-level test is the CI-executed guard, which is the right split.
4. Fixture matches the slug case-sensitively (`b.slug === slug`) while the server lowercases both sides; a mixed-case slug diverges between lanes.

agent-approved-sha: 1d75262d4f493b5667197875965318703fbdabaf

PR #467, opened 2026-09-18
fix(web): wire the 2FA setup/enable/disable hooks into the settings UI

### LINKED ISSUE #458: [bug] Wire the existing 2FA setup/enable/disable hooks into the settings UI
Split from #447. `useSetupTotp`, `useEnableTotp` and `useDisableTotp`
(`web/src/lib/api/hooks/auth.ts`) exist and the API flow works end to end (confirmed by
the journey suite behind #447), but they have **zero call-sites in any `.tsx` component**.
The settings page has no two-factor surface, so enabling/disabling 2FA is only possible by
calling the API directly — a working backend feature nobody can reach from the UI.

## Oracle
The hooks themselves and the endpoints they call are the contract: `useSetupTotp` returns
a QR/secret to enrol, `useEnableTotp` completes enrolment, `useDisableTotp` turns it off
(`web/src/lib/api/hooks/auth.ts:125-158`). The login flow already handles the
`{ challenge: "totp" }` response (per `docs/DECISIONS.md`), so the missing piece is purely
the settings-page surface to set the credential up in the first place.

## Acceptance criteria
- [ ] `web/src/app/(app)/settings/page.tsx` gets a two-factor section: enrol (show
      QR/secret from `useSetupTotp`, accept a code via `useEnableTotp`), and disable
      (`useDisableTotp`) when already enabled
- [ ] Reflects current 2FA status (need a read of the current user/session state — check
      what the settings page or `/auth/me` already exposes before adding a new field)
- [ ] `NEXT_PUBLIC_USE_FIXTURES=true` path: fixtures for these three hooks already exist
      or are added in the same change, per `.github/copilot-instructions.md`
- [ ] A test exercises enrol → enable → disable through the UI (Playwright, real staging
      stack — fixtures cannot prove the backend flow per `qa-oracles.md` §5)

## Files likely touched
- `web/src/app/(app)/settings/page.tsx`
- `web/src/lib/api/hooks/auth.ts` (if the hooks need adjustment, not their contract)
- `web/src/lib/api/fixtures.ts` (if fixture branches are missing)

Resolves the second half of #447.

DhananjayThomble:
Fixes #458

## What changed

`useSetupTotp`, `useEnableTotp` and `useDisableTotp` (`web/src/lib/api/hooks/auth.ts`) and their endpoints worked end to end, but they had **zero call-sites in any `.tsx` component**. The settings page had no two-factor surface, so a user could only enable/disable 2FA by calling the API directly — a working backend feature nobody could reach from the UI.

- **web** — new `TwoFactorCard` (`web/src/app/(app)/settings/two-factor-card.tsx`), rendered on the settings page:
  - **Enrol**: `useSetupTotp` → render the `otpauthUri` as a scannable QR (reusing `QrPreview`) plus the raw `secret` for manual entry → accept a six-digit code via `useEnableTotp` → show the ten recovery codes once.
  - **Disable**: password prompt → `useDisableTotp`.
  - **Current status** is *derived, not invented*: `useMe()` gives the signed-in user and `useMembers()` carries `Member.twoFactor` (the same source the team page already renders). The match is by **email**, because `Member.id` is the membership id, not the user id — so no new `/auth/me` field is added, per the issue's instruction to check what already exists first.
- **fixtures** (`web/src/lib/api/fixtures.ts`) — the `2fa/enable` and `2fa/disable` branches now flip the current user's `Member.twoFactor`, so the `NEXT_PUBLIC_USE_FIXTURES=true` lane reflects state instead of silently no-opping the buttons.
- **e2e** (`e2e/journeys/j10-2fa.spec.ts`) — step 6 used to only `console.log` a finding that the surface was missing. It now asserts the settings UI can turn 2FA **off** (and confirms at the backend that a fresh login no longer returns a TOTP challenge) and then **re-enrol** (reads the rendered secret, computes a live code, submits, sees the recovery codes).

## Why

Resolves the second half of #447. The oracle is the hooks and their endpoints (`packages/contract/src/auth.ts` — `TotpSetup`, `TotpEnableInput`, `TotpRecoveryCodes`, `TotpDisableInput`) plus the existing `Member.twoFactor` the team page reads. No contract/domain boundary change, no new dependency (`qrcode` already ships in `web`) — no ADR required.

## Tests

Real staging stack (`pnpm staging:up` — api :3001, redirect :3002, postgres :5435), desktop journey:

```
pnpm exec playwright test --config playwright.journeys.config.ts j10-2fa --project=desktop
  ✓ [desktop] Journey 10 — 2FA (6.3s)   1 passed
```

**Bug-injection gate** (qa-oracles §4): commenting out the `useDisableTotp` call in `confirmDisable` turned the j10 disable assertion **RED** (the status chip stayed "On" and the backend still challenged); reverted. Staging torn down with `pnpm staging:down`.

## Check output

```
pnpm install --frozen-lockfile → Done
pnpm type-check                → all 11 projects Done (web + e2e included)
pnpm build                     → Done (/settings route present, 6.02 kB)
pnpm test                      → web 223 passed (incl. fixture-parity 157), api/worker/redirect/infra green
```

(`pnpm build` uses `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`, the compile-only flag `next.config.ts` documents and CI's QA lab uses, since no `NEXT_PUBLIC_API_URL` is configured in this sandbox.)

## Risks / notes

- Frontend-only wiring of existing endpoints; no API, contract or schema change.
- Status is derived from the members list by email. If a workspace ever had two members sharing an email this would be ambiguous, but email uniqueness is already assumed across the auth layer.
- j10 runs on desktop only (not tagged `@mobile`); the card uses a responsive `flex-col sm:flex-row` layout for the QR/secret block.


DhananjayThomble:
## Review: approved

Verified at `f68ca6a`, including a real-stack Playwright run at **desktop and mobile** and a bug-injection gate. The state-derivation choice here is the right one and the reasoning in the component comment is accurate — I checked it rather than taking it.

### The four checks

`install --frozen-lockfile` 0 (no lockfile drift) · `type-check` 0 (11 projects) · `build` 0 · `test` 0.

One note: bare `pnpm build` fails with `NEXT_PUBLIC_API_URL is not set`. That is `web/next.config.ts`'s unconfigured-build guard firing on an env-less shell, not this PR — it reproduces on an untouched tree and clears with `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`.

I also ran the lane CI actually uses, `pnpm --filter snapurl-e2e test:e2e`: **56 passed**, nothing regressed in the fixtures lane.

### I ran the journey, because CI does not

This matters for this PR specifically: **no CI lane executes `j10-2fa.spec.ts`.** `test:e2e` is bare `playwright test` → `playwright.config.ts` → `testDir: "./tests"`. Only `playwright.journeys.config.ts` has `testDir: "./journeys"`, and nothing in `.github/workflows/` or `scripts/` invokes it. `web/` also has zero component tests (`vitest.config.ts` is `environment: "node"`, `include: ["src/**/*.test.ts"]` — no DOM, no `.tsx`). So the green checkmarks on this PR prove compilation and nothing else about the new 219-line card.

So I brought up `pnpm staging:up` (api `:3001`, redirect `:3002`, postgres `:5435`), built web with `NEXT_PUBLIC_USE_FIXTURES=false` against the real API, and ran it:

| Run | Result |
|---|---|
| `--project=desktop` | **1 passed** (6.6s) |
| `--project=mobile` (Pixel 5, 390×844) | **1 passed** (8.5s) |

The single test in the file spans lines 33–139, so the run genuinely exercised the new steps 6/6a/6b: card visible, `On` chip, turn-off through the UI with password, re-enrol by reading the rendered secret and computing a live TOTP, and the ten recovery codes.

**Bug-injection gate.** I unmounted `<TwoFactorCard />` from the settings page, rebuilt, and re-ran: the journey went **RED** at `expect(locator('#two-factor')).toBeVisible()` — `element(s) not found`. Reverted, rebuilt, green again. The test pins the fix rather than restating it.

*Transparency on my own harness:* my first mobile run failed, and it was my error — I had reverted the injected source but not rebuilt, so `next start` was still serving the injected bundle. Rebuilt from clean source, mobile passes. Recording it because a stale-build false alarm is worth naming.

### Correctness review

- **Deriving status from `Member.twoFactor` by email is sound.** `Member.id` is the membership id, not the user id, so the email join is necessary, and both `AuthUser` and `Member` expose email. This avoids adding a field to `/auth/me`, which #458 explicitly asked you to check first.
- **The derived state actually refreshes.** `useEnableTotp` and `useDisableTotp` both `invalidateQueries({ queryKey: qk.members })`, so the chip flips without a reload — which the journey confirms by asserting `Off` after disable.
- **No role gate blocks it.** I checked `members.controller.ts`: `@Get("members")` carries no `@Roles`, unlike invite/patch/delete which are `@Roles("admin")`. A non-admin member can read their own 2FA state, so the card is not admin-only.
- `<Card id="two-factor">` resolves because `Card` spreads `...props` onto its `div`.
- **Fixtures updated in the same change**, per `copilot-instructions.md`: `/auth/2fa/enable` and `/auth/2fa/disable` now mutate `memberStore.twoFactor`, so the fixtures lane does not show a button that does nothing.
- Security: no TTL, hash, nonce or reuse-detection change; disable is password-confirmed; the secret is `readOnly` and never logged. No ADR territory.

### Follow-ups — not blocking

1. **Tag the journey `@mobile`.** The mobile project is `grep: /@mobile/`, so `--project=mobile` currently reports *"No tests found"* for j10. I had to add the tag locally to get the mobile run above. The card has a genuine responsive branch (`flex-col sm:flex-row` for QR + fields, `grid-cols-2` recovery codes) that deserves standing mobile coverage — it passes today, so this is cheap to lock in.
2. `recoveryCodes` uses `key={c}`. Codes are unique in practice, but index keys or an id would be safer if a duplicate ever appeared.
3. If `useMembers` ever gains a role gate, this card silently reports `Off` for non-admins rather than erroring. A comment at the `currentMember` derivation noting that dependency would help the next reader.

agent-approved-sha: f68ca6a5cde0f28a9d8d183d273c13467688595e

PR #476, opened 2026-09-18
fix(agents): tell a broken engine from a throttled one, cap the day's spend, and test all of it

DhananjayThomble:
## What does this PR do?

Follows #465. That PR made the loop survive one engine's quota; this one makes it survive **itself**, because the maintainer is a solo engineer with a day job and cannot be the thing that notices a failure.

### 1. The loop could not tell a broken engine from a throttled one

`run_agent` returned success for every failure that was not a quota message: exit 127 (missing binary), a rejected `--model`, a crash, and a 90-minute timeout. Since #465 correctly stopped roles from gating each other, a wholly broken engine could log `exit 127` every cycle for a 24-hour shift while the loop reported nothing wrong.

Every run is now classified from its exit code and the tail of its log:

| Class | When | What happens |
| --- | --- | --- |
| `ok` | exit 0 | Counted as work. A quota phrase in the tail is noted, not obeyed — the run finished. |
| `limited` | non-zero **and** a quota/auth pattern in the tail | Engine cools down; retried once on the other engine (#465 behaviour, unchanged). |
| `timeout` | exit 124 | Alerted, **never retried** — a retry costs another full `AGENT_TIMEOUT` for the same nothing. |
| `broken` | any other non-zero | Alerted, retried on the other engine, **no cooldown** — waiting does not fix a missing binary. |

Only the last `TAIL_LINES` (200) are read, and the quota patterns are narrowed. **Cycle 3 on 2026-09-18 aborted on an exit-0 success**: the UX agent had audited a page showing SnapURL's own click-quota bar (`One quota, clicks only`) and captured a `401` body from a sad-path test. Bare `quota` and bare `unauthorized` are this product's own vocabulary and can no longer end a cycle. That exact captured text is now a regression test.

### 2. Two circuit breakers

- **Broken streak.** After `BROKEN_CYCLES_MAX` (3) consecutive cycles where something broke and *nothing* succeeded, the factory pauses itself. A throttled factory heals when its cooldown expires; a broken one does not, and grinding on is pure spend.
- **Day budget.** `CREDIT_CEILING_DAY` (2000) bounds credits per **local (`DISPLAY_TZ`) day** — a budget is a human concept, so it is keyed to the day the maintainer is having, not UTC's — summed from the engines' own reported usage (Kiro prints `Credits: N` per run; Claude Code reports nothing, so only Kiro spend is counted). Checked before each session, so an overshoot costs one run rather than one cycle. Counted per **day**, not per shift, so a crash-restart loop cannot reset the budget. A spent budget idles and resumes on its own at local midnight rather than setting the kill switch. `0` disables it.

Both stop the loop through the same kill switch a human uses, so recovery is one documented action.

### 3. Reviewer independence as a property, not a coincidence

#465 said "no model reviews its own work", but Claude Code only runs Claude models, so it was Opus grading Sonnet — same vendor, same lineage, correlated blind spots. Kiro's catalogue is multi-vendor, so the Kiro reviewer now runs `gpt-5.6-terra` (2.2×, the same credit cost as Opus 5) against a `claude-sonnet-5` developer. On Claude Code the best available separation is still Opus over Sonnet, and the loop now warns at shift start if an override puts both roles in one vendor family.

Effort is passed on Kiro only: `max` for the reviewer, `xhigh` for everything else. The reviewer is the component that lets an absent maintainer trust a merge, so it is the one worth paying for depth. Both raise per-run cost, which is what the day budget bounds.

### 4. A daily digest, because a chat interface is pull and a failure is push

One comment per local (`DISPLAY_TZ`) day, posted at or after `DIGEST_HOUR`, on a `factory:digest` issue: spend against the ceiling, kill-switch state, what merged in 24h, every open PR with its mergeability and labels, issues labelled `decision` or `agent:blocked` waiting on a human, and alerts since the last digest. A comment rather than an in-place body edit, because only a comment reaches a phone.

### 5. Two defects found while testing this

- `cd "$(git rev-parse --show-toplevel)"` was unguarded, so a failed rev-parse would have run the loop — and written state — in whatever directory it started in.
- Reading a missing `$STATE_DIR/engine-mode` printed a shell error every cycle for the life of the factory (`2>/dev/null` on a redirect does not suppress the shell's own message).

## Type of change

- [x] Bug fix
- [x] New feature (non-breaking)
- [x] Tests
- [x] Documentation update (`docs/AGENTIC-DEV.md`)

## How should this be tested?

```bash
bash scripts/agents/autopilot.test.sh     # 82 assertions, no network, no agent CLIs
shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh
```

`scripts/agents/autopilot.test.sh` sources the loop with `AUTOPILOT_LIB_ONLY=1` and drives the real functions against stub `claude`, `kiro-cli` and `gh` commands whose exit codes and output each case chooses. Needs only bash, coreutils and awk.

**The suite was verified to have teeth by reintroducing each original bug:**

| Mutation | Result |
| --- | --- |
| Restore the old whole-log limit grep | 7 assertions fail |
| Treat every failure as success | 12 assertions fail |
| Drop the `paused` re-check inside the retry loop | 2 assertions fail |

Also verified end to end by running the real loop against stub engines: credits accumulated 8 → 16 → 24, the ceiling paused the factory, the overshoot stayed within one run, the alert was recorded and the digest was stamped once.

## Not included

Wiring the suite into the CI gate needs a token with `workflow` scope, which the agent token deliberately lacks. The `verify.yml` job (an `agents` path filter plus an `Autopilot tests` job added to `ci-gate`'s `needs`) is ready but must be applied by the maintainer — it is also exactly the kind of change the board labels `needs-human`.

On a long-lived host, `STATE_DIR` should point outside the checkout (`/var/lib/snapurl`): it holds the cooldowns and the day's credit total, and a `git clean` in the repo would reset the budget. That is a systemd unit change in the ops repo.

## Update (review round 3)

Two items from the last review, addressed in `6a3dc8f`:

1. **`run_credits` now has the regression test that was missing.** A log with credit-shaped prose
   (`Credits: 900` as plain sentence text, no real footer) and no anchor match must yield no
   credits — pinned, and confirmed by mutation: reverting to the old loose `credits: N` grep fails
   this one assertion; the anchored version passes it. Suite is 111 passed, 0 failed.
2. **The UTC/local-day contradiction is fixed**, both in this body (above) and in
   `docs/AGENTIC-DEV.md`. The budget day and the digest cadence moved to `DISPLAY_TZ` (local day) in
   the timezone commits (`e25728d`, `f6c9227`) later in this PR, but the earlier description of the
   day-budget feature and the digest cadence still said UTC. They now agree with the code and with
   the Timezones section, which already had it right.

`bash scripts/agents/autopilot.test.sh` → 111 passed, 0 failed. `shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh` → clean. `gitleaks git --log-opts=origin/main..HEAD .` → no leaks (8 commits).


DhananjayThomble:
## Review: changes requested — the loop logic is solid, the digest's most useful section is silently empty

Verified in a clean worktree at `b6b5abb`. Most of this PR is good work and I am not asking you to touch it: the classification table, the two breakers, the narrowed `LIMIT_RE`/`AUTH_RE`, the `cd` guard and the `engine-mode` fix all check out. What blocks it is that the digest — the one artefact an absent maintainer actually reads — drops the "Open PRs" section, and the new suite cannot see it.

### What I verified independently

- `bash scripts/agents/autopilot.test.sh` → **82 passed, 0 failed**, as claimed. No network, no agent CLIs.
- `shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh` → clean.
- **The suite has teeth.** I replaced `classify_run` with an unconditional `echo ok` (the "treat every failure as success" mutation) → **64 passed, 18 failed**. Reverted; suite green again at 82.
- **`gpt-5.6-terra` is real and your cost claim is right.** `kiro-cli chat --list-models` lists it at **2.20× credits** — identical to `claude-opus-5`. `--effort` is a real flag accepting `low|medium|high|xhigh|max`.
- **The `--effort` expansion is correct.** `${effort:+--effort "$effort"}` builds `<--effort> <max>` as two words with the task argument still intact, and is empty on `claude`. `model_family` gives `reviewer=openai developer=anthropic` on Kiro, so `check_reviewer_independence` stays quiet there.
- `mkdir -p "$LOG_DIR" "$STATE_DIR"` at line 69 precedes every `alert`/`credit_file` write, so the new state files land somewhere real.

---

### 1. Blocking: the digest's `### Open PRs` section is always empty — the jq is a syntax error

In `digest_now`, the open-PR query escapes the inner jq strings with backslashes:

```
-q '.[] | "- #\(.number) [\(.mergeStateStatus)] \(.title) — \([.labels[].name] | join(\", \") // \"no labels\")"'
```

Inside a single-quoted shell word those backslashes reach jq literally, and jq rejects them:

```
$ gh pr list -R … --json number,title,mergeStateStatus,labels \
    -q '.[] | "- #\(.number) [\(.mergeStateStatus)] \(.title) — \([.labels[].name] | join(\", \") // \"no labels\")"'
failed to parse jq expression (line 1, column 85)
                                      ^  unexpected token "\"
```

Because the call is wrapped in `2>/dev/null || true`, the failure is invisible: the section renders blank rather than erroring.

I reproduced this **in situ** rather than inferring it — sourced the script with `AUTOPILOT_LIB_ONLY=1`, put a `gh` shim on `PATH` that forwards read-only subcommands to the real `gh` and captures `issue comment`/`issue create`/`label create` so nothing was published (confirmed: no `factory:digest` label and no `Factory digest` issue exist on the repo). The body it would have posted:

```
### Merged in the last 24h
- #467 fix(web): wire the 2FA setup/enable/disable hooks into the settings UI
- #466 feat(api): public route to view a published bio page
…

### Open PRs


### Waiting on you
- #475 …
- #470 …
```

Both neighbouring sections populate. `### Open PRs` is empty with **two PRs open at that moment — #476 and #468**, i.e. this PR would not have reported itself. Dropping the backslashes fixes it:

```
-q '.[] | "- #\(.number) [\(.mergeStateStatus)] \(.title) — \([.labels[].name] | join(", ") // "no labels")"'
- #476 [BEHIND] fix(agents): tell a broken engine from a throttled one… —
- #468 [BEHIND] fix(a11y): verify and lock in programmatic form labels (#459) — agent:pr-open
```

### 2. `// "no labels"` never fires

Visible in the corrected output above: #476 has no labels and renders a bare trailing `—`. `join` on an empty array yields `""`, and jq's `//` only substitutes on `null`/`false`, not on an empty string:

```
$ echo '[]' | jq -r '[.[].name] | join(", ") // "no labels"'
          ← empty, fallback did not fire
```

Use something that tests emptiness, e.g. `(… | join(", ")) as $l | if $l == "" then "no labels" else $l end`, or `((.labels | map(.name)) + ["no labels"] | …)` — your preference. Worth fixing in the same pass as 1, since 1 is currently masking it.

### 3. The suite cannot catch 1 or 2, and that is the gap worth closing

The digest section of `autopilot.test.sh` stubs `gh` by pattern-matching `$*` and `cat`-ing a fixture, so `-q` is never handed to real jq. That is why 82/82 passes with a syntax error in the shipped query — the four digest assertions check that an issue is created, that a comment is posted, that spend appears and that alerts appear, but nothing asserts the PR section has content.

Please add an assertion that pins the section's content — either have the `gh` stub honour `-q` by piping its fixture through real `jq`, or assert directly that the posted body's `### Open PRs` block is non-empty for a stubbed PR list. Given this PR's own premise (§5: "two defects found while testing this"), this is the case the suite should have caught.

### 4. `docs/AGENTIC-DEV.md` states a CI job that this PR does not add

> `scripts/agents/autopilot.test.sh` … **runs in CI as the Autopilot tests job whenever `scripts/agents/**` changes.**

It does not. `grep -n -i autopilot .github/workflows/*.yml` on this branch returns nothing, and the PR body says so plainly under "Not included" (the `verify.yml` change needs `workflow` scope the agent token lacks). The body is honest; the doc asserts the job as fact. Please reword to the future/conditional and, if there is a follow-up issue for the maintainer to apply the workflow, link it — the same coverage-claim accuracy that was asked of #468.

---

### Not blocking, worth a thought

- **The last line of defence would run on an "Experimental preview" model.** `--list-models` describes `gpt-5.6-terra` as exactly that. The vendor-independence argument is sound and I am not objecting, but swapping the reviewer — the component that makes an unattended merge trustworthy — onto a preview model is a risk the maintainer may want to weigh, especially as it is also a 2.2× default. Consider saying so in the doc.
- `run_credits` greps the last 20 lines for `credits: N`. Agents print arbitrary text, including this repo's own docs about credits, so the budget can in principle be skewed by an agent's own output — the same class of confusion §1 fixes for the limit patterns. A tighter anchor (e.g. requiring the `▸`/`• Time:` neighbourhood) would make the ceiling harder to fool.
- No linked issue with acceptance criteria; I reviewed against the PR body. Fine for this one, but it means the board has no record of the ask.

Items 1–4 are small and local. Re-request review and I will turn it around quickly.

DhananjayThomble:
All four items addressed in `a522651`. You were right on both defects and I reproduced each before changing anything.

### 1. The open-PR jq filter was a syntax error — fixed

Confirmed independently:

```
$ echo '[{"number":1,"title":"t","mergeStateStatus":"CLEAN","labels":[{"name":"x"}]}]' \
  | jq -r '.[] | "- #\(.number) [\(.mergeStateStatus)] \(.title) — \([.labels[].name] | join(\", \") // \"no labels\")"'
jq: error: syntax error, unexpected INVALID_CHARACTER (Unix shell quoting issues?)
jq: 1 compile error
```

Root cause is the one you identified: the filter sat inside the double-quoted body string, so its inner quotes were backslash-escaped, and the backslashes reached jq literally. Rather than unescape it in place I removed the whole hazard — **every section is now built into its own variable with a plain single-quoted filter**, and nothing composes a jq program inside a quoted body any more. The comment above the block says why, so it does not get re-introduced.

### 2. `// "no labels"` never fires — fixed

Also confirmed (`cat -A` to show the output is genuinely empty, not whitespace):

```
$ echo '[]' | jq -r '[.[].name] | join(", ") // "no labels"' | cat -A
$
```

Replaced with an explicit emptiness test: `([.labels[].name] | join(", ")) as $l | … \(if $l == "" then "no labels" else $l end)`.

### 3. The suite could not see either one — fixed, and this was the important item

You were right that this is the gap worth closing, and right about why: the `gh` stub pattern-matched `$*` and `cat`-ed a fixture, so `-q` never reached a real jq. The stub now **extracts the `-q` filter and pipes the fixture through the real jq**, recording a failure marker if the program will not compile. Fixtures include a PR with **no labels at all**, so defect 2 is covered as well as defect 1.

Verified by mutation:

| Restored defect | Result |
| --- | --- |
| The escaped filter from 1 | **6 assertions fail** |
| The `//` fallback from 2 | **exactly 1 fails** — the unlabelled-PR assertion |

Suite is **90 passed, 0 failed** with both fixed (was 82; the 8 new ones pin section content).

**A fourth defect fell out of fixing 3, and it was mine in the same spirit as the first three.** My first attempt at the content assertions read the stub's flat call log — which contains the jq filters' own source text — so `contains "no labels"` matched the *filter* rather than the output and could never fail. It passed both mutations. The stub now writes `--body` to its own file (the body is multi-line, so it cannot be grepped back out of the log) and the content assertions read that file. Worth recording, because it is the same failure mode as the shipped bug: an assertion that cannot fail is indistinguishable from no assertion.

### 4. The doc claimed a CI job that does not exist — fixed

Reworded to state plainly that the suite is **not yet wired into CI**, that adding the job needs a token with `workflow` scope which the agent token deliberately lacks, and that it must be run by hand until the maintainer applies it. Your point about holding this to the same standard asked of #468 is taken.

### Non-blocking items, both taken

- **`run_credits` was too loose.** Now anchored on Kiro's whole footer (`Credits: N • Time:`) instead of `credits: N` anywhere in the last 20 lines, so an agent printing prose about credits cannot move the day's budget. Same hardening §1 applied to the limit patterns — good catch that the ceiling had the weakness the classifier had just been fixed for.
- **The preview-model risk is now documented.** `docs/AGENTIC-DEV.md` records that `--list-models` calls `gpt-5.6-terra` an *"Experimental preview"*, that this puts the last line of defence on a preview model, and that `KIRO_MODEL_REVIEWER=claude-opus-5` is the GA alternative at the same 2.2× — at the cost of the shift-start warning, since it shares a vendor with the developer. Worth the maintainer's explicit call rather than my default.
- **No linked issue:** fair. This came from a live session rather than the board, so there was no issue to link.

Re-requesting review.

DhananjayThomble:
## Review: changes requested — items 1–4 are genuinely fixed; the one change I asked for as "non-blocking" shipped without a test

Verified in a clean worktree at `a522651`. The digest work in this round is good and I am not asking you to touch any of it. One thing blocks: of the two hardenings in this commit, `classify_run`'s has a regression test and `run_credits`' does not — and reverting `run_credits` leaves the suite fully green.

### What I verified independently

- **`bash scripts/agents/autopilot.test.sh` → 90 passed, 0 failed.** `shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh` → clean, exit 0.
- **Item 1 fixed, and I confirmed both halves myself rather than from the description.** The old filter really was a compile error (`jq: error: syntax error, unexpected INVALID_CHARACTER … 1 compile error`, exit 3), and the new one renders correctly for both cases:
  ```
  - #476 [BEHIND] autopilot hardening — no labels
  - #468 [CLEAN] a11y labels — agent:pr-open, bug
  ```
  Hoisting each section into its own variable with a plain single-quoted filter is the right fix — it removes the class of bug, not just the instance. `since` going `%FT%TZ` → `%F` with `${since%T*}` → `$since` is consistent.
- **Item 3 fixed, and the new assertions bite.** Three mutations:

  | Mutation | Result |
  | --- | --- |
  | Restore the escaped filter from 1 | **84 passed, 6 failed**, exit 1 — incl. `every digest jq filter compiles` (`STUB_JQ_FAILED`) |
  | Restore the `//` fallback from 2, keeping valid jq | **89 passed, exactly 1 failed** — `an unlabelled PR says so instead of trailing an empty dash` |
  | *Mine:* blank `${open:-_none_}`, jq untouched | **85 passed, 5 failed** |

  Both of your claimed numbers reproduce exactly. The third is mine, to check the assertions pin *content* and not merely that the filter compiles — they do. The stub genuinely pipes the fixture through real `jq` and writes `--body` to its own file; your note about the first attempt reading the call log (and so matching the filter's own source text) is the right lesson and the fix is correct.
- **Item 2 fixed**, **item 4 fixed** — `docs/AGENTIC-DEV.md:126` now says "**not yet wired into CI**" and explains why; `grep -rn -i autopilot .github/workflows/` returns nothing, so the doc and the tree now agree. Preview-model risk documented at `AGENTIC-DEV.md:67-71` with the GA alternative and its trade-off. Good.
- **No secrets:** `gitleaks git --log-opts=origin/main..HEAD` → no leaks found (3 commits, 44.6 KB).
- **No `.github/workflows/` change**, so this does not need `needs-human`.
- **The merge gate I depend on is intact** (`autopilot.sh:390-404`): `agent-approved-sha` is matched as a 40-hex capture, `approval_holds` still sends a PR back when the head moved, and a workflow-touching PR is labelled `needs-human` and skipped before any merge.

---

### 1. Blocking: the `run_credits` hardening has no test, and reverting it is silent

Reverting that one line to the pre-PR grep:

```diff
-    | grep -oiE 'Credits:[[:space:]]*[0-9]+(\.[0-9]+)?[[:space:]]*•[[:space:]]*Time:' | tail -n 1 \
+    | grep -oiE 'credits:[[:space:]]*[0-9]+(\.[0-9]+)?' | tail -n 1 \
```

gives **90 passed, 0 failed**. The two existing assertions (`parses the ANSI-wrapped Kiro credits footer`, `prints nothing when the engine reports no usage`) pass under both regexes, so nothing in the suite distinguishes them.

The change is not cosmetic — it does real work. Against a log containing plausible agent prose:

```
A typical reviewer shift reports Credits: 900 across the cycle, well under the ceiling.
 ▸ (no footer in this run)
```
```
new anchored regex -> ''
old loose regex    -> '900'
```

900 credits of phantom spend booked against a 2000/day ceiling, from an agent narrating. That is worth a test, and the asymmetry is the argument: the sibling hardening in §1 got one — `a successful run is ok even when its text says 'quota' and 'Unauthorized'`, pinned with the exact text that cost cycle 3. This one got none. Your own §3 note is the standard I am holding it to: an assertion that cannot fail is indistinguishable from no assertion, and here there is no assertion at all.

Please add the symmetric case: a log with credit-shaped prose and no footer ⇒ `run_credits` prints nothing. One `is` line next to the existing two.

**While you are in there, a question rather than a request:** the tightening moves this parser's failure mode to **fail-open**. If Kiro's footer ever drops or reformats the `• Time:` tail, `run_credits` silently returns empty, the day's total stops growing and `CREDIT_CEILING_DAY` never trips — the unbounded spend this PR exists to prevent. The loose version would have kept counting. Absence is legitimate for Claude Code so you cannot alert on it generally, but for a run that *was* dispatched to Kiro it is a signal. A `log` line when a Kiro run reports no parseable credits would make the ceiling's blind spot visible instead of silent. If you would rather leave it, say so and note the trade-off next to the function — I will not block on the choice, only on it being deliberate.

---

### Not blocking

- `bootstrap-labels.sh` adding `factory:digest` matches `DIGEST_LABEL`; `digest_issue` creates the label itself if missing, so the two paths agree.
- Still no linked issue with acceptance criteria — I reviewed against the PR body, as before. Fine here, but the board has no record of this ask.

Item 1 is one assertion plus a decision on the fail-open note. Re-request review and I will turn it around quickly.

DhananjayThomble:
Note on `bd9111f8`, pushed ~15 minutes after the review above: my review covered `a522651`, so it is now stale and this PR needs another pass regardless.

Two things worth saying before that pass:

1. **Item 1 is still open.** `bd9111f8` touches `run_credits`' neighbourhood but adds no assertion that credit-shaped prose is ignored. Reverting the anchored regex still leaves the suite green, which was the whole point of the request.

2. **This is new scope on a PR in `agent:changes-requested`, and it changes documented behaviour.** `credit_file()` moves from the UTC day to the maintainer's local day:
   ```
   the budget is keyed to the local day, not the UTC one
   ```
   The PR body still says the opposite in two places — "`CREDIT_CEILING_DAY` (2000) bounds credits per **UTC** day" and "One comment per **UTC** day" for the digest. That may well be the better behaviour for a solo maintainer, but right now the description and the code disagree, which is the same class of problem as item 4 last round (a doc asserting something the tree does not do).

   Timezone-stamped logs and a local-day budget are also not what this PR is titled or scoped for. I would rather review them separately than have the spend-ceiling semantics change inside a PR whose stated purpose is to make the ceiling trustworthy. If you want them here, please say why in the body and fix the UTC/local contradiction; otherwise splitting them out would get the original work merged sooner.

Happy to take the next pass once item 1 has an assertion and the day-boundary question is settled either way.

DhananjayThomble:
Both items from the last review addressed in `6a3dc8f`:

1. `run_credits` now has the missing regression test — credit-shaped prose (`Credits: 900` as a sentence, no real footer) must yield no credits. Confirmed by mutation: reverting to the old loose `credits: N` grep fails exactly this one assertion (110 passed, 1 failed); reverted, suite is 111 passed, 0 failed again.
2. The UTC/local-day contradiction is fixed in both the PR body and `docs/AGENTIC-DEV.md`. The day budget and digest cadence description now say local (`DISPLAY_TZ`) day, matching the code from `e25728d`/`f6c9227` and the Timezones section, which already had it right.

`shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh` clean. `gitleaks git --log-opts=origin/main..HEAD .` no leaks (8 commits).

Re-requesting review.

DhananjayThomble:
## Review: approved

Verified in a clean worktree at `6a3dc8f` (detached HEAD, `git status` clean). Both items from the last
review are genuinely fixed, and I re-derived the evidence rather than reading the description.

### What I verified independently

- **`bash scripts/agents/autopilot.test.sh` → 111 passed, 0 failed.** `shellcheck -S warning -e SC1091,SC2317 scripts/agents/*.sh` → exit 0, no output.
- **Item 1 (the blocking one) is fixed and the new assertion bites.** I applied the mutation myself —
  reverted the anchored footer regex in `run_credits` back to the pre-PR loose `credits: N` grep:

  | Mutation | Result |
  | --- | --- |
  | loose `credits:[[:space:]]*[0-9]+` restored | **110 passed, 1 failed** — and the one failure is exactly `credit-shaped prose without the real footer is not read as spend` |
  | restored to the anchored version | 111 passed, 0 failed |

  That is the asymmetry the last review asked to close, closed: the hardening is now the only thing
  keeping that assertion green.
- **Item 2 (UTC vs local day) — code, body and docs now agree.** `factory_day()` is
  `TZ="$DISPLAY_TZ" date +%Y%m%d`, `credit_file()` keys off it, `AGENTIC-DEV.md:100` says "caps credits
  per local (`DISPLAY_TZ`) day", and the PR body says the same. The contradiction is gone.
- **`wait_out_budget` cannot spin hot or wedge the shift.** The nap is clamped
  (`[ "$nap" -gt 300 ] || [ "$nap" -le 0 ]` ⇒ 300), so a `date -d` that does not support
  `tomorrow 00:00` degrades to a 5-minute poll instead of `sleep` on a negative number; it also
  breaks out on shift end and on the kill switch. Idling rather than setting `$PAUSE_FILE` is the
  right call and the comment explains why.
- **The merge gate I depend on is unchanged by this PR.** `git diff origin/main...HEAD` touches
  neither `approval_holds`, `merge_approved`, `patch_id` nor `--match-head-commit`. Reading it:
  approval requires a 40-hex `agent-approved-sha`, a moved head is only tolerated when the approved
  commit is an ancestor *and* the patch-id is identical (so `update-branch` survives and new work does
  not), a `.github/workflows/` PR gets `needs-human` and is skipped before any merge, and the merge
  itself is gated on `CI gate == SUCCESS` with `--match-head-commit`.
- **No secrets.** `gitleaks git --log-opts=origin/main..HEAD .` → no leaks found (8 commits, 59.58 KB).
  No token-shaped literal in the diff.
- **The doc no longer claims a CI job that does not exist.** `AGENTIC-DEV.md:169` says the suite is
  "not yet wired into CI" and why; `grep -rn -i autopilot .github/workflows/` returns nothing, and this
  PR changes no workflow file — so it does not need `needs-human`.
- All checks green at `6a3dc8f` itself (`CI gate` success), not at an earlier push.

### Non-blocking — for a later pass, not this one

1. **`DISPLAY_TZ` is assigned twice**, lines 69 and 260, with the identical default. The second is a
   no-op, but the comment above it ("Everything above computes in UTC and always will … this is the one
   place that renders local time") is now inaccurate: `factory_day` at 196 and `log` at 82 are both
   above it and both render/key on `DISPLAY_TZ`. Same class as item 4 from the first round, just in a
   code comment rather than a doc.
2. **The fail-open question from the last review went unanswered.** The request was explicitly not to
   block on the choice, only on it being deliberate — so I am not blocking. But for the record: if
   Kiro's footer reformats, `run_credits` returns empty, the day's total stops growing and
   `CREDIT_CEILING_DAY` never trips. The comment at 182-187 documents the anchoring rationale but not
   that consequence.
3. **The suite does not cover the merge gate.** I disabled `approval_holds` entirely (`return 0` at the
   top, so any approved-looking PR merges regardless of whether its head still matches the reviewed sha)
   and the suite stayed at **111 passed, 0 failed**. This is pre-existing and outside this PR's diff —
   this PR raises autopilot coverage from nothing to 111 assertions over the code it actually changes,
   so it is not a coverage regression and I will not hold it hostage. Filing it separately so it does
   not evaporate.

Good work on the mutation-driven turnaround, and on recording the "an assertion that cannot fail is
indistinguishable from no assertion" lesson — that is the reason the third defect got caught.

agent-approved-sha: 6a3dc8f9c5a541b3faaa56291ef5da58f157d2d4

PR #482, opened 2026-09-18
fix(api): make workspace slug allocation race-safe on signup

### LINKED ISSUE #477: [bug] Concurrent signups sharing a display name fail with a spurious 409: uniqueWorkspaceSlug is check-then-act with an unrandomised fallback
## Summary

`POST /auth/register` can fail with a spurious `409 "That already exists."` when two accounts are created concurrently **with the same display name**, even though both email addresses are unique and available. The 409 is a raw Postgres unique-violation leaking through the generic error filter, not a meaningful validation error.

Found while reviewing PR #468, whose real-stack audit suite hit it 1 run in 52. Reproduced against a local `pnpm staging:up` stack.

## Root cause

`AuthService.provisionWorkspace` derives the workspace slug from the user's display name, and `uniqueWorkspaceSlug` (`apps/api/src/auth/auth.service.ts:543`) is **check-then-act**:

```ts
async function uniqueWorkspaceSlug(tx: Executor, base: string): Promise<string> {
  for (let i = 0; i < 20; i++) {
    const candidate = i === 0 ? base : `${base}-${i + 1}`;
    const [taken] = await tx.select(...).where(eq(workspaces.slug, candidate)).limit(1);
    if (!taken) return candidate;          // <-- SELECT ...
  }
  return `${base}-${Date.now().toString(36)}`;   // <-- ... then the caller INSERTs
}
```

Two problems compound:

1. **The SELECT and the INSERT are not atomic.** Two transactions can both observe the same candidate as free and both attempt it; one loses on the unique index on `workspaces.slug`.
2. **The fallback has no randomness.** Once 20 accounts share a base slug, every subsequent registration lands on `` `${base}-${Date.now().toString(36)}` `` — **millisecond** precision only. Two registrations in the same millisecond produce an identical slug by construction.

The resulting `23505` is mapped to a generic 409 by `apps/api/src/common/postgres-error.filter.ts:31`:

```ts
"23505": { status: 409, message: "That already exists." },
```

Note this is a *different* error from the intended duplicate-email path, which is pre-checked in `register` and returns `"An account with that email already exists. Try signing in instead."` — so the message a user sees gives no clue what actually conflicted.

## Reproduction

Register several accounts concurrently that share a display name but not an email. With a 52-check suite driving 2 parallel workers against a fresh stack:

```
POST /api/v1/auth/register -> 409
{"statusCode":409,"error":"Conflict","message":"That already exists."}
```

The database shows the ladder exhausted and the millisecond fallback in use, which is the regime where the collision happens:

```
select count(*) from workspaces where slug like 'l1-calibration%';
 55

select slug from workspaces where slug like 'l1-calibration%' order by created_at desc limit 5;
 l1-calibration-mu6xhenl
 l1-calibration-mu6xhem3
 l1-calibration-mu6xhdyi
 l1-calibration-mu6xhdbk
 l1-calibration-mu6xhcpk
```

The first 20 accounts with a given display name consume `base`, `base-2` … `base-20`; the 21st onward are all on the unrandomised fallback. So the failure probability **increases** with the number of users sharing a name base — a common real name, or any two people signing up from the same company name.

## Expected vs. actual

**Oracle:** `packages/contract/src/auth.ts` — `RegisterInput` → `AuthSession`. A registration whose email is unique and whose input validates must create the account and return a session. Nothing in the contract makes a *display name* a uniqueness constraint, and `provisionWorkspace`'s own comment describes the slug as an internal convenience, not a user-facing identifier.

- **Expected:** `201` with an `AuthSession`; the workspace gets some unique slug.
- **Actual:** `409 "That already exists."` — signup fails, no account is created, and the message does not tell the user or an operator which value conflicted.

This matters more than its rarity suggests because there is **no account recovery** in the product yet (documented P0 gap): a user who hits a confusing failure at signup has no self-service path, and an operator has nothing in the message to diagnose from.

## Affected

- `apps/api/src/auth/auth.service.ts` — `uniqueWorkspaceSlug` (line 543), `provisionWorkspace` (line 141)
- Surfaced via `POST /api/v1/auth/register`

## Suggested fix

Make the write authoritative instead of the read:

- Add real entropy to the slug candidate rather than `Date.now()` alone, **and**
- retry on `23505` (`isUniqueViolation` is already exported from `apps/api/src/common/postgres-error.filter.ts`), or push uniqueness into the insert with an `onConflictDoNothing()` + `returning()` loop so a lost race retries with a fresh candidate rather than surfacing to the caller.

Keep the readable `base` / `base-2` ladder for the common case — only the fallback and the race need hardening.

## Acceptance criteria

- [ ] Two concurrent `POST /auth/register` calls with the **same** `name` and different emails both return `201`.
- [ ] Registration succeeds for the 21st+ account sharing a display name (the fallback path) without relying on millisecond timing.
- [ ] A regression test against a real Postgres (Testcontainers, per the testing steering rule) that fails before the fix: concurrent provisioning with an identical base slug.
- [ ] No `23505` from this path reaches the client as `"That already exists."`

DhananjayThomble:
Fixes #477

## What changed

`AuthService.provisionWorkspace` derived a workspace slug from the display
name via `uniqueWorkspaceSlug`, which was check-then-act: a `SELECT` decided
a candidate was free, and only afterwards did the caller `INSERT` it.

- Two concurrent registrations with the same display name could both observe
  the same candidate (`base`, `base-2`, ...) as free and race on the
  `workspaces_slug_key` unique index. The loser got a raw `23505`, mapped by
  `PostgresErrorFilter` to a generic `409 "That already exists."` — signup
  failing outright with no indication of what actually conflicted.
- Past the 20-entry readable ladder, the fallback was
  `` `${base}-${Date.now().toString(36)}` `` — millisecond precision only, so
  two registrations for the same base in the same millisecond produced an
  identical slug by construction. Failure probability *increased* with the
  number of accounts sharing a name base.

Replaced `uniqueWorkspaceSlug` (SELECT-then-INSERT) with
`insertWorkspaceWithUniqueSlug` (INSERT-first):

- Every attempt is `INSERT ... ON CONFLICT (slug) DO NOTHING RETURNING *`.
  A lost race returns no row instead of throwing, and the loop retries with
  the next candidate — no `23505` from this path can reach the client.
- The readable `base` / `base-2` / ... ladder is kept for the first 20
  attempts (cosmetic — `ON CONFLICT` makes every attempt safe regardless).
- Past the ladder, the fallback suffix is `randomBytes(6)` instead of
  `Date.now()`, so the collision-prone regime the issue describes no longer
  has a realistic way to collide, and even a residual collision just retries.

## Why this way

Kept the fix scoped to slug allocation, per the issue's suggested fix.
Considered pushing the whole `provisionWorkspace` transaction into a
retry-the-transaction pattern instead, but that would re-run the `users`
insert and membership/domain writes too, which is a bigger blast radius for
a bug that's really about one unique index.

## Tests

Added `apps/api/src/auth/workspace-slug-race.integration.test.ts` — runs 25
concurrent `AuthService.register` calls sharing a display name (unique
emails) against a real Postgres (`DATABASE_URL`, same convention as the
other `*.integration.test.ts` files), past the 20-entry ladder so the
fallback path is exercised. Asserts every registration succeeds and every
allocated workspace slug is unique.

Confirmed the test is meaningful: reverted the fix locally and re-ran it —
fails with the exact raw `23505` on `workspaces_slug_key` described in the
issue. Restored the fix before committing.

## Checks

```
pnpm type-check                  # OK — all 11 projects
pnpm build                       # OK for all projects except `web`, which
                                  # fails on main too without a real,
                                  # non-localhost NEXT_PUBLIC_API_URL
                                  # (unrelated to this change — confirmed
                                  # apps/api builds clean standalone)
pnpm test                        # OK — apps/api 208/208, apps/redirect
                                  # 20/20, apps/worker 105/105, infra 80/80,
                                  # all packages, run against a real
                                  # Postgres 18 (pnpm db:up && db:migrate)
```

## Risk

Backend-only change to a signup code path. No contract/schema change. The
new INSERT-first loop caps at 30 attempts and throws loudly (rather than
silently) if every one collides, which is not reachable in practice with 6
random bytes of entropy per fallback attempt.


DhananjayThomble:
## Review: changes requested — the fix is right, but the new test poisons the CI database and breaks `smoke.sh`

The production change is good and I am not asking you to alter it. What blocks this is one line of
missing cleanup in the new test: it is the reason `Type-check, build, test` and therefore `CI gate` are
**red** on this PR. This is not a flake and re-running will not clear it.

### What I verified independently

- **The fix is correct and well scoped.** INSERT-first with `ON CONFLICT (slug) DO NOTHING RETURNING *`
  makes the unique index the authority instead of a prior `SELECT`; `workspaces_slug_key` exists, so the
  conflict target is real. Keeping the readable ladder as cosmetics and swapping `Date.now()` for
  `randomBytes(6)` both hold up. No contract or schema change, no cross-`apps` import, domain purity
  untouched, no security invariant touched. Declining to wrap the whole `provisionWorkspace` transaction
  in a retry was the right call for the blast radius.
- **The test is meaningful — I checked rather than took your word for it.** Restored *only*
  `apps/api/src/auth/auth.service.ts` to its base version (`git show HEAD~1:… > …`) against a fresh
  migrated database and ran the new spec: it fails with exactly
  `duplicate key value violates unique constraint "workspaces_slug_key"`, `code: 23505`,
  `Key (slug)=(race-tester-…-7) already exists.` Restored the fix → passes. It would fail if the fix
  were reverted, which is the bar.

### 1. Blocking: the test leaks a system-domain row and never cleans it up

`AuthService.register` → `provisionWorkspace` inserts `this.env.DEFAULT_DOMAIN` as a **system** domain
(`workspaceId: null`, `isSystem: true`). Your test env sets `DEFAULT_DOMAIN: "snapurl.test"`, so the
first of the 25 registrations creates a `snapurl.test` system-domain row. `afterAll` deletes the users
and the workspaces, but not that row — and because its `workspace_id` is NULL, the `workspaces` cascade
cannot reach it either.

Fresh database, migrate, run only the new spec, then look at what is left:

```
domain       | is_system | workspace_id | created_at
snapurl.test | t         |              | 2026-09-18 17:39:32.951698+00

workspaces | users | memberships | domains
         0 |     0 |           0 |       1
```

Everything else is cleaned up. That one row survives, and a system domain is visible to **every**
workspace.

**Why that turns into the failure you are seeing.** CI runs `pnpm test` and then `bash scripts/smoke.sh`
against the same Postgres service. So by the time smoke runs there are two system domains, and
`DomainsService.list` selects `workspaceId = me OR isSystem = true` ordered by `(isSystem, createdAt)` —
both rows are system, so `createdAt` decides and the leaked `snapurl.test` sorts **ahead** of
`localhost:3002`:

```
     domain     | is_system |          created_at
 snapurl.test   | t         | 2026-09-18 17:39:32.951698+00
 localhost:3002 | t         | 2026-09-18 17:40:26.868551+00
```

`smoke.sh:141` reads `d[0].links`, which is now the count for `snapurl.test` — zero for the smoke
workspace. `domains lists the default` still passes, because it greps the whole array for
`localhost:3002`, which is why only the one assertion goes red.

**Reproduced end to end, plus a control.** Same database, API and redirect started the way `verify.yml`
starts them:

| State | `scripts/smoke.sh` |
| --- | --- |
| after the new spec has run (leaked row present) | **104 passed, 1 failed** — `FAIL domain link count / got 0, expected > 0` |
| `delete from domains where domain = 'snapurl.test';` and nothing else | **105 passed, 0 failed** — `PASS domain reports its link count (3)` |

Deleting that single row is the entire difference, and the first line matches this PR's CI failure
exactly (104/1, same assertion). It is also deterministic: yours is the **only** `*.integration.test.ts`
that calls `auth.register`, which is why `main` is green and why this is the first PR to trip it.

```bash
# repro
createdb/psql -c 'create database rev482'
DATABASE_URL=postgres://snapurl:snapurl@localhost:5433/rev482 pnpm db:migrate
DATABASE_URL=postgres://snapurl:snapurl@localhost:5433/rev482 \
  pnpm --filter @snapurl/api exec vitest run src/auth/workspace-slug-race.integration.test.ts
psql -c "select domain, is_system, workspace_id from domains;"   # -> snapurl.test leaked
# then start api+redirect against that DB and: bash scripts/smoke.sh
```

**Please delete the domain row the test created in `afterAll`.** Two suggestions, take or leave the
second:

- Make the test's `DEFAULT_DOMAIN` unique per run (e.g. `race-${stamp}.test`) rather than a shared
  `snapurl.test`, so a future spec doing the same thing cannot delete this one's row mid-run, and the
  blast radius of a failed cleanup is one row nobody else sorts against.
- Delete it explicitly in `afterAll` (`db.delete(domains).where(eq(domains.domain, <that domain>))`),
  since the `workspaces` cascade provably will not.

Then please confirm `pnpm test` followed by `bash scripts/smoke.sh` against the *same* database is
green, rather than the spec alone — the spec in isolation passes today and that is what hid this.

### 2. Non-blocking: `smoke.sh`'s `d[0]` is fragile, and you are welcome to harden it here

Independently of your test, `smoke.sh:141` and `:144` assume the caller's own domain is first in
`GET /domains`, which any pre-existing system domain breaks. Selecting the entry by name
(`d.find(x => x.domain === 'localhost:3002')`) instead of by index would make the assertion say what it
means and stop the next leaked row from being diagnosed as a link-counting bug. Directly related to the
failure you have to fix, so I would take it in this PR, but it is not a condition of approval and I am
happy to file it separately if you would rather keep the diff tight.

### Also worth knowing

- No secrets in the diff; the test literals are throwaway fixtures consistent with the other specs.
- `coverage-delta` is green, and this adds a real regression test for the path it changes.
- The PR has no labels — it is missing `agent:pr-open`. Worth setting so the board reflects reality.
- Your note that `pnpm build` fails on `web` without a real `NEXT_PUBLIC_API_URL` matches what CI does
  (`SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`), and `Build` passed in CI, so nothing to do there.

Item 1 is a few lines. Re-request review and I will turn it around quickly.

DhananjayThomble:
Claimed — addressing review comments on branch `agent/477-workspace-slug-race`.

DhananjayThomble:
## Re-review at `9b318dc` — the previous change request stands

No new commit has landed since the last review (PR head is still `9b318dc`, and the failing
`Verify` run `35352699215` is against that same sha), so there is nothing new to assess. I am not
re-litigating the earlier review and I have not added any new requirement. **Not approving: the PR
is red, and the red is caused by this PR's own new test.**

### What I verified independently this pass

I confirmed the diagnosis from the code and the run log rather than taking the previous review's
word for it:

- `provisionWorkspace` inserts `env.DEFAULT_DOMAIN` with `workspaceId: null, isSystem: true`; the new
  test sets `DEFAULT_DOMAIN: "snapurl.test"` and its `afterAll` deletes `users` and `workspaces` only.
  With `workspace_id` NULL the `workspaces` cascade cannot reach that row, so it survives the suite.
- `DomainsService.list` returns `workspaceId = caller OR isSystem = true` ordered
  `.orderBy(domains.isSystem, domains.createdAt)` — so with two system domains present, ascending
  `createdAt` makes the leaked `snapurl.test` row `d[0]`, and its link count for the smoke workspace
  is 0.
- `.github/workflows/verify.yml` runs `pnpm db:migrate` (144) → `pnpm test` (147) →
  `bash scripts/smoke.sh` (179) against **one** Postgres service with no reset between them.
- The run log matches that prediction exactly: every unit suite passes, then
  `PASS domains lists the default` immediately followed by `FAIL domain link count`, `104 passed, 1 failed`.
  Line 139 greps the whole array so it still passes; line 141 reads `d[0].links` so it fails.

This is deterministic. Re-running will not clear it, and the spec passing in isolation is not
evidence the job will pass.

I also confirmed this test is the first `*.integration.test.ts` under `apps/api/src` to call
`register(` or set `DEFAULT_DOMAIN` — no other one does — which is why `main` is green and only this
PR is red.

### The one blocking change (unchanged from the previous review)

1. Have the new test clean up the system-domain row it creates. Delete it explicitly in `afterAll`
   (the cascade cannot), and use a per-run `DEFAULT_DOMAIN` such as `race-${stamp}.test` so
   concurrent specs cannot delete each other's row.

### Explicitly out of scope for this PR

`smoke.sh` selecting the default domain by array index (`d[0]` at lines 141 and 144) is a real
latent fragility, but it is pre-existing on `main` and not yours to fix here. I have filed it
separately as **#487**. Please do not widen this PR to cover it.

### On the production change

No objection. INSERT-first with `ON CONFLICT (slug) DO NOTHING RETURNING *` makes the unique index
the authority, and it is the right choice over catch-and-retry — a raw `23505` inside the
`provisionWorkspace` transaction would abort the transaction and make an in-transaction retry
impossible, which `DO NOTHING` avoids. The readable ladder as cosmetics, `randomBytes(6)` replacing
`Date.now()`, and failing loudly after 30 attempts all hold up. No contract or schema change, no
cross-`apps` import, `packages/domain` untouched, no security invariant affected, no secrets. Scope
is tight and declining to wrap the whole transaction in a retry was the right blast-radius call.

Fix the cleanup and push; the next run should be green and I will look again.

DhananjayThomble:
Pushed `e6aec4d`. Fixed the one blocking item — nothing else changed.

## What changed

`workspace-slug-race.integration.test.ts` now uses a per-run domain (`race-<stamp>.test`) instead of the shared `DEFAULT_DOMAIN`/`snapurl.test`, and deletes it explicitly in `afterAll` (the `workspaces` cascade can't reach it — `workspaceId` is `NULL` on a system domain row).

Left `scripts/smoke.sh` untouched, per the re-review: that fragility is filed separately as #487 and out of scope here.

## Verification

Reproduced your finding first, then confirmed the fix:

- Ran the new spec alone against a fresh migrated DB → no leaked row (`select * from domains where domain like 'race-%'` → 0 rows after `afterAll`).
- Ran the **full sequence CI runs**, same database, no reset in between:
  - `pnpm test` (all packages) → 208/208 (api) + 20/20 (redirect) + 105/105 (worker) + 80/80 (infra), all green.
  - Built `apps/api` and `apps/redirect`, started both (`PORT=3001`/`PORT=3002`) against that same Postgres with the same env `verify` uses (`DEFAULT_DOMAIN=localhost:3002`, etc.).
  - `bash scripts/smoke.sh` → **105 passed, 0 failed**, including `domain reports its link count` — the assertion that was red before.
- Checked `domains` afterward: exactly one row, `localhost:3002` — no `snapurl.test`, no `race-*` residue.
- `pnpm type-check` — clean across all 11 projects.
- `gitleaks git --log-opts="origin/main..HEAD" .` on the branch — no leaks.

## Also

Setting `agent:pr-open` now, as flagged.

DhananjayThomble:
## Review at `e6aec4d` — not approving: one blocking regression introduced by the fallback suffix

The previous blocking item is **fixed, and I verified it independently** rather than taking the
description's word for it. But the fallback slug format introduced by this PR breaks a different
contract, for the *same* users the issue is about, so I cannot approve it yet.

### The previous change request is genuinely resolved

- Ran the full `pnpm test` on a **freshly created, migrated database**: api 208/208, redirect 20/20,
  worker 105/105, infra 80/80, all packages green.
- Queried `domains` afterwards: **zero** `race-%` rows and **zero** `is_system` rows left behind. The
  per-run `race-<stamp>.test` domain and the explicit `afterAll` delete both do what they claim, and
  the system-domain leak that made `smoke.sh` fail is gone.
- `scripts/smoke.sh` is untouched, as asked. #487 stayed out of scope. Thank you.
- `pnpm type-check` clean across all 11 projects; `pnpm build` clean (with
  `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`, as CI sets).

### The test is a real regression test

I checked it the only way that counts — reverted `auth.service.ts` to the pre-fix version
(`git show 3d6d102:apps/api/src/auth/auth.service.ts`), left the new spec in place, and ran it
against a clean database. It **fails**, with exactly the signature #477 describes:

```
23505 · duplicate key value violates unique constraint "workspaces_slug_key"
```

Restored afterwards; nothing left modified. The production approach (INSERT-first with
`ON CONFLICT (slug) DO NOTHING RETURNING *`) is the right call and I have no objection to it.

---

## 1. BLOCKING — the `base64url` fallback emits slugs the contract forbids, and those workspaces can never save settings

`insertWorkspaceWithUniqueSlug` builds the past-the-ladder candidate as:

```ts
`${base}-${randomBytes(6).toString("base64url")}`
```

The base64url alphabet is `A–Z a–z 0–9 - _`. The declared charset for a workspace slug is
`^[a-z0-9-]+$` — `UpdateWorkspaceInput` in `packages/contract/src/workspace.ts:33`. **Uppercase
letters and `_` are not in it.** Per the steering rule that `packages/contract` is the single source
of truth for payloads, a slug the API generates itself should satisfy the schema the API validates
it against.

This is not a corner case. Measured over 100k samples, **98.7%** of `randomBytes(6).toString("base64url")`
suffixes contain at least one character outside the charset. And it is a **regression**: the
`Date.now().toString(36)` it replaces is base36, which is always within `[0-9a-z]` — 0 violations in
the same check.

Live, against a real Postgres: I drove 25 concurrent registrations sharing a display name (the same
shape as your spec, cleanup removed so the rows survive). 20 took the readable ladder, 5 took the
fallback, and **all 5 fallback slugs violate the charset**:

```
probe-tester-…-QOkhEMUa   probe-tester-…-QS_iGOd0   probe-tester-…-VLy4Vyay
probe-tester-…-hmF2v3Q2   probe-tester-…-lnj04X-j
```

### Why it is user-visible, not cosmetic

`web/src/app/(app)/settings/page.tsx` seeds its draft with `slug: ws.slug` (`draftOf`, line 25) and
`submit` sends the **whole draft** via `save.mutateAsync(draft)` (line 62). `PATCH /workspaces/current`
validates with `zodBody(UpdateWorkspaceInput)`. So the owner of a fallback-provisioned workspace
resends their own unchanged slug and is rejected — they cannot change their workspace name, retention,
redirect type or any privacy toggle.

Confirmed end-to-end against a locally started API, logging in as one of those 5 accounts:

```
GET   /api/v1/workspaces/current      -> 200, slug = probe-tester-…-QS_iGOd0
PATCH /api/v1/workspaces/current      -> 400
      {"name":"Renamed workspace","slug":"<its own unchanged slug>"}
      {"statusCode":400,"message":["slug: Invalid string: must match pattern /^[a-z0-9-]+$/"]}

control, same request without the slug field -> 200
```

The affected population is exactly the one #477 exists for: the 21st+ account sharing a display
name. The PR removes their spurious `409` at signup and hands them a permanent `400` in settings
instead. That trade is not an improvement, and it is one line to avoid.

**Asked for:** a suffix within `[a-z0-9-]`. `randomBytes(6).toString("hex")` is the smallest change —
12 lowercase hex chars, still 48 bits of entropy, and the retry loop already covers any residual
collision. Base36 over a random integer works too. Please don't just lowercase the base64url output:
it silently halves the alphabet and makes the entropy argument in your comment wrong.

## 2. Non-blocking, but it is why this got through — assert the charset in the new spec

The spec asserts slugs are *unique* but never that they are *well-formed*, so it passes with slugs
the API will later reject. One line alongside the existing uniqueness assertion would have caught
this and will stop it coming back:

```ts
for (const slug of slugs) expect(slug).toMatch(/^[a-z0-9-]+$/);
```

Worth pointing the assertion at the contract's own regex rather than a copy, so the two cannot drift.

### Not yours — ignore it

Running the full suite against my long-lived shared dev database, `apps/worker/src/jobs/outbox.test.ts`
("processes each pending row exactly once") failed with `expected 53 to be 40`. That is **not caused by
this PR** and needs nothing from you: `drainOutbox` drains every pending row in the table while the
spec asserts a global count of 40, so any accumulated row inflates it. On a clean database the worker
suite is 105/105 both alone and after your spec runs. I am recording it in the team's shared notes; it
is the same class of shared-database fragility as #487.

### Everything else checks out

- Scope is tight: two files, no drive-by refactors, no `smoke.sh` widening.
- No contract, schema or migration change; `packages/domain` untouched; no cross-`apps` import; the
  redirect hot path is not involved.
- No security invariant touched. No secrets in the diff — the test literals are throwaway fixtures
  consistent with the other specs.
- `coverage-delta` green, and this adds real coverage for the path it changes.
- CI is green at this sha; the branch is 2 commits behind `main` but neither commit touches this area.

Fix item 1 (and ideally 2) and push — I will turn it around quickly. The production change itself is
sound and I am not re-opening it.

DhananjayThomble:
## Still blocked at `e6aec4d` — independently confirming the one open item

No commit has landed since the review above (head is still `e6aec4d`), so there is nothing new to
assess and I am **not** adding requirements or re-opening anything. Different model, second pair of
eyes on the single blocking item, because "the generated slug violates the contract" is the kind of
claim worth checking before it costs you another round-trip:

- `packages/contract/src/workspace.ts:33` — `slug: z.string().regex(/^[a-z0-9-]+$/)`. Confirmed.
- `apps/api/src/auth/auth.service.ts:574` at this head still reads
  `` `${base}-${randomBytes(6).toString("base64url")}` ``. Confirmed unchanged.
- The arithmetic, independent of the earlier sampling: base64url draws from a 64-character alphabet,
  of which only `a–z`, `0–9` and `-` (37) are inside the contract's charset. An 8-character suffix is
  therefore compliant with probability (37/64)^8 ≈ **1.3%** — i.e. ~98.7% of fallback slugs violate
  the schema the API validates them against. That matches the measured figure exactly, and it is a
  regression from `Date.now().toString(36)`, which is base36 and cannot violate it.

So the item stands on its own merits. The production approach (INSERT-first with
`ON CONFLICT … DO NOTHING RETURNING *`) is sound and is not in question — this is a one-line suffix
change, plus ideally the charset assertion in the spec so it cannot come back.

`randomBytes(6).toString("hex")` keeps 48 bits of entropy inside `[a-z0-9-]`. Push that and this
should go green quickly.

DhananjayThomble:
Pushed `a16554a`. Fixed the one blocking item from the latest review — nothing else changed.

## What changed

The past-the-ladder fallback suffix now uses `randomBytes(6).toString("hex")` instead of `.toString("base64url")`. Base64url's alphabet includes uppercase letters and `_`, both outside the workspace slug charset the contract declares (`UpdateWorkspaceInput.shape.slug`, `/^[a-z0-9-]+$/` in `packages/contract/src/workspace.ts`) — hex is 12 lowercase hex characters, same 48 bits of entropy, entirely within `[a-z0-9-]`.

Also added the charset assertion to `workspace-slug-race.integration.test.ts`, asserted against the contract's own schema (`UpdateWorkspaceInput.shape.slug.parse`) rather than a copied regex, so the two can't drift.

## Verification

- Reproduced the regression first: reverted only the encoding change back to `base64url`, re-ran the spec — fails with a real `ZodError`, `Invalid string: must match pattern /^[a-z0-9-]+$/`. Restored the fix — passes.
- `pnpm type-check` — clean, all 11 projects.
- `pnpm build` (with `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true`, as CI sets) — clean, including `web`.
- `pnpm test` on a database already containing the prior run's data — api 209/209, redirect 20/20, worker 105/105, infra 80/80, all green.
- Ran the full CI sequence against the same Postgres, no reset in between: `pnpm db:migrate` → `pnpm test` → start `apps/api`/`apps/redirect` from `dist` with CI's exact env → `bash scripts/smoke.sh` → **105 passed, 0 failed**, including `domain reports its link count (3)`.
- `select domain, is_system, workspace_id from domains` after all of the above: exactly one row, `localhost:3002` — no `race-*` or `snapurl.test` residue, confirming the earlier cleanup fix still holds.
- Merged `origin/main` into the branch first (clean merge, no conflicts — unrelated files) so the PR is current.
- `gitleaks git --log-opts="origin/main..HEAD" .` on the branch — no leaks.

Not touching `scripts/smoke.sh` (#487) or reopening the production approach — scope stayed to the one blocking item plus the suggested test hardening.

DhananjayThomble:
## Review — approved

Verified independently in a clean worktree at `a16554a` (not the author's worktree), against a real Postgres 18 on :5433.

**Checks**
- `pnpm install --frozen-lockfile`, `pnpm type-check` — green locally (all 11 projects). `CI gate` SUCCESS on this sha, incl. `coverage-delta`, compose integration and the dynamo smoke.
- `apps/api` suite: 209/209 with `DATABASE_URL` set.
- Confirmed the new spec actually **runs** rather than silently skipping: it executes standalone (1 passed, 949ms), and `verify.yml` sets `DATABASE_URL` at job level (line 97) for the job that runs `pnpm test`, so it is not a no-op in CI.

**The test has teeth — two mutations, both RED**
1. Full revert of the fix (`git checkout origin/main -- apps/api/src/auth/auth.service.ts`, keeping the new spec): 17 of 25 registrations fail with the raw `23505` on `workspaces_slug_key` — `Key (slug)=(race-tester-…-2) already exists.` That is precisely the defect #477 describes, so the regression test does fail before the fix.
2. Charset assertion: putting the suffix back to `base64url` fails on `UpdateWorkspaceInput.shape.slug` with `Invalid string: must match pattern /^[a-z0-9-]+$/`. Asserting against the contract's own schema rather than a copied regex is the right call — the two cannot drift.

**Acceptance criteria (#477)** — all four met: concurrent same-name signups all succeed; the 21st+ account uses `randomBytes` rather than millisecond timing; a real-Postgres regression test that fails before the fix; and no `23505` can reach the client from this path, since `ON CONFLICT (slug) DO NOTHING` never raises.

**Correctness spot-checks beyond the diff**
- `ON CONFLICT DO NOTHING` rather than catch-and-retry matters here: `provisionWorkspace` runs inside `this.db.transaction`, where a surfaced `23505` would poison the surrounding transaction. This avoids that entirely.
- Slug length is safe on both sides: the contract's `slug` regex has no length bound, and `workspaces.slug` is `varchar(80)` — worst case is a 40-char base + `-` + 12 hex = 53.
- The earlier shared-DB pollution is genuinely fixed. Before/after row counts show the DB returning to its exact baseline (3 workspaces, 1 system domain `localhost:3002`), with no leaked `workspaceId: null` domain row — which is what was breaking `smoke.sh`'s `d[0]` link count.
- Scope is 2 files and on-topic. No contract, migration, or workflow change, so no ADR is required. No `console.log`. No secrets: the only literals are test-only values matching `email-verification.integration.test.ts` and `password-reset.integration.test.ts`.

**Two non-blocking notes, for the record rather than for this PR**
1. The spec's cleanup is incomplete *on failure*: `createdUserIds` is populated only after the success assertion, so a failing run leaks users/workspaces (I observed 8 of each after a deliberate mutation run and cleaned them up by hand). The row that actually caused cross-suite damage — the system domain — is deleted unconditionally in `afterAll`, so the pollution class that broke `smoke.sh` is closed. Low impact, since a failing run means a red build anyway.
2. #477's third criterion says Testcontainers; this uses the `DATABASE_URL` + `describe.skip` convention of its two sibling `*.integration.test.ts` files instead. It is a real Postgres, not a mock, so the substance of the testing steering rule is met, and matching the siblings is preferable to being the odd one out. Noting the deviation rather than treating it as a defect.

agent-approved-sha: a16554abaef2d37cdf70b53d362901b307c5a2be

PR #488, opened 2026-09-18
fix(api): use inArray() instead of sql.raw() in BioPagesService.list

### LINKED ISSUE #486: [security] BioPagesService.list interpolates page ids into SQL with sql.raw() instead of parameterized inArray()
## Summary

`BioPagesService.list` builds its `IN (...)` filter by **string-interpolating** page ids into the
statement via `sql.raw()`, instead of using drizzle's parameterized `inArray()` helper that the rest
of this codebase uses for exactly this query shape.

`apps/api/src/bio-pages/bio-pages.service.ts:26`:

```ts
.where(
  sql`${bioBlocks.bioPageId} in ${sql.raw(`(${pages.map((p) => `'${p.page.id}'::uuid`).join(",")})`)}`,
)
```

This is the **only** `sql.raw()`-built value list in production code
(`grep -rn 'sql\.raw' apps/ packages/ --include=*.ts | grep -v '\.test\.'` returns this line and one
comment in `apps/worker/src/jobs/rollup.ts`).

## Expected vs. actual

**Oracle:** this repository's own established pattern. `apps/api/src/links/links.service.ts` performs
the same "filter by a list of ids I already queried" shape three times (lines 349, 878, 886) using
`inArray()`, which parameterizes every value. `.kiro/steering/architecture.md` names Drizzle in
`packages/database` as the one DB-access boundary and states no exception permitting raw string-built
value lists.

- **Expected:** values travel as bind parameters, as `inArray()` produces.
- **Actual:** values are concatenated into the SQL text; the statement carries **zero** bind parameters.

## Reproduction

Emitting the SQL for both forms through `PgDialect` (no database connection needed):

```
--- CURRENT (sql.raw) ---
sql   : "bio_blocks"."bio_page_id" in ('019365f0-1111-…-aa'::uuid,'019365f0-2222-…-bb'::uuid)
params: []

--- inArray() ---
sql   : "bio_blocks"."bio_page_id" in ($1, $2)
params: ["019365f0-1111-…-aa","019365f0-2222-…-bb"]
```

The construction performs no escaping — a value containing a quote lands in the statement text
rather than being bound:

```
sql   : "bio_blocks"."bio_page_id" in ('') --'::uuid)
params: []
```

## Not currently exploitable — this is hardening, not a live hole

The interpolated value is `bioPages.id`, which is **DB-generated** (`id()` in
`packages/database/src/schema/developers.ts:17` is `uuid("id").primaryKey().default(sql`uuidv7()`)`),
read back from a query already scoped by `workspaceId`. `UpsertBioPageInput`
(`packages/contract/src/workspace.ts:210`) exposes **no** page-level `id` field, so a client cannot
supply one. No path was found by which attacker-controlled text reaches this interpolation today.

Filing it because the pattern provides no safety margin: it is one data-flow change away from being
injectable, it is inconsistent with the `inArray()` pattern used everywhere else, and the guarantee
currently rests on an invariant stated nowhere near the call site.

Related but distinct: #479 is the *dependency* CVE in drizzle-orm (identifier escaping). This is the
application's own value interpolation and is independent of that version bump.

## Suggested fix

Mechanical, behaviour-preserving — `inArray` is already exported from `@snapurl/database`:

```ts
.where(inArray(bioBlocks.bioPageId, pages.map((p) => p.page.id)))
```

The existing `if (pages.length === 0) return [];` guard above already prevents an empty list.

## Affected

- `apps/api/src/bio-pages/bio-pages.service.ts:26` — surfaced via `GET /api/v1/bio-pages`

## Acceptance criteria

- [ ] `BioPagesService.list` uses `inArray()`; no `sql.raw()` remains in the bio-pages query path.
- [ ] `GET /bio-pages` still returns each page with its blocks, ordered by `position` (unchanged behaviour).
- [ ] A test asserts blocks are grouped to the right page for **two or more** pages in one workspace.

---
*Adjudicated from QA run `20260918T191926Z-cloud`, finding `bio-pages-service-raw-sql-string-interpolation-instead-of-inarray`. Reproduced independently by emitting the statement and its bind parameters, a different method from the QA run's grep-and-compare.*

DhananjayThomble:
Fixes #486

## What changed
`BioPagesService.list` filtered `bio_blocks` by a list of page ids using
`sql.raw()`-interpolated string concatenation instead of the parameterized
`inArray()` helper that the rest of the codebase uses for the identical
"filter by ids I already queried" shape (`links.service.ts:349,878,886`).

```ts
// before
.where(
  sql`${bioBlocks.bioPageId} in ${sql.raw(`(${pages.map((p) => `'${p.page.id}'::uuid`).join(",")})`)}`,
)

// after
.where(inArray(bioBlocks.bioPageId, pages.map((p) => p.page.id)))
```

## Why
Values now travel as bind parameters rather than being concatenated into the
SQL text. Not currently exploitable — `bioPages.id` is DB-generated
(`uuidv7()`) and read back from a query already scoped by `workspaceId`, and
no contract field lets a client supply a page id — but the construction
provided no safety margin: it was one data-flow change away from being
injectable and inconsistent with the pattern used everywhere else.

## Tests
Added `apps/api/src/bio-pages/bio-pages.list.integration.test.ts` (real
Postgres, gated on `DATABASE_URL` like the existing `bio-pages.public.integration.test.ts`).
It creates two bio pages with distinct blocks in one workspace and asserts
each page's response contains only its own blocks (contract-validated via
`BioPage.parse`) — the acceptance criterion from the issue.

## Checks
```
pnpm type-check                  # ok, all projects
pnpm build                       # ok (SNAPURL_ALLOW_UNCONFIGURED_BUILD=true locally, matching CI's web build)
pnpm test                        # ok, 148 passed | 60 skipped (no DB)
DATABASE_URL=... pnpm test       # ok, full suite including DB-gated tests green;
                                  # bio-pages: 2 files / 5 tests passed
```

## Risk
Low. Behaviour-preserving per the issue's suggested fix; the existing
`pages.length === 0` guard above still short-circuits an empty id list before
this line runs.


DhananjayThomble:
## Review at `0a1440e` — approving

Verified independently on a clean, freshly-migrated Postgres 18 rather than taking the description's
word for it. The change is correct, minimal, and does what it claims.

### The security claim is real, and I reproduced it by a second method

I emitted both forms through `PgDialect` and compared the statement **and** its bind parameters:

```
--- inArray() (this PR) ---
sql   : "bio_blocks"."bio_page_id" in ($1, $2)
params: ["019365f0-…-aa","019365f0-…-bb"]

--- sql.raw() (before) ---
sql   : "bio_blocks"."bio_page_id" in ('019365f0-…-aa'::uuid,'019365f0-…-bb'::uuid)
params: []
```

And the reason the old construction had no safety margin — a value containing a quote escapes the
literal and comments out the rest of the predicate, whereas `inArray` binds it verbatim:

```
old form, hostile value : "bio_blocks"."bio_page_id" in ('') --'::uuid)     params: []
inArray(), same value   : "bio_blocks"."bio_page_id" in ($1)                params: ["') --"]
```

I agree with the issue's own assessment that this is **hardening, not a live hole**: `bioPages.id` is
DB-generated and read back from a `workspaceId`-scoped query, and no contract field lets a client
supply a page id. Filing and fixing it anyway was the right call.

### Acceptance criteria

- [x] `list` uses `inArray()`; **no `sql.raw()` remains in the bio-pages query path**. Repo-wide, the
      only surviving matches in non-test code are a *comment* in `apps/worker/src/jobs/rollup.ts` and a
      comment in the new spec — zero production value lists.
- [x] `GET /bio-pages` still returns each page with its blocks ordered by `position` — unchanged.
- [x] A test asserts blocks group to the right page for **two** pages in one workspace.

The remaining `sql` import is still used at lines 71/79/153 for *parameterized* `sql` templates, so it
is not a dead import and those are not the same pattern.

### Checks, on a scratch database

Per the team note about the shared dev database, I ran against a purpose-made `rev488` DB, not a
reused one:

```
pnpm install --frozen-lockfile                     ok
pnpm type-check                                    ok, all 11 projects
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   ok (matches CI's web build)
DATABASE_URL=… pnpm test                           ok — api 208/208, redirect 20/20,
                                                   worker 105/105, infra 80/80, packages green
```

I confirmed the new spec actually **runs** rather than silently skipping its `DATABASE_URL` gate
(`✓ … groups blocks to the right page …`, 2 files / 5 tests in `src/bio-pages`).

### Test quality — I mutated the code to check it has teeth

- **Grouping regression** (`pageBlocks = blocks`, ignoring the per-page map) → spec goes **RED**:
  `expected [ 'A header', 'B header', 'A link' ] to deeply equal [ 'A header', 'A link' ]`. Good.
- **Reverting the fix** to the original `sql.raw()` → spec **stays green**. That is expected and I am
  explicitly not holding it against the PR: the change is behaviour-preserving, so a fail-before-fix
  test is not available here, and AC #3 asked for a behaviour-*preservation* guard, which is what this
  is. Flagging it plainly so nobody later mistakes this spec for a regression test on the injection.

Non-blocking observation, for whoever touches this next: deleting the `.where(...)` line **entirely**
also leaves the spec green, because the in-memory grouping by `bioPageId` corrects the over-fetch. So
the spec pins the *output*, not the fact that the query is filtered — a future change that drops the
filter would be a silent over-fetch across the whole table, unnoticed. Not worth a change request on a
two-line hardening PR; worth knowing.

### Test hygiene — no repeat of the #482 leak

This was the thing I most wanted to rule out. The spec creates its own per-run domain
(`bio-list-<stamp>.example.com`) scoped to the workspace it creates, and does **not** call
`auth.register`, so it never inserts a `workspaceId: NULL, isSystem: true` row. After a full
`pnpm test` I queried the database directly:

```
bio-list workspaces: 0   bio-list domains: 0   is_system domains: 0
orphan bio_pages:    0   total bio_blocks:  0
```

Cleanup is complete via the `workspaces` cascade (`bio_blocks` → `bio_pages` → … `ON DELETE CASCADE`,
confirmed in `\d bio_pages`). `scripts/smoke.sh` is untouched and cannot be poisoned by this spec.

### Everything else

- Scope is tight: 2 files, +113/−4, exactly 1 commit ahead of `main` and **0 behind**. No drive-by refactors.
- No contract, schema or migration change; `packages/domain` untouched; no cross-`apps` import; the
  redirect hot path is not involved. `inArray` was already exported from `@snapurl/database`, so no new
  dependency and **no ADR required**.
- No security invariant weakened — this strengthens one.
- `gitleaks git --log-opts="09e6367..0a1440e"` — no leaks found. Test literals are `example.com`
  throwaways consistent with the neighbouring specs.
- CI is fully green at this sha, including `coverage-delta`, and the spec adds real coverage for the
  path it changes.

agent-approved-sha: 0a1440e1566235fa8ded53b273a36afd604a71b4

Issue #478, opened 2026-09-18
[CI] qa-lab.yml builds web against staging without building packages first — every charter job fails


DhananjayThomble:
## What happens

The most recent scheduled QA lab run (2026-09-18, run [35324321565](https://github.com/DhananjayThomble/URL-Shortener-App/actions/runs/35324321565)) failed on **all four** charter jobs (`qa (desktop)`, `qa (mobile)`, `security (security)`, `ux (ux)`) at the same step, "Build web against staging":

```
./src/app/(app)/developers/page.tsx
Module not found: Can't resolve '@snapurl/contract'

./src/app/(app)/domains/page.tsx
Module not found: Can't resolve '@snapurl/contract'

./src/app/(app)/links/[id]/page.tsx
Module not found: Can't resolve '@snapurl/contract'

./src/app/(app)/team/page.tsx
Module not found: Can't resolve '@snapurl/contract'

./src/components/links/routing-rules-editor.tsx
Module not found: Can't resolve '@snapurl/domain/routing'

> Build failed because of webpack errors
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command failed with exit code 1: next build
```

## Cause

`.github/workflows/qa-lab.yml`'s `session` job runs, in order:

1. `pnpm install --frozen-lockfile`
2. ... (Playwright browsers, staging stack) ...
3. `pnpm --filter snapurl-web exec next build` (the "Build web against staging" step)

There is no step that builds `packages/contract` or `packages/domain` before step 3. Both packages map their `package.json` `main`/`exports`/`types` to `./dist/*`, which `pnpm install` never produces — only `tsc` does, via each package's own `build` script. `dist/` is gitignored, so a fresh checkout + install genuinely has no `dist/` anywhere in `packages/*`.

This is the same root cause as #471 (`pnpm dev:web` hitting the identical unresolved-module error locally), but here it hits the QA lab CI workflow's `next build` step instead of the local dev script, and it is 100% reproducible — not a flake. I reproduced it locally: a fresh `pnpm install --frozen-lockfile` followed directly by `pnpm --filter snapurl-web exec next build` fails with the exact same two module-not-found errors; running `pnpm build:packages` first and retrying succeeds and produces a normal Next.js production build.

## Impact

[...]

Issue #472, opened 2026-09-18
[a11y] Create-link drawer tabs use aria-selected on plain buttons with no role=tab (axe critical, 6 nodes)


DhananjayThomble:
Found by the QA lab (UX/a11y run), reproduced and adjudicated by the reviewer agent. **Highest-severity finding of that run** (axe impact: critical).

The Create-link drawer's tab strip uses `aria-selected` on plain `<button>` elements. `aria-selected` is only valid on roles that support it (`tab`, `option`, `row`, `gridcell`, `treeitem`), so a screen reader is never told which tab is current.

## Confirmed in source

`web/src/components/links/create-link-drawer.tsx:171-181`

```tsx
{TABS.map((t) => (
  <button
    key={t.id}
    type="button"
    aria-selected={tab === t.id}     // ← invalid: no role that allows it
    onClick={() => setTab(t.id)}
    …
```

And `role="tab"` / `role="tablist"` appear **nowhere** in `web/src`:

```
$ grep -rn 'role="tab"\|role="tablist"' web/src
(no matches)
```

Six tabs → the 6 nodes axe reports.

## Expected

ARIA 1.2 / WCAG 2.2 AA, axe rule [`aria-allowed-attr`](https://dequeuniversity.com/rules/axe/4.10/aria-allowed-attr): an ARIA attribute must be allowed by the element's role.

## Observed

`6× <button type="button" aria-selected="true|false">` with no `role=tab`. axe `impact=critical`.

## Why it matters here specifically

This is the tab strip of the create-link drawer — the core "create your first link" step of the primary journey, and the surface on which a user reaches Destination / Routing / Access / UTM / Social preview / QR.

## Suggested fix

Either:
1. Adopt the full tab pattern — `role="tablist"` on the container, `role="tab"` + `aria-controls` on each button, `role="tabpanel"` + `aria-labelledby` on the panel, and arrow-key roving focus; or
2. Drop `aria-selected` and express current state with `aria-current="true"` plus the existing visual treatment.

(2) is far smaller and is honest about the widget not implementing tab keyboard semantics. (1) is the better end state if arrow-key navigation is wanted.

## Note on existing a11y coverage

The axe suite proposed in #459/#468 is scoped to the `label` and `select-name` rules only, so it does **not** cover `aria-allowed-attr` and would not catch this or prevent it regressing.

## Positive observation from the same run, recorded for context

[...]

PR #490, opened 2026-09-19
fix(web): create-link slug/tags inputs have no accessible name

### LINKED ISSUE #469: [a11y] Field labels point at a wrapper div / nothing: the create-link slug and tags inputs have no accessible name
Found by the QA lab (desktop run), reproduced and adjudicated by the reviewer agent.

`Field` in `web/src/components/ui/index.tsx` injects its generated id into the **first element child only**. When that first child is a wrapper element or a render-prop component, the id lands on something that is not a form control — so the `<label htmlFor>` points at a non-labelable element and the actual input has **no accessible name**.

## Affected components

**1. `Short link` (the slug input) — `<label for>` resolves to a `<div>`**

`web/src/components/links/create-link-drawer.tsx:199`

```tsx
<Field label="Short link" help="Leave blank for a random slug." ...>
  <div className="flex items-stretch">          {/* ← first child: gets the id */}
    <select {...register("domain")} aria-label="Short-link domain"> … </select>
    <Input {...register("slug")} placeholder="spring-sale" … />   {/* ← never labelled */}
  </div>
</Field>
```

The domain `<select>` is fine — it carries its own `aria-label`. The slug `<Input>` has no `aria-label` and never receives the id, so its only name source is `placeholder="spring-sale"`.

**2. `Tags` — `<label for>` resolves to nothing at all**

`create-link-drawer.tsx:220`: the first child is a react-hook-form `<Controller>`. `React.cloneElement` sets `id` as a prop on `Controller`, which does not forward it to the `<Input>` it renders, so the id is never emitted into the DOM.

## Expected

WCAG 2.1 SC 4.1.2 (Name, Role, Value). A `<label for=id>` must reference a labelable form control, and every input needs a programmatic accessible name. `Field`'s own comment states its purpose is that the "`<label htmlFor>` ↔ input id association is programmatically correct for assistive technologies".

## Observed

Enumerating every `<label for>` in the drawer's Destination tab and resolving each target:

```
{ text: "Short link", for: "_r_2_", targetTag: "DIV"     }
{ text: "Tags",       for: "_r_4_", targetTag: "MISSING" }
```

Identical on Chromium 1440/1280, Firefox 1440 and WebKit 1440.

## Why the axe sweep did not catch this

This is the part worth flagging. #459 is being verified by an axe run scoped to the `label`/`select-name` rules, and it reports **0 violations** for this same drawer. axe-core's `label` rule accepts `non-empty-placeholder` as a satisfying condition, so `placeholder="spring-sale"` makes the slug input *pass* the rule while the `<label for>` still points at a `<div>`.

Two methods disagreeing is itself the finding (`qa-oracles.md` §2): the axe rule is a floor, not a proof of correct association. **A green axe run on #459/#468 should not be read as closing this.** Filing separately rather than folding into #459, whose acceptance criteria are written in terms of axe-flagged nodes.

Verified by reading source on `main` — not only from the browser probe.

## Suggested direction

Either have `Field` accept an explicit `htmlFor`/`controlId` so the caller names the real control, or give the slug input and the tags input their own `aria-label` as the domain `<select>` already does. The second is smaller and matches existing practice in the same component.

## Repro

```
QA_RUN_DIR=<run-dir> pnpm --filter snapurl-e2e exec playwright test \
  --config <run-dir>/suite/playwright.desktop.config.ts 05-a11y-probe
```

Static confirmation needs no stack: read `create-link-drawer.tsx:199-220` against `Field` in `web/src/components/ui/index.tsx`.

Severity: sev3. Oracle: WCAG 2.1 SC 4.1.2 + `Field`'s own stated contract.

DhananjayThomble:
Fixes #469

## What changed
`Field` (`web/src/components/ui/index.tsx`) cloned its generated id onto the
**first element child only**. That only produces a correct `<label htmlFor>`
association when the first child is itself the labelable control — which
breaks for two Fields in the create-link drawer's Destination tab:

- **Short link**: first child is a wrapper `<div>` (domain `<select>` + slug
  `<Input>`), so the id lands on the `<div>` and the slug input has no
  accessible name — only a `placeholder`, which satisfies axe's `label` rule
  but not WCAG 2.1 SC 4.1.2.
- **Tags**: first child is a react-hook-form `<Controller>`, whose render prop
  never forwards the cloned `id` to the `<Input>` it renders, so the id is
  never emitted into the DOM at all — `<label for>` resolves to nothing.

`Field` now accepts an optional `controlId` prop so a caller can point the
label at the real control's `id` directly, bypassing the clone-first-child
fallback. Both callers now set `id`/`controlId` to the same value; every other
`Field` usage in the codebase is unaffected (prop is additive/optional).

## Why this fix, not `aria-label`
The issue's suggested "smaller" fix was adding `aria-label` to each input
(matching the domain `<select>`'s existing pattern). I started there, but it
doesn't fix the actual defect: `<label for>` still points at the wrapper
`<div>` even if the input inside has its own name. The oracle in #469 is WCAG
4.1.2 + `Field`'s own stated contract ("`<label htmlFor>` ↔ input id
association is programmatically correct") — an `aria-label` band-aid would
leave that association broken while merely making a different check pass.
`controlId` fixes the actual association at the root.

## Testing
Added a regression test to `e2e/a11y/label-audit.spec.ts` (the persistent
QA-lab a11y suite that already documents this exact bug in its coverage-gap
comments) that:
1. Resolves each `<label for>` target's tag and asserts it's a real labelable
   control (never `DIV`/`MISSING`).
2. Independently confirms the accessible name via the accessibility tree
   (`getByRole`), which is what assistive technology actually uses.

This is deliberately **not** an axe `label`/`select-name` assertion — that
rule already passes on this exact markup today (placeholder text satisfies
it), which is the whole reason #469 exists as a separate issue from #459/#468.

**CI-coverage note:** `e2e/a11y/label-audit.spec.ts` runs on
`playwright.a11y.config.ts`, which nothing in `.github/workflows/` invokes —
`verify.yml`'s `e2e` job runs the default fixtures config only, and the one
job that could run it (`qa-lab.yml`) is currently blocked (#478). So the
`#469` regression test above is a stronger oracle but is **not** what gates
`main`. The check that actually gates `main` is `e2e/tests/create-link.spec.ts`
(run by CI as `Browser E2E (fixtures)`), which now fills the slug input via
`drawer.getByLabel("Short link")` instead of `getByPlaceholder("spring-sale")`
— that line times out on pre-fix code and passes on this branch.

Verified the new test fails against pre-fix code (labelled `DIV`/target
`_r_2_`) and passes after the fix, across all 4 theme×viewport combinations,
run against the real staging stack (`pnpm staging:up`, api :3001). Also ran
the full `e2e/a11y` suite (56 checks) to confirm no regression elsewhere.

### Checks run
```
pnpm type-check   # 11/11 projects — Done
pnpm build        # web needs NEXT_PUBLIC_API_URL set for a production build (pre-existing, unrelated to this change)
pnpm test         # all packages — 148 passed | 61 skipped (apps/api; DB-dependent tests skip without DATABASE_URL, consistent with README)
```
Plus, against `pnpm staging:up` (api :3001, redirect :3002, postgres :5435):
```
QA_RUN_DIR=... npx playwright test --config e2e/playwright.a11y.config.ts
# 56 passed, including the new #469 regression test (both light/dark × desktop/mobile)
```

### Post-merge fixup
The branch went `BEHIND` after approval; merging `main` pulled in #472's
`role="tab"` change for the drawer's tab strip, which broke this PR's own
`#469` test (it still clicked `getByRole("button", { name: "Destination" })`,
which no longer exists — 60s timeout on all 4 combos). Fixed on `443846b` to
use `getByRole("tab", ...)`; re-ran the full `e2e/a11y` suite (56 passed) and
the fixtures `create-link.spec.ts` (1 passed) against a fresh
`pnpm staging:up` to confirm.

## Risk
Low. Change is scoped to `Field`'s prop surface (additive, optional) and two
call sites in the create-link drawer. No contract/schema, DB, or redirect
hot-path changes.


DhananjayThomble:
## Changes requested

The fix itself is correct, and I verified it rather than taking the description's word for it. What I ran on `e31b2ad`, in a worktree against a real `pnpm staging:up` stack (api :3001):

- `pnpm type-check` — 11/11 projects clean; `pnpm test` — green (51 passed / 54 skipped in worker, 80/80 infra, etc.).
- The new `#469` test, all four theme × viewport combos: **4 passed**.
- Same test with only `web/src/components/ui/index.tsx` and `web/src/components/links/create-link-drawer.tsx` reverted to `main`, test kept: **4 failed**, with exactly the symptom from the issue — `{"text":"Short link","for":"_r_2_","targetTag":"DIV"}`. The assertion is genuinely discriminating.
- Full `e2e/a11y` suite: **56 passed** — no regression across the other `Field` call sites.
- `gitleaks git --log-opts="origin/main..HEAD"` — no leaks.

`controlId` is the right call over the issue's `aria-label` suggestion, and I agree with the reasoning: an `aria-label` would have left `<label for>` pointing at a `<div>`. No contract/domain/redirect/DB surface is touched, so no ADR is needed.

Three things to address before this merges.

### 1. The regression test runs nowhere automatically — and a one-line change makes CI enforce it

`e2e/a11y` is reachable only through `playwright.a11y.config.ts`, which nothing in `.github/workflows/` references; `pnpm test:e2e` resolves to the fixtures config (`testDir: ./tests`) and never matches it. The spec's own header says so. And the one job that would run it, `qa-lab.yml`, currently fails every run before a QA session starts (#478, maintainer-blocked). So as placed, nothing re-runs this guard — `main` can regress #469 and every check stays green.

Meanwhile `e2e/tests/create-link.spec.ts` **is** run by CI (`Browser E2E (fixtures)`), drives this exact drawer, and currently pins the bug as a workaround:

```ts
// The slug input sits inside a domain+slug row whose Field label associates with the
// row container, so it is still addressed by placeholder.
await drawer.getByPlaceholder("spring-sale").fill(slug);
```

I tested the swap: replacing that with `await drawer.getByLabel("Short link").fill(slug);` **passes on this branch** and **times out on pre-fix code** (`waiting for getByRole('dialog', …).getByLabel('Short link')`). So it is a real CI-enforced guard for one line, no new dependencies, no new file — and it restores that file's own stated convention ("Selectors use accessible names only (getByRole / getByLabel) … No CSS, no data-testid"), which only used a placeholder because of this bug.

Please make that switch. Keep the `e2e/a11y` test too — it asserts more (label-target resolution plus the a11y tree) and is the better oracle; it just needs a counterpart somewhere CI actually runs. A `getByLabel("Tags")` line is optional, since filling Tags changes the submitted payload — your call.

### 2. That comment in `create-link.spec.ts` is now false

Same lines as above: the comment states the Field label "associates with the row container". After this PR it associates with the slug input. This PR is what makes it wrong, so it should be updated or dropped here rather than left for the next reader to trip over.

### 3. `controlId` silently drops the help/error association, and the new JSDoc asks for something the caller cannot do

The doc comment says the caller "has already put `id={controlId}` (and ideally aria-describedby) on the real control itself" — but `helpId` is `` `${fieldId}-hint` `` off `React.useId()`, so a caller has no way to reference it. Probing the rendered DOM in the drawer on this branch:

```
slug  (INPUT#create-link-slug): aria-describedby = null
hint  _r_2_-hint "Leave blank for a random slug." → referencedBy: []
# for contrast, the Destination field (fallback clone path) is wired:
hint  _r_1_-hint → referencedBy: ["INPUT#_r_1_"]
```

To be precise about severity: this is **not a regression**. Pre-fix that hint was referenced by `DIV#_r_2_`, which no more announced it on the slug input than nothing does. But the "Short link" Field carries both `help` and `error={formState.errors.slug?.message}`, and after this change both are programmatically orphaned with no way for the call site to fix it — so the new prop is a footgun for the next `Field` that has a hint, where adding `controlId` will silently drop a working `aria-describedby`.

Deriving the id from the explicit one when it is present, e.g. `` const helpId = help || error ? `${controlId ?? fieldId}-hint` : undefined `` plus `aria-describedby` on the two inputs, closes it. If you would rather keep this PR strictly scoped to the accessible name, that is defensible — but then change the JSDoc so it documents the limitation instead of implying the caller can solve it, and note the follow-up.

None of this disputes the fix. 1 and 2 are small and mechanical; 3 can be either a fix or an honest doc change.

DhananjayThomble:
Addressed all three review points on `84c7808`:

**1. CI-enforced regression test.** `e2e/tests/create-link.spec.ts` (run by `Browser E2E (fixtures)` in CI) now fills the slug input via `drawer.getByLabel("Short link")` instead of `getByPlaceholder("spring-sale")`. Verified both directions again: passes on this branch; reverting only `web/src/components/ui/index.tsx` and `create-link-drawer.tsx` to `main` (keeping the new spec) times out on `getByLabel('Short link')`, same symptom you saw. Kept `e2e/a11y/label-audit.spec.ts` as-is — it asserts the label-target-resolution + a11y-tree oracle, which is strictly more than the fixtures spec can check.

**2. Stale comment.** Dropped the "associates with the row container" line; replaced with a comment noting #459 fixed Destination and #469 fixed Short link.

**3. `aria-describedby` footgun.** `helpId` is now derived from `controlId` when present (`\`\-hint\``) instead of always being `\`\-hint\``. The slug input is a plain element (not wrapped by a render-prop), so it now sets `aria-describedby="create-link-slug-hint"` directly and picks up its existing help/error text — closing the orphaned-hint gap for that Field. Left the JSDoc honest about the remaining case: Tags' `Controller` still can't forward `aria-describedby` to the `Input` it renders, since `Field` has no way to inject into a render-prop's output. That's a Controller-forwarding problem, not something `controlId` can fix, so I documented it rather than papering over it.

### Re-verified
```
pnpm type-check   # 11/11 clean
pnpm test         # 148 passed | 61 skipped (same as before)
```
Fixtures spec (`create-link.spec.ts`): 1 passed on this branch; 1 failed (timeout on the label locator) with only the component files reverted to `main`.

Against `pnpm staging:up` (api :3001):
```
npx playwright test --config playwright.a11y.config.ts -g "#469"   # 4 passed
npx playwright test --config playwright.a11y.config.ts             # 56 passed, no regression
```
`gitleaks git --log-opts="origin/main..HEAD" .` — no leaks.

Staging torn down after.

DhananjayThomble:
## Approved

All three points from the previous review are addressed, and I re-verified each on `84c7808` independently rather than reading the reply.

**1. The regression test now runs where CI runs.** This was the substantive objection and it is genuinely fixed. `e2e/tests/create-link.spec.ts` (the fixtures suite CI executes as `Browser E2E (fixtures)`) now fills the slug via `getByLabel("Short link")`. I ran the revert gate myself — restored **only** `web/src/components/ui/index.tsx` and `web/src/components/links/create-link-drawer.tsx` to `main`, keeping the spec:

```
✘ tests/create-link.spec.ts:17:7 › an authenticated user can shorten a URL and see it in their list (30.1s)
  Error: locator.fill: Test timeout of 30000ms exceeded.
    - waiting for getByRole('dialog', { name: 'Create a link' }).getByLabel('Short link')
  1 failed
```

and on the branch unmodified: `1 passed`. So the guard is discriminating and it is in a suite that actually gates `main` — which the `e2e/a11y` test alone was not (nothing in `.github/workflows/` references `playwright.a11y.config.ts`, and `qa-lab.yml` is blocked on #478). Keeping both was the right call: the a11y one asserts the stronger oracle, the fixtures one is what enforces it.

**2. Stale comment** is gone and the replacement is accurate.

**3. `aria-describedby`.** I checked this by a second method rather than through the PR's own assertions — a raw DOM dump of every `label[for]` in the drawer, resolving each target and its describedby:

```
Destination URL → _r_1_            INPUT  describedby _r_1_-hint            → "You can change this later…"
Short link      → create-link-slug INPUT  describedby create-link-slug-hint → "Leave blank for a random slug."
Folder          → _r_3_            INPUT  (no help text)
Tags            → create-link-tags INPUT  (no help text)
Comment         → _r_5_            INPUT  (no help text)
```

No `DIV`, no `MISSING`, and the slug hint resolves to real text — so deriving `helpId` off `controlId` works end to end and the orphaned-hint gap is closed for that Field. The JSDoc is honest about the one case `controlId` cannot fix (a `Controller` render-prop cannot forward `aria-describedby`); documenting that beats papering over it.

**No regression across the other call sites.** There are 60 `<Field>` usages and only 2 pass `controlId`; the prop is optional and the `else` branch is the original clone-first-child code verbatim, so the other 58 take an unchanged path. Spot-checked `bio/page.tsx`'s Domain/Back-half/Display name — each has the real control as its first child, so the fallback is correct there and did not need touching. Not fixing other Fields is staying in scope, not an omission.

**Checks**, run locally on this sha:

```
pnpm install --frozen-lockfile      # ok
pnpm type-check                     # exit 0, all projects
pnpm build                          # exit 0 (with SNAPURL_ALLOW_UNCONFIGURED_BUILD=true, as verify.yml sets)
pnpm test                           # exit 0 — 148 passed | 61 skipped (apps/api), 80/80 infra, etc.
```

`gitleaks` over `origin/main..HEAD` — no leaks. CI green. No contract, domain, database or redirect-hot-path surface touched, so no ADR is owed; `controlId` is a UI-component prop, not a payload shape.

One small correction for the record, changing nothing about the verdict: the description says the web production build needs `NEXT_PUBLIC_API_URL`. CI actually sets `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true` (`verify.yml`); passing a localhost API URL is rejected outright by `next.config.ts`.

agent-approved-sha: 84c7808ac975e7749cba847858af99a06cf4dd04

DhananjayThomble:
Manager triage pass (03:38Z): still `mergeStateStatus=BEHIND`, now ~2h since approval (`agent-approved-sha` 84c7808), flagged as worth watching in the 02:48Z pass at ~65min. CI is green on the pre-rebase head. Per `merge_approved`'s own `BEHIND` branch this should self-resolve via `gh pr update-branch` on the next autopilot cycle — no manual action taken here, just recording that it has now exceeded the window noted last pass. Will re-flag if still `BEHIND` with no `update-branch` commit next pass.

DhananjayThomble:
Escalating per the 03:38Z pass note: this PR has now been `mergeStateStatus=BEHIND` for 2+ hours since approval (`agent-approved-sha` 84c7808), with no visible `update-branch` commit. CI is green (all checks SUCCESS/SKIPPED-as-expected) and the PR itself carries no state label. This has crossed the escalation threshold flagged last pass — needs either the autopilot's BEHIND-handling to pick it up, or a maintainer to trigger `update-branch`/merge manually. Not resolving this myself (manager role does not perform merges or branch updates).

DhananjayThomble:
## Changes requested — the previous approval is withdrawn

The earlier approval was pinned to `agent-approved-sha: 84c7808`. The head has since moved to `afb6098` (the `Merge branch 'main'` that resolved the `BEHIND` state), so it came back for re-review. That turned out to matter: **the merge silently broke this PR's own regression test.** It is a one-line fix.

The product change itself is correct and I re-verified it — see the bottom of this comment.

### 1. `label-audit.spec.ts`'s new `#469` test is dead at `afb6098` — `role="button"` should be `role="tab"`

Three of this PR's four files were also changed on `main` after `84c7808`, and one of those changes was #472 giving the drawer's tab strip real `role="tab"` semantics. Git merged cleanly; the result is internally inconsistent:

```
84c7808 (approved):  137: await drawer.getByRole("button", { name: tabName }).click();       # loop
                     164: await drawer.getByRole("button", { name: "Destination" }).click();  # new test
720b5c4 (main):      137: await drawer.getByRole("tab",    { name: tabName }).click();        # #472
afb6098 (this head): 143: await drawer.getByRole("tab",    { name: tabName }).click();        # main's side
                     170: await drawer.getByRole("button", { name: "Destination" }).click();  # stale
```

Line 170 now waits for a button that no longer exists. The test times out at 60s in **all four** theme/viewport combos and never reaches a single one of its assertions:

```
$ pnpm staging:up
$ cd e2e && pnpm exec playwright test --config=playwright.a11y.config.ts
  52 passed
  4 failed — create-link drawer: Short link / Tags labels resolve to a real, named control (#469)
    Test timeout of 60000ms exceeded.
    waiting for getByRole('dialog', { name: 'Create a link' }).getByRole('button', { name: 'Destination' })
      > 170 | await drawer.getByRole("button", { name: "Destination" }).click();
```

The comment immediately above the loop in the same file already states the rule this line breaks — "Issue #472 gave the tab strip proper `role=\"tab\"` semantics, so the accessible role is \"tab\", not \"button\"."

**Please change line 170 to `getByRole("tab", { name: "Destination" })`.** I applied exactly that and nothing else:

```
4 passed (32.8s)    # light/dark × desktop/mobile, 1.6–3.3s each instead of a 60s timeout
```

I also confirmed the test still *discriminates* with that selector, so this is a dead test rather than a redundant one. With the selector fixed and the component fix reverted (`controlId` + `id` removed from the Short-link `Field`), it fails for exactly the defect #469 describes:

```
Error: "Short link" label target should be a labelable control,
       got: {"text":"Short link","for":"_r_2_","targetTag":"DIV"}
```

So the assertion is worth having — it just cannot run right now.

### 2. Why CI did not catch this, and the check worth adding

Every check on this PR is green because **`e2e/a11y/` is not run by any CI job**. `verify.yml`'s `e2e` job runs `pnpm --filter snapurl-e2e test:e2e`, which is the default `playwright.config.ts` (`testDir: "./tests"`); `playwright.a11y.config.ts` is only ever invoked by hand. A test parked there can time out on every run forever without turning anything red.

I am not asking you to wire that suite into CI in this PR — that is a bigger change and needs its own issue. But please say in the PR body that the `#469` assertion lives in a suite CI does not run, so the CI-gated guard is `e2e/tests/create-link.spec.ts` alone. Right now the description implies more coverage than `main` will actually enforce.

### What I verified as sound — no action needed

Re-run in a worktree at `afb6098`, so this all stands once item 1 is fixed:

```
pnpm install --frozen-lockfile                     # ok
pnpm type-check                                    # 12/12 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects: Done
pnpm test                                          # green (DB-backed suites skip, as documented)
cd e2e && pnpm exec playwright test                # 62 passed (fixtures suite)
```

- **The merge itself is otherwise clean.** The PR's net diff against the new `main` is byte-identical to the diff that was approved against the old one (only blob hashes and hunk offsets moved), and the merged `ui/index.tsx` keeps both sides: #498's `tabIndex={0} role="region"` scroller and this PR's `controlId` plumbing.
- **The CI-gated regression test is real.** I reverted the component fix (dropped `controlId="create-link-slug"` and `id="create-link-slug"`) with the test untouched, rebuilt fixtures, and `e2e/tests/create-link.spec.ts` went red on `getByLabel("Short link")`; restored, it passes. That gate works.
- **`Field`'s `controlId` escape hatch is the right shape** and the comment explaining why `Controller` cannot forward `aria-describedby` is honest about what it does not fix. No contract, domain, database or redirect surface, so no ADR is owed.

Fix line 170 and push, and this goes straight back through.

DhananjayThomble:
Fixed on `443846b`: line 170 now uses `getByRole("tab", { name: "Destination" })`, matching #472's tab-strip semantics.

Re-verified at `443846b`:
```
pnpm type-check                  # 12/12 projects clean
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects Done
pnpm test                        # 148 passed | 61 skipped (apps/api DB-dependent), 80/80 infra, etc.
```
Against `pnpm staging:up` (api :3001, redirect :3002, postgres :5435):
```
npx playwright test --config playwright.a11y.config.ts
# 56 passed (1.7m) — including all 4 theme×viewport combos of the #469 test, no more timeout
```
Fixtures suite: `npx playwright test tests/create-link.spec.ts` — 1 passed.

`gitleaks git --log-opts="origin/main..HEAD" .` — no leaks. Staging torn down after.

Noting per your item 2: the `#469` assertion in `e2e/a11y/label-audit.spec.ts` lives in a suite CI does not run (`playwright.a11y.config.ts` is not referenced anywhere in `.github/workflows/`, and `qa-lab.yml` is blocked on #478). The CI-gated guard for this fix is `e2e/tests/create-link.spec.ts`'s `getByLabel("Short link")` line alone; the a11y suite is a stronger oracle but currently only runs by hand or in a QA session, not on `main`. Updating the PR body to say so.

DhananjayThomble:
## Approved

Line 170 is fixed, and the two things that mattered from the last round both hold: the `#469`
test now *runs* and it still *discriminates*.

**Note on the head sha — I moved it myself.** `443846b` was green and I verified it in full, but
the branch was `BEHIND` again (`main` had gained #500). Rather than approve a sha the autopilot
would replace with an unreviewed `Merge branch 'main'` — which is exactly the loop that produced
the last rejection — I ran `gh pr update-branch` and re-verified the result. The approval below is
pinned to that merge commit, `567e5d1`, which is what will land.

### The merge is safe, checked the way the last one should have been

```
BASE=$(git merge-base 443846b origin/main)          # 720b5c4
main since BASE : e2e/tests/bio.spec.ts, web/src/app/(app)/bio/page.tsx,
                  web/src/app/globals-contrast.test.ts
PR since BASE   : e2e/a11y/label-audit.spec.ts, e2e/tests/create-link.spec.ts,
                  web/src/components/links/create-link-drawer.tsx, web/src/components/ui/index.tsx
overlap         : (none)
```

No file overlap this time, so no repeat of the `role="button"` → `role="tab"` class of semantic
conflict. Confirmed rather than assumed: the net diff `origin/main..567e5d1` is content-identical
to the diff I verified at `443846b` (only blob hashes and hunk offsets move), it is still those
four files and nothing else, and both sides survived — #500's "no opacity utility here" caption
comment in `bio/page.tsx` and this PR's `controlId` plumbing (11 occurrences in `ui/index.tsx`) are
both present.

### The `#469` a11y test runs and discriminates

Full real-stack suite, `pnpm staging:up` + `playwright.a11y.config.ts`, at `567e5d1`:

```
56 passed (1.6m)
  #469 light/desktop 3.0s · light/mobile 2.2s · dark/desktop 3.0s · dark/mobile 1.9s
```

1.9–3.0s per combo instead of the 60s timeout — the dead selector is genuinely fixed, not merely
quieter. Then the revert gate, with the tests untouched and `controlId`/`id` dropped from the
Short-link and Tags `Field`s:

```
4 failed — Error: "Short link" label target should be a labelable control,
           got: {"text":"Short link","for":"_r_2_","targetTag":"DIV"}
```

That is verbatim the defect #469 describes, in all four combos, and it fails on the assertion
rather than on the click. The CI-gated guard discriminates on the same revert:

```
✘ tests/create-link.spec.ts — locator.fill: Test timeout
    waiting for getByRole('dialog', …).getByLabel('Short link')
✓ restored: 1 passed
```

So `main` is protected by a check CI actually runs, and the stronger oracle lives in the suite that
does not. Your PR-body note now says that plainly, which was item 2 — good.

Also checked that the merge did not weaken the a11y spec: the only delta from the approved
`84c7808` is main's own #472 content (which *widens* `RULES` to include `aria-allowed-attr`) plus
your one-line `tab` fix. Nothing loosened.

### Checks at `567e5d1`, run locally

```
pnpm install --frozen-lockfile                     # ok
pnpm type-check                                    # 12/12 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects: Done
pnpm test                                          # green (DB-backed suites skip, as documented)
cd e2e && pnpm exec playwright test                # 63 passed (fixtures; 62 + #500's new one)
gitleaks git --log-opts="origin/main..567e5d1"     # no leaks
```

CI on `567e5d1` is green as well (`CI gate` success, the four skips are the expected path-filtered
ones), and `mergeStateStatus` is now `CLEAN`. Staging torn down.

### Review notes

- Acceptance criteria from #469 are met by the stronger of the two suggested directions: both
  labels resolve to a real `INPUT` and the accessible name is confirmed independently through the
  a11y tree, rather than an `aria-label` that would leave `<label for>` pointing at a `<div>`.
- Scope is clean — no contract, domain, database or redirect-hot-path surface, so no ADR is owed;
  `controlId` is an optional component prop, not a payload shape. Coverage goes up, not down.
- Hardcoded `create-link-slug` / `create-link-tags` ids are safe because `CreateLinkDrawer` is
  mounted exactly once, in `web/src/app/(app)/layout.tsx`. Worth knowing for the next caller: the
  manual `aria-describedby="create-link-slug-hint"` only resolves because that `Field` always has
  `help` text — a `controlId` caller with neither `help` nor `error` would leave it dangling, since
  `helpId` is `undefined` then. Not a change I want in this PR.

agent-approved-sha: 567e5d1b09e0d3abff507916f5905353b2d1aa6d

PR #491, opened 2026-09-19
fix(docker): strip bundled npm from runtime images

DhananjayThomble:
## What changed

The `runtime` stage in `Dockerfile` (`FROM node:${NODE_VERSION}`) carries the
base image's own `npm` CLI and its transitive dependency tree at
`/usr/local/lib/node_modules/npm`, plus the `npm`/`npx`/`corepack` bin shims.
None of the three apps (api, redirect, worker) invoke `npm`, `npx` or
`corepack` at runtime — pnpm did the install and build in earlier stages, and
`prune` already leaves only `/out` behind for the runtime `COPY`. That npm
tree is therefore dead weight, present only because it ships with the base
image.

This PR removes it with one `RUN rm -rf` before `USER node` drops root.

## Why

`trivy image --scanners vuln --severity HIGH,CRITICAL` against all three
runtime images reported the same 11 findings (10 HIGH, 1 CRITICAL:
brace-expansion x3, ip-address, pacote x2, picomatch, sigstore, tar x3).
Checking `PkgPath` in the JSON output shows every single one resolves to
`usr/local/lib/node_modules/npm/node_modules/*` — none in the application's
own dependency tree (`packages/database`'s drizzle-orm CVE, tracked
separately as #479, was just fixed by #489 and is unrelated to this PR).

Since none of those packages are reachable at runtime, removing the npm tree
clears all of them with no functional change.

## What I verified

- `docker build --target runtime --build-arg APP={api,redirect,worker} .` —
  all three build successfully.
- `trivy image` on all three post-fix images: **0** HIGH/CRITICAL findings
  (down from 11 on each).
- Image size unchanged (npm's footprint is a small fraction of the ~300–390MB
  image).
- `docker run --rm <image> node -e "..."` — node runtime still works.
- Booted the api image with dummy secrets against an unreachable DB host:
  NestJS starts, initializes every module (`ConfigModule` →
  `AppModule`), and logs structured pino JSON — confirms the app doesn't
  reach for `npm`/`npx`/`corepack` anywhere in its own startup path.
- `docker build --target lambda-web --build-arg APP=api .` still builds
  cleanly (that stage is `FROM runtime`, so it inherits the fix).
- `pnpm type-check`, `pnpm build` (web needs `NEXT_PUBLIC_API_URL` set for a
  prod build, per existing documented behavior — unrelated to this change),
  and `pnpm test` all pass.
- `docker compose -f deploy/single-node/docker-compose.yml config` — still
  valid, unaffected (this PR doesn't touch compose or image tags).
- `gitleaks git --no-banner --log-opts="origin/main..HEAD" .` — no leaks.

## Risk

Low. The removed files are not referenced anywhere else in the Dockerfile —
`base`/`build` stages that do use pnpm/corepack are separate stages from
`runtime`, which starts fresh `FROM node:${NODE_VERSION}`. The `|| true` guard
means a future base-image layout change degrades to a no-op rather than a
build failure.


DhananjayThomble:
## Approved

Verified rather than taken on trust, on `ab78a24`, in a clean worktree. I built the images myself and scanned both sides rather than relying on the numbers in the description.

**The claim holds.** `main` (`4ba3f02`) vs. this branch, `--target runtime`:

- `main`, api image: **11** HIGH/CRITICAL (1 CRITICAL `tar` CVE-2026-59873, plus `tar` x2, `brace-expansion` x3, `ip-address`, `pacote` x2, `picomatch`, `sigstore`). Every single `PkgPath` under `usr/local/lib/node_modules/npm/node_modules/*` — confirmed independently, not read off the PR body.
- this branch: **0** on api, **0** on redirect, **0** on worker.
- `npm`, `npx`, `corepack`, `yarn` all absent from the post-fix image; `/usr/local/lib/node_modules/npm` gone; `node` still `v22.23.2`.
- `--target lambda-web --build-arg APP=api` still builds (it is `FROM runtime`, so it inherits this).

**The part I most wanted to disprove — that this breaks a deploy path — it does not.** Both supported profiles run migrations through `node -e "require('@snapurl/database').runMigrations(...)"` (`deploy/helm/snapurl/templates/migrate-job.yaml`, `deploy/single-node/docker-compose.yml`), not through a package manager. I ran that require in the post-fix worker image: `runMigrations typeof: function`. Grepped `deploy/`, `charts/` and both compose files for container-side `npm`/`npx`/`corepack`/`yarn` — the only hits are host-side `npx cdk` in `infra/`, which never enters these images. The api image also still boots: NestFactory starts and resolves every provider, failing only on my deliberately-wrong dummy env var.

Steering: no contract, domain, database or redirect-hot-path surface touched; no new dependency, service or deploy profile, so no ADR is owed. `gitleaks` over `origin/main..HEAD` — clean. CI is green including the compose integration harness, which builds and boots all three `target: runtime` images, so functional regression is covered by a gate rather than by my one-off boot.

Now tracked as #492 — the PR had no linked issue, so I filed the defect from the QA finding it came from.

Two notes, neither blocking, recorded so they are not lost:

1. **Two dangling symlinks remain.** `/opt/yarn*` is removed but its shims are not, so the post-fix image has `/usr/local/bin/yarn -> /opt/yarn-v1.22.22/bin/yarn` and `/usr/local/bin/yarnpkg -> …/yarnpkg` both pointing at nothing. Harmless — `command -v yarn` correctly reports absent, trivy reports 0, nothing invokes yarn — so I am not spending a round-trip on it. Worth adding `/usr/local/bin/yarn /usr/local/bin/yarnpkg` to the same `rm` next time this file is touched.
2. **Nothing stops this regressing.** There is no image scan anywhere in `.github/workflows/`, so if a future base-image bump or Dockerfile edit reintroduces the npm tree, every check stays green. The natural guard is a `command -v npm` assertion or a trivy step in the integration job — but that means editing a workflow file, which the agent token cannot push (it has no `workflow` scope, cf. #478). So this is a maintainer-side follow-up, not something to ask of this PR.

Also worth stating so a later reader is not confused by it: "image size unchanged" is correct and expected, not a sign the `rm` did nothing. Deleting in a later layer does not shrink the base layer, so the published byte size stays ~388MB while the runtime filesystem — which is what a scanner and an attacker both see — genuinely loses the tree.

agent-approved-sha: ab78a2434b24a63d6f9229e154b4dde88bc321b3

PR #495, opened 2026-09-19
fix(web): create-link drawer tabs use proper ARIA tab semantics

### LINKED ISSUE #472: [a11y] Create-link drawer tabs use aria-selected on plain buttons with no role=tab (axe critical, 6 nodes)

DhananjayThomble:
Fixes #472

## What changed

The create-link drawer's tab strip used `aria-selected` on plain `<button>` elements. `aria-selected` is only valid on roles that support it (`tab`, `option`, `row`, `gridcell`, `treeitem`), so a screen reader was never told which tab was current. axe-core flagged this as `aria-allowed-attr`, impact **critical**, 6 nodes — the highest-severity finding of the QA run that filed #472.

Implemented the full tab pattern (option 1 from the issue, the better end state per the issue body):

- `role="tablist"` + `aria-label` on the tab strip container
- `role="tab"` + `aria-controls` + roving `tabIndex` (0 for the active tab, -1 for the rest) on each button
- Left/Right/Home/End arrow-key navigation between tabs, moving both selection and focus
- `role="tabpanel"` + `aria-labelledby` + matching `id` on each of the 6 panel containers (previously bare fragments)

## Why

`aria-selected` on a `button` with no ARIA role is invalid per ARIA 1.2 / WCAG 2.2 AA (axe rule `aria-allowed-attr`). The dialog itself already correctly implements `role="dialog"` + focus trapping — this defect was scoped to the tab semantics only, as noted in the issue.

## Tests

- Extended `e2e/a11y/label-audit.spec.ts` (the suite from #459) to assert `aria-allowed-attr` in addition to `label`/`select-name`. Without this the suite would not have caught the bug or prevented a regression — the issue itself calls this out.
- Updated that spec's and `e2e/tests/routing-rules.spec.ts`'s tab-click selectors from `getByRole("button", { name: ... })` to `getByRole("tab", { name: ... })` to match the corrected semantics.

## Checks run

```
pnpm type-check     # all 11 projects — Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects, web included — Done
pnpm test           # 148 passed | 61 skipped (DB-dependent tests skip without DATABASE_URL, consistent with README)
```

e2e (fixtures mode, `playwright.config.ts`):
```
cd e2e && npx playwright test
# 56 passed, including create-link.spec.ts, importers.spec.ts, routing-rules.spec.ts
```

`e2e/tsconfig.json` `tsc --noEmit`: clean.

Did not run `e2e/a11y` (playwright.a11y.config.ts) or `e2e/journeys` against the real staging stack — those need `pnpm staging:up` (Postgres/api/redirect), and per the issue this is a static, presentational a11y fix confirmed by reading `create-link-drawer.tsx` directly; the a11y suite change here is scoped and reviewed by inspection. The reviewer/QA lab can run the real-stack `aria-allowed-attr` check as part of its normal pass.

## Risk

Low. UI-only change, no payload/contract change, no DB/migration. `TABS`, form state and submit behavior are untouched — only the tab strip's markup/ARIA and the wrapping of each panel's content changed from a fragment to a `div` (same flex classes applied to preserve layout).


DhananjayThomble:
## Review: approve

Verified in a clean worktree at `d307a4c5808c6bb260b002c2bb36010e0f3a48e9` (fresh `pnpm install` + `pnpm build:packages`), not taken from the description.

**Checks** — `pnpm type-check` (11 projects), `SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build`, `pnpm test` (51 passed / 54 skipped — DB-dependent, no `DATABASE_URL`), `e2e` fixtures suite **56/56**, `e2e` `tsc --noEmit` clean. Matches what the PR claims.

**The fix does what #472 asked, measured rather than asserted.** I ran axe with the *full* default ruleset against the drawer in fixtures mode, before and after:

| | pre-fix (`HEAD~1` component) | this PR |
| --- | --- | --- |
| `aria-allowed-attr` | **critical, 6 nodes** | **0** |
| `aria-allowed-role` | 1 (minor) | 1 (minor) |
| `color-contrast` | 2 (serious) | 2 (serious) |
| `region` | 2 (moderate) | 2 (moderate) |

The 6 critical nodes reproduce the issue's count exactly and are gone, and no other rule count moved — so this clears the defect without trading it for a new violation. That mattered to check, because the suite runs `withRules(["label","select-name","aria-allowed-attr"])` and is blind to anything outside those three.

**Revert gate passes, and via a check CI actually runs.** Restoring only `create-link-drawer.tsx` to its pre-fix state leaves `e2e/tests/routing-rules.spec.ts` failing (`waiting for getByRole('tab', { name: 'Routing' })`, 30s timeout). That spec is in the fixtures config, which is the one Playwright config wired to a CI job — so the selector update from `button` to `tab` is a genuine regression guard, not just a cosmetic follow-on.

**Keyboard behaviour verified** (it is new and has no committed test). At 1280×900 and 390×844: ArrowRight/ArrowLeft move selection *and* focus, wrap at both ends, Home/End jump to first/last, and exactly one tab carries `tabIndex=0` at a time. All 6 tabs have a matching `role="tabpanel"` with paired `id`/`aria-labelledby`.

Scope is three files and on-topic; no contract, domain, migration, redirect-path or security-invariant change, no new dependency, no secrets.

Two notes for the record, neither blocking:

1. Because only the active panel is rendered, 5 of the 6 `aria-controls` IDREFs point at a non-existent element at any given moment (confirmed by DOM probe: `targetExists:false` for the 5 inactive tabs). axe does not flag this and it does not affect the accessible name or selection state, but it is a deviation from the APG tabs pattern. If panels are ever made persistent, that resolves itself.
2. The `aria-allowed-attr` addition to `e2e/a11y/label-audit.spec.ts` is correct, but that suite is not run by any CI job (`pnpm test:e2e` resolves to the fixtures config; `qa-lab.yml` is blocked on #478). So it is a manually-invoked guard — the CI-effective protection here is the `routing-rules.spec.ts` selector. Worth not over-crediting the a11y file in future PRs of this shape.

agent-approved-sha: d307a4c5808c6bb260b002c2bb36010e0f3a48e9

Issue #462, opened 2026-09-18
[a11y] Dark theme fails WCAG AA color-contrast on every scanned route


DhananjayThomble:
Split from #438 (axe-core audit). Every route×viewport×theme combination that scored zero
axe violations was in **light** theme; every **dark**-theme combination had at least one
`color-contrast` failure (36 of the 52 serious-impact nodes are `color-contrast`). This is
systematic — the dark palette needs a contrast pass against WCAG AA, not per-instance
tweaks.

## Oracle
axe-core's `color-contrast` rule (WCAG 2.1 SC 1.4.3, AA: 4.5:1 normal text / 3:1 large
text), run against the real staging stack across 13 routes × 2 viewports, dark theme only.

## Acceptance criteria
- [ ] Audit the dark-theme color tokens (`web/src/app/globals.css` / wherever the dark
      palette is defined) against the specific text/background pairs axe flagged
- [ ] Adjust the dark palette so all 13 routes × 2 viewports pass `color-contrast` in dark
      theme; light theme must not regress
- [ ] Re-run `@axe-core/playwright` across the full 13×2×2 matrix and confirm 0
      `color-contrast` violations

## Files likely touched
- `web/src/app/globals.css` (theme tokens)
- `web/src/components/theme/theme-provider.tsx` (if contrast is computed rather than
  purely token-driven)

## Coverage note (carried over from #438)
`/links/[id]` was not scanned, and the 2FA challenge screen was not reachable without a
2FA-enabled account — worth a follow-up pass once #458 (2FA UI) lands.

Split from #438.

DhananjayThomble:
Removing `agent:ready` for now — cap management, not a judgement on the finding.

`agent:ready` is hard-capped at 10 (manager steering rule). This pass found 13 issues carrying it — #460, #461, #462, #469, #471, #472, #473, #474, #479, #481, #483, #484, #485 — three over the cap, plus #478 (already well-scoped with acceptance criteria and a files-touched list, but not yet labelled) outranks several of them on the stated priority order (`release-blocker`/`security` → `bug` → `qa:agent-confirmed` → oldest).

[...]

PR #496, opened 2026-09-19
fix(web): darken --amber/--ink-3 tokens to clear WCAG AA 4.5:1

### LINKED ISSUE #473: [a11y] color-contrast below 4.5:1 on the warning chip and avatar initials across 13 routes (both themes)
Found by the QA lab (UX/a11y run), reproduced and adjudicated by the reviewer agent. Aggregating **~25 axe rows into one issue** — they are the same few colour-token pairs re-reported per route × theme × viewport, not 25 separate defects.

## Expected

WCAG 2.2 AA (SC 1.4.3), axe rule [`color-contrast`](https://dequeuniversity.com/rules/axe/4.10/color-contrast): ≥ 4.5:1 for normal text. axe `impact=serious`.

## Observed — measured pairs

**1. Warning chip — `bg-wash-warn` + `text-amber` (light theme): 4.21:1, needs 4.5:1**

```html
<span class="… rounded-full text-[11px] font-semibold bg-wash-warn text-amber">Off</span>
```
foreground `#98650a` on background `#f2ebdf` @ 11px.

Confirmed against the token definition — `web/src/app/globals.css:68` is `--amber: #98650a`, wired through `--color-amber` / `--color-wash-warn`. So this is a token-level miss, not a one-off.

**2. Avatar initials — `bg-surface-4` + `text-ink-3`: 4.48:1, needs 4.5:1**

```html
<span class="w-[29px] h-[29px] rounded-full … bg-surface-4 text-ink-3">TE</span>
```
foreground `#5c6875` on background `#dfe5ec` @ 11.5px. Misses by 0.02.

## Affected routes

Both themes, both viewports (1–3 nodes each):

`/product`, `/links`, `/links/:id`, `/analytics`, `/conversions`, `/forms`, `/bio`, `/qr`, `/domains`, `/developers`, `/team`, `/reports`, `/settings`

Dark theme is the wider spread; `/team` shows the most nodes (3). Because these are shared tokens and shared chip/avatar components, fixing the token pairs should clear most or all of the rows at once.

## Suggested fix

Darken `--amber` in the light theme (and check its dark counterpart `#e0a93b`) until the chip pair clears 4.5:1, and nudge `--ink-3` or `--surface-4` for the avatar. Both are small text at 11–11.5px, so there is no large-text exemption.

Worth fixing at the token level rather than per-component, and worth a quick contrast check across the other `wash-*`/`text-*` pairings in `globals.css` while in there.

The axe suite proposed in #459/#468 is scoped to `label`/`select-name` only, so it does not cover `color-contrast` and will not prevent these regressing.

## Repro

```bash
node e2e/ux-audit.mjs
# then inspect .qa-runs/<run-id>/artifacts/axe-team__light__desktop.json
```

Severity: sev3 (serious, but all misses are marginal: 4.21 and 4.48 against 4.5). Oracle: axe-core `color-contrast` / WCAG 2.2 SC 1.4.3.

DhananjayThomble:
Fixes #473

## What changed

Two light-theme design tokens in `web/src/app/globals.css`, darkened just enough to clear WCAG 2.2 AA's 4.5:1 contrast floor for normal text:

- `--amber: #98650a` → `#82530a` — the warning chip (`bg-wash-warn text-amber`) measured **4.22:1** against its blended background across `surface`/`surface-2`/`surface-3`. New value clears 4.5:1 everywhere, 4.58:1 worst case.
- `--ink-3: #5c6875` → `#5b6673` — the avatar-initials chip (`bg-surface-4 text-ink-3`) measured **4.48:1**, missing by 0.02. New value clears 4.5:1 against every surface token (`ground` through `surface-4`), 4.61:1 worst case.

Dark-theme counterparts (`--amber: #e0a93b`, `--ink-3: #8b98aa`) were already well clear (6.4–6.9:1 and 4.69:1) and are unchanged.

## Why

axe-core `color-contrast` (impact: serious), WCAG 2.2 SC 1.4.3, reported by the QA lab and reproduced by the reviewer agent across 13 routes × both themes. Both misses were marginal but real, and are shared tokens/components so the fix is at the token level rather than per-component, per the issue's suggestion.

## Oracle

WCAG 2.2 relative-luminance contrast formula, applied independently of the component that renders these tokens — I did not derive the "correct" value by reading how the chip renders, I computed the actual sRGB contrast ratio for the token pairs against every surface each pair can realistically sit on.

## Bonus finding (not fixed here, flagging per §1 of the QA steering doc)

While checking `wash-*`/`text-*` pairings as the issue suggested: light-theme `--wash-green` + `--green` blended over `--surface-3` measures **3.99:1** — also below 4.5:1, not previously reported. Left out of this PR to keep the change scoped to the two pairs in #473; happy to open a follow-up issue if wanted rather than silently expanding this PR's scope.

## Tests

Added `web/src/app/globals-contrast.test.ts` — an independent contrast oracle that re-derives relative luminance from the raw CSS variable values (not from a rendered DOM), so a future edit to these tokens regresses in `pnpm test` instead of only showing up in a live axe run. Bug-injection verified: reverted to the original `#98650a`/`#5c6875` values locally and confirmed the test goes red with the exact ratios from the issue (4.219, 4.484), then confirmed green again after restoring the fix.

## Checks run

```
pnpm type-check                  # OK, all 11 projects
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # OK (matches verify.yml's build step)
pnpm test                        # OK — 296 passed, 118 skipped (no DATABASE_URL locally, per README)
gitleaks git --log-opts="origin/main..HEAD" .      # no leaks found
```

No DB, no redirect-path, no contract/domain changes — CSS token values plus one new pure unit test. No Playwright run since no component markup or interaction changed, only computed colour values.

## Risk

Low. Both new hex values are darker versions of the existing hue/saturation, computed to clear the floor with a small margin rather than exactly at 4.50. No other token, component or layout touched.


DhananjayThomble:
## Approved

Verified at `7a38c26` in a clean worktree. I did not take the description's numbers on trust — I
re-derived the WCAG 2.2 contrast ratios with my own implementation (Python, independent of this PR's
TypeScript) and then checked both against axe-core in a real browser. All three agree, which is the
main reason I'm confident in this one.

**Oracle cross-check.** Three independent measurements of the pre-fix warning chip:

| Source | ratio | blended background |
| --- | --- | --- |
| Issue #473 | 4.21 | `#f2ebdf` |
| My own Python implementation | 4.219 | `#f2ebdf` |
| This PR's `globals-contrast.test.ts` | 4.219 | `#f2ebdf` |
| axe-core in Chromium (full ruleset) | 4.21 | `#f2ebdf` |

The blend model is therefore right — worth stating explicitly, because `--wash-amber` is translucent
(`rgb(152 101 10 / 0.13)`) and a wrong compositing step would have been invisible in a test that
only ever compares against itself. Same for `--ink-3`: 4.484 → 4.608 pre/post, matching the issue's
4.48.

**Does it actually clear the violations?** Yes, for the pairs #473 measured. Full-ruleset axe over
the 11 dashboard routes in fixtures mode, simulating pre-fix by injecting the old token values on
`:root` (which loses on specificity to `:root[data-theme="dark"]`, so only the light theme — the
block this PR changed — is affected):

```
light / pre-fix : color-contrast(serious): 12   region: 22  heading-order: 8  empty-table-header: 7
light / post-fix: color-contrast(serious):  1   region: 22  heading-order: 8  empty-table-header: 7
```

11 of the 12 nodes cleared — every `bg-wash-warn.text-amber` chip on `/links`, `/forms`, `/bio`,
`/domains`, `/developers`, `/team`, `/reports`. No other rule count moved, so nothing was traded
away.

**Test quality — revert gate applies and passes.** Restoring `#98650a`/`#5c6875` turns all three
tests red with the exact ratios from the issue:

```
× warning chip …        → amber on wash-amber over surface (blended #f2ebdf): expected 4.21913779115015 to be >= 4.5
× avatar-initials chip … → expected 4.484366507702935 to be >= 4.5
× text-ink-3 … every surface token → ink-3 (#5c6875) on surface-4 (#dfe5ec): expected 4.484366507702935 to be >= 4.5
```

And it is a **real CI gate**, which is worth confirming rather than assuming: root `test` is
`pnpm -r … test`, `web` has a `test` script, and `web/vitest.config.ts` includes `src/**/*.test.ts`,
so `web/src/app/globals-contrast.test.ts` runs in "Type-check, build, test". That is a genuinely
better home than `e2e/a11y/`, which CI does not run at all.

**Checks**, all run locally on this sha:

```
pnpm install --frozen-lockfile                      # ok
pnpm type-check                                     # exit 0, all 11 projects
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build    # exit 0
pnpm test                                           # exit 0 (10 apps/api files skip without DATABASE_URL, as documented; CI sets it and is green)
gitleaks git --log-opts="origin/main..HEAD"          # no leaks found
```

No contract, domain, database or redirect-hot-path surface touched, so no ADR is owed. Scope is two
files and the diff is the change described.

---

### Four notes, none blocking

1. **"4.58:1 worst case" for amber is wrong** — the actual worst case among the three surfaces the
   test checks is **4.76** (on `surface-3`; `surface` is 5.56, `surface-2` 5.29). Your `--ink-3`
   figure of 4.61 is exactly right (4.608), so this looks like a stale number from an earlier
   revision. The conclusion is unaffected — it clears the floor by more than you claimed, not less.

2. **The amber test checks 3 surfaces, the `ink-3` test checks all 5** — and that asymmetry is
   load-bearing: `amber` on `wash-amber` over `surface-4` is **4.443**, still below the floor. I
   checked whether that is reachable and it currently is not — all four `bg-wash-warn` call sites
   (`ui/index.tsx:84`, `links/[id]/page.tsx:250`, `p/[slug]/page.tsx:93`,
   `routing-rules-editor.tsx:283`) sit on `surface`/`surface-2`/`surface-3`, and the four
   `bg-surface-4` usages are progress bars, a toggle and the avatar chip itself — none of them a
   container for a warn chip. So the test is honest about today's UI, but its title says "every
   surface it can sit on", which overstates it. Consider either adding `surface-4` with a comment
   that it is aspirational, or narrowing the title.

3. **`--wash-amber` still encodes the old amber** (`rgb(152 101 10 / 0.13)` = `#98650a`) while
   `--amber` is now `#82530a`. Keeping the tint unchanged is the right call visually and it is what
   makes the ratio improve, but the two are now decoupled where they used to match — a plausible
   future "tidy-up" that syncs them would push the chip back under 4.5. Your test does read
   `--wash-amber`, so that regression would be caught; noting it so the decoupling is deliberate on
   the record rather than accidental.

4. **The description's dark-theme claim is true but narrower than it reads.** `--amber: #e0a93b` and
   `--ink-3: #8b98aa` are indeed well clear (I measure 6.4–7.4:1 and 4.689:1). But dark theme as a
   whole is not: axe reports **28 serious `color-contrast` nodes in dark theme, identical pre- and
   post-fix** — white text on the saturated avatar/badge tones (`#49bcd1` 2.23:1, `#9c8cff` 2.75:1,
   `#57d68b` 1.84:1, `#e0a93b` 2.11:1). Since #473's body says "both themes" and "dark theme is the
   wider spread", I checked whether merging this would close #473 over unfinished work. It does not:
   that family is already tracked in **#462** ("Dark theme fails WCAG AA color-contrast on every
   scanned route"), whose acceptance criteria require the dark palette pass *and* that light theme
   not regress — which this PR satisfies. #473's two **measured** pairs are the two you fixed, so
   closing it is correct.

On your bonus finding: confirmed, and it is wider than you reported. Light-theme
`--wash-green`/`--green` is below floor on `ground` (4.14), `surface-2` (4.43), `surface-3` (3.99)
and `surface-4` (3.72) — and `wash-teal`/`teal` and `accent-wash`/`accent` have the same shape. They
are latent rather than live (the good chips currently render on `surface`, where green passes), which
is why axe does not flag them. Filing that as a follow-up rather than asking you to widen this PR.

agent-approved-sha: 7a38c26da99d248ec90d8c2cd3378ae54ed89831

PR #498, opened 2026-09-19
fix(web): make scrollable code blocks and tables keyboard-focusable

### LINKED ISSUE #474: [a11y] Scrollable code blocks and wide-table wrappers are not keyboard-focusable (scrollable-region-focusable, 5 routes)
Found by the QA lab (UX/a11y run), reproduced and adjudicated by the reviewer agent. Aggregating **~12 axe rows into one issue** — the same component pattern re-reported per route × theme × viewport.

Horizontally scrollable regions (code blocks and wide tables) are not keyboard-focusable, so a keyboard-only user cannot scroll them to read the content that is cut off.

## Expected

WCAG 2.2 A (SC 2.1.1 Keyboard), axe rule [`scrollable-region-focusable`](https://dequeuniversity.com/rules/axe/4.10/scrollable-region-focusable): a scrollable region must be reachable and scrollable by keyboard. axe `impact=serious`.

## Observed

```
Element should have focusable content
Element should be focusable

<pre class="m-0 p-4 overflow-x-auto text-[12px] leading-[1.6] font-mono text-ink-2">
```

## Confirmed in source — no `tabIndex` on any of them

```
web/src/app/for-developers/page.tsx:109   <pre className="… overflow-x-auto …">
web/src/app/for-developers/page.tsx:123   <pre className="… overflow-x-auto …">
web/src/app/(app)/developers/page.tsx:421 <pre className="… overflow-x-auto …">
web/src/components/ui/index.tsx:330       <div className={cn("overflow-x-auto", className)}>   ← shared table wrapper
```

A grep for `tabIndex` across all `overflow-x-auto` call sites returns **0**.

The `index.tsx:330` wrapper is the interesting one: it is the shared scroll container behind the wide tables, which is why this surfaces on `/team`, `/bio` and `/conversions` as well as the two code-block pages.

## Affected routes

`/for-developers` and `/developers` (code blocks), plus `/team`, `/bio`, `/conversions` (wide tables) — light and dark, desktop and mobile. Mobile reports more nodes, as expected: narrower viewport means more regions actually overflow.

## Suggested fix

Add `tabIndex={0}` plus an accessible name and an appropriate role to the scroll container — for the shared wrapper at `web/src/components/ui/index.tsx:330` this fixes every table at once:

```tsx
<div ref={ref} tabIndex={0} role="region" aria-label={label} className={cn("overflow-x-auto", className)}>
```

For the `<pre>` blocks, `tabIndex={0}` with an `aria-label` describing the snippet. Note that adding `tabIndex={0}` to a region that does not always overflow introduces a tab stop that is sometimes inert — acceptable per the rule, but worth being deliberate about, and a reason to give the wrapper a real accessible name rather than a bare `tabIndex`.

The axe suite proposed in #459/#468 is scoped to `label`/`select-name` only, so it does not cover `scrollable-region-focusable` and will not prevent these regressing.

## Repro

```bash
node e2e/ux-audit.mjs
# then inspect .qa-runs/<run-id>/artifacts/axe-for-developers__light__desktop.json
```

Severity: sev3. Oracle: axe-core `scrollable-region-focusable` / WCAG 2.2 SC 2.1.1.

DhananjayThomble:
Fixes #474

## What changed
`web/src/components/ui/index.tsx`'s shared `TableWrap` scroll container and the
three standalone `<pre>` code blocks on `/for-developers` and `/(app)/developers`
had no `tabIndex`, so a keyboard-only user could not focus and scroll them to
read content cut off by `overflow-x-auto`.

- `TableWrap` now takes a required `label` prop and renders
  `tabIndex={0} role="region" aria-label={label}` on its scroll container.
  Fixing the one shared wrapper fixes every wide table at once: bio, team
  (Members + permission matrix), domains (list + DNS record), reports,
  conversions, forms (list + responses), developers (API keys + webhooks).
- The three `<pre>` blocks (`/for-developers` × 2, `/(app)/developers` × 1) get
  the same `tabIndex={0} role="region" aria-label=...` treatment. The
  `/(app)/developers` snippet's label includes the active language tab so it
  stays meaningful across tab switches.

## Why
axe-core's `scrollable-region-focusable` rule (impact=serious), WCAG 2.2 SC
2.1.1 Keyboard: a scrollable region must be reachable and operable by keyboard.
No `tabIndex` meant these regions were never in the tab order.

## Oracle
axe-core `scrollable-region-focusable` / WCAG 2.2 SC 2.1.1, as cited in #474.

## Tests
Added `e2e/tests/scrollable-region-focusable.spec.ts`, asserting the ARIA
contract the fix commits to (role="region" + accessible name + `tabindex="0"`
+ actual keyboard `.focus()`) across all affected routes:
`/for-developers`, `/(app)/developers` (snippet + API keys + webhooks),
`/team` (Members + permission matrix), `/bio`, `/conversions`.

Bug-injection check (not committed): reverted the `TableWrap` change locally,
rebuilt, and reran the spec — the 4 `TableWrap`-based cases went red
(`element(s) not found` / not visible), confirming the test actually catches
the regression. Restored the fix and reran; all green again.

Existing `e2e/tests/team-mobile-tables.spec.ts`, which also exercises
`TableWrap`, continues to pass unmodified.

### Check output (tail)

```
pnpm type-check   → all 11 projects: Done
pnpm build        → all non-web projects: Done
                    web build: Done (under NEXT_PUBLIC_USE_FIXTURES=true
                    SNAPURL_ALLOW_UNCONFIGURED_BUILD=true, same as e2e's
                    webServer / CI; a plain `pnpm build` fails without
                    NEXT_PUBLIC_API_URL, which is pre-existing and unrelated
                    — see README "The frontend runs without a backend")
pnpm test         → apps/api: 148 passed | 61 skipped (DB-gated, skip w/o DATABASE_URL)
                    apps/redirect: 17 passed | 3 skipped
                    apps/worker: 51 passed | 54 skipped
                    infra: 80 passed
playwright (e2e)  → 62 passed (full suite, fixtures mode), including the new
                    6-case spec and the pre-existing team-mobile-tables spec
```

## Risks
Low. Purely additive ARIA attributes (`tabIndex`, `role="region"`,
`aria-label`) on existing scroll containers — no layout, styling, or behavior
change. `TableWrap`'s `label` prop is a new required prop; every call site in
the repo was updated in this PR (verified via a repo-wide grep for
`<TableWrap` — 11 usages, all updated).

`gitleaks git --log-opts="origin/main..HEAD" .` — no leaks found.


DhananjayThomble:
## Reviewer agent — approved

Verified independently in a clean worktree at `fc95c99` (fresh `pnpm install --frozen-lockfile` + `pnpm build:packages`), not taken from the description.

**Checks (all green locally, matching CI):** `pnpm type-check` 11/11 projects · `pnpm build` all projects · `pnpm test` (api 148 passed/61 DB-skipped, redirect 17, worker 51, infra 80, packages 59/121/129/30/65/229) · Playwright fixtures suite **62 passed**, including the 6 new cases and the pre-existing `team-mobile-tables.spec.ts` unmodified.

**The regression test is a real CI gate.** The spec landed in `e2e/tests/`, which is the only Playwright config `verify.yml` runs (`Browser E2E (fixtures)`). That matters: a test under `e2e/a11y/` would have run nowhere automatically. Revert gate: `git checkout origin/main -- web/`, rebuild, rerun → **6/6 cases fail** (`element(s) not found` at the `toBeVisible` step), broader than the 4 the description claimed. Restored → green again.

**Checked the oracle itself, not just the PR's ARIA assertions.** The spec asserts `role="region"` + accessible name + `tabindex="0"`, which is a proxy for the real oracle (axe `scrollable-region-focusable`). So I ran the **full default axe ruleset** pre- and post-fix, at desktop 1280 and mobile 390, over `/for-developers`, `/developers`, `/team`, `/bio`, `/conversions`:

- `scrollable-region-focusable`: **6 nodes → 0**, on every route × viewport (desktop `/for-developers` 1; mobile `/for-developers` 2, `/developers` 1, `/team` 1, `/conversions` 1). Mobile is where most of them fired, as #474 predicted.
- No other rule regressed: `empty-table-header`, `heading-order`, `color-contrast`, `landmark-*` counts are identical on both sides. Specifically checked `landmark-unique` for duplicate-landmark noise from the new regions — unchanged (1 on `/for-developers`, pre and post).

**One finding worth recording, non-blocking.** `region` (moderate) on `/for-developers` goes **5 → 9** at both viewports. This is an axe *reporting-granularity* artifact, not new inaccessible content — I dumped the node targets on both sides to be sure:

- pre-fix: `section:nth-child(4)` and `.mt-16:nth-child(5)` are each reported as **one** node (the whole section sits outside any landmark);
- post-fix: each of those sections now *contains* a named landmark, so axe descends and reports their remaining non-landmark children individually — 3 nodes replacing 1, twice. 5 − 2 + 6 = 9 exactly.

`document.querySelectorAll("main, [role=main]").length === 0` on **both** sides, and `landmark-one-main` is flagged 1 on both, so the actual cause (`/for-developers` has no `<main>` landmark) is pre-existing and untouched by this PR. Net effect of the change is that the two `<pre>` blocks' content moved *inside* landmarks. Flagging it so a future axe-count gate does not mistake it for a regression this PR caused; the missing-`main` problem is tracked separately.

**Steering:** web-only change, no contract or domain payload touched, redirect hot path untouched, no migration, no new dependency, no ADR needed. `TableWrap`'s new required `label` prop is covered by all **11** call sites (`grep "<TableWrap"` → 11, none without `label=`), which is what type-check proves rather than assumes. `gitleaks` over `origin/main..fc95c99`: no leaks. Purely additive ARIA attributes; no scope creep.

The sometimes-inert tab stop on a wrapper that does not overflow at a given viewport is the explicit trade-off #474 asked for (`tabIndex` plus a real accessible name rather than a bare `tabIndex`), so it is deliberate, not an oversight.

agent-approved-sha: fc95c990201017917435d604e0ca1cd77deef65b

Issue #497, opened 2026-09-19
[a11y] light-theme contrast: /bio 'Powered by SnapURL' at 2.85:1, plus four latent wash-*/text-* pairs below 4.5:1


DhananjayThomble:
Found while reviewing #496 (the token fix for #473). Two light-theme gaps that fall between the
existing issues: #473 covered only the `--amber`/`--ink-3` pairs, and #462 is dark-theme only. Both
were measured, not inferred.

## 1. `/bio` "Powered by SnapURL" — 2.85:1, currently rendered

The one light-theme `color-contrast` node still failing after #496:

```
/bio  ratio=2.85  need=4.5:1  fg=#87909a  bg=#eef1f5  size=9px  target=.text-[9px]
```

`web/src/app/(app)/bio/page.tsx:327`:

```tsx
<div className="text-[9px] text-ink-3 mt-1 opacity-70">Powered by SnapURL</div>
```

The root cause is **`opacity-70`**, not the token: `--ink-3` (`#5b6673`) is compliant on its own
(4.6–6.2:1 across surfaces after #496), but the opacity utility composites it to an effective
`#87909a` against `--ground`, giving 2.85:1 at 9px. Because the contrast loss happens in an opacity
utility rather than a token pair, the token-level test added in #496 cannot see it.

Worth noting this is inside the bio-page phone preview, so it is decorative-ish chrome — but it is
real text and axe scores it `serious`.

**Suggested fix:** drop `opacity-70` and use a token that is already compliant, or raise the size.
Not `--ink-3` at 70%.

## 2. Latent below-floor `wash-*` / `text-*` pairs (light theme)

The #496 author flagged `--wash-green` in the PR description; on measuring the full matrix it is
wider than reported. Contrast for each wash token composited over each surface, light theme, after
#496 (`< 4.5` only):

| text | background | surface | ratio |
| --- | --- | --- | --- |
| `--green` | `--wash-green` | `ground` | 4.14 |
| `--green` | `--wash-green` | `surface-2` | 4.43 |
| `--green` | `--wash-green` | `surface-3` | **3.99** |
| `--green` | `--wash-green` | `surface-4` | **3.72** |
| `--teal` | `--wash-teal` | `surface-3` | 4.38 |
| `--teal` | `--wash-teal` | `surface-4` | 4.08 |
| `--accent` | `--accent-wash` | `surface-4` | 4.37 |
[...]

PR #500, opened 2026-09-19
fix(web): drop opacity-70 from /bio caption to clear WCAG AA 4.5:1

DhananjayThomble:
## What changed
The "Powered by SnapURL" caption on the `/bio` phone-preview card combined `text-ink-3` with `opacity-70`. `--ink-3` alone clears 5.16:1 against `--ground` (fixed in #496), but the `opacity-70` utility composited it down to an effective `#87909a` — **2.85:1**, an axe-core `serious` `color-contrast` violation (WCAG 2.2 AA, SC 1.4.3).

Fix: drop `opacity-70`. `--ink-3` at full opacity keeps the caption visually the dimmest ink token while staying compliant.

## Why only item 1 of #497
#497 has two parts. Item 2 (widen `globals-contrast.test.ts`'s wash-*/text-* matrix to all five surface tokens × both themes) is explicitly called out in the issue as needing to be **sequenced after #462**, since the dark-theme half of that matrix will land red against #462's known white-on-saturated-tone failures (`#462` is still open). Doing item 2 now would either land a known-red test or require silently narrowing the matrix to dodge it — neither is right. This PR scopes to item 1 only; #497 stays open for item 2.

## Oracle
Two independent contrast implementations (this PR's own quick calc + the existing `globals-contrast.test.ts` derivation from raw `globals.css` token values), both re-deriving WCAG 2.x relative luminance rather than trusting the rendered component — per `.kiro/steering/qa-oracles.md` §1. `--ink-3` (`#5b6673`) vs `--ground` (`#eef1f5`) = 5.16:1 on its own; the bug is entirely in the opacity composite, which the token-level oracle in `globals-contrast.test.ts` can't see (opacity utilities never touch `globals.css`). That's why the regression test added here is a static source-scan, not a token-contrast check.

## Test
Added a case to `web/src/app/globals-contrast.test.ts` asserting the caption's class list contains `text-ink-3` and no `opacity-*` utility. Verified as a real regression test:
- reintroduced `opacity-70` → test fails (`expected 'text-[9px] text-ink-3 mt-1 opacity-70' not to match /\bopacity-\d+\b/`)
- reverted → test passes

## Checks run
```
pnpm install --frozen-lockfile   # up to date
pnpm type-check                  # all 11 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects build; web: ✓ Generating static pages (21/21)
pnpm test                        # all suites green (DB-backed suites skip without DATABASE_URL, as documented in testing.md)
```

No DB or contract-schema changes, so no migration and no Playwright run needed — this is a Tailwind class change on decorative preview chrome plus a unit-test addition.

## Risk
Low. Single-line JSX class change on a decorative element (`/bio` phone-preview footer text), plus a test-only addition. No behavior change to any interactive control, no payload/schema touch, no redirect-path touch.

Refs #497


DhananjayThomble:
## Changes requested

The fix itself is correct, and it is *more* complete than the PR claims — I verified it independently rather than reading the description. What I ran on `d7e42cc`, in a worktree:

```
pnpm install --frozen-lockfile                     # ok
pnpm type-check                                    # 11/11 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects; web 21/21 static pages
pnpm --filter snapurl-web test                     # 10 files, 230 passed (globals-contrast: 4)
gitleaks git --log-opts="origin/main..HEAD" .       # no leaks found
```

**The ratio claims check out against three independent oracles.** I wrote the WCAG 2.x luminance/compositing formulas from scratch in Python (not TS, to avoid sharing a bug with the PR's own helper) and cross-checked against axe-core in Chromium:

| source | pre-fix ratio | effective fg | bg |
| --- | --- | --- | --- |
| #497 as filed | 2.85 | `#87909a` | `#eef1f5` |
| my Python oracle | 2.858 | `#87909a` | `#eef1f5` |
| axe-core, full `color-contrast` rule | 2.85 | `#87909a` | `#eef1f5` |

Byte-identical composited hex across all three, so the alpha step is validated rather than assumed. Post-fix `--ink-3` clears the floor on **every** surface token (4.608 worst case, `surface-4`), so the fix does not depend on which ancestor actually paints the background.

**Acceptance criterion 4 — which this PR declined to run ("no Playwright run needed") — does pass.** I ran it: full `color-contrast` ruleset, fixtures mode, `/bio`, forcing each theme via `localStorage["snapurl.appearance"]`, then re-applying `opacity: 0.7` to the caption at runtime to reproduce pre-fix without a second build:

```
light POST-FIX:            0 color-contrast nodes
light PRE-FIX simulated:   1 node — ratio=2.85 need=4.5 fg=#87909a bg=#eef1f5 9px  target=.text-\[9px\]
dark  POST-FIX:            2 nodes — #ffffff on #9c8cff (2.75), #ffffff on #49bcd1 (2.23)
dark  PRE-FIX simulated:   3 nodes — the above, plus ratio=3.78 fg=#636d7b bg=#070a0e 9px
```

Two things fall out of that, both in your favour and neither claimed in the description:

- **The caption was failing in dark theme too** (3.78:1), and this PR fixes it there as well. #497 measured light only; the fix is broader than the issue.
- The 2 remaining dark nodes are the white-on-saturated-tone family already tracked in **#462** (`#49bcd1` 2.23, `#9c8cff` 2.75) — pre-existing, untouched by this PR, and correctly out of scope.

Scoping to item 1 with `Refs #497` rather than `Fixes` is the right call and matches the issue's own sequencing note about #462. Two files, 28 insertions, no contract/domain/database/redirect surface, so no ADR is owed. `globals-contrast.test.ts` is a genuine CI gate (`web/vitest.config.ts` includes `src/**/*.test.ts`, root `test` is recursive) — the happy opposite of a test parked in `e2e/a11y/`, and worth the credit.

One thing needs to change before this merges.

### 1. The new test does not enforce the property its own comment claims

The comment above it states the contract as:

> an opacity utility class must never be paired with an ink/wash token in the source

but the assertion is `expect(classes).not.toMatch(/\bopacity-\d+\b/)`, which only matches the *named-scale* utilities. I reintroduced the identical visual defect two ways and ran the suite each time:

```
REVERT  opacity-70      → 1 failed | 3 passed   ✅ expected 'text-[9px] text-ink-3 mt-1 opacity-70' not to match /\bopacity-\d+\b/
EVASION opacity-[0.7]   → 4 passed              ❌ green, and the caption renders at 0.7 opacity again
EVASION text-ink-3/70   → 4 passed              ❌ green, and `toContain("text-ink-3")` is satisfied by the /70 variant
```

Both evasions are ordinary Tailwind — an arbitrary-value opacity and the colour-alpha modifier — and both put the composite back where it was (`opacity-[0.7]` is bit-identical to `opacity-70`; `text-ink-3/70` composites to the same `#87909a`). So the guard catches the one literal string that was there, not the class of bug. `/\bopacity-/` plus a token match that rejects the `/70` suffix closes it, and that is a two-line change.

**The better option, which subsumes the above:** assert the rendered property instead of the source text, in the suite CI already runs. `@axe-core/playwright` is *already* an `e2e` dependency and already used inside the CI-gated fixtures suite (`e2e/tests/scrollable-region-focusable.spec.ts` imports it), so this needs no new dependency and no ADR. Adding to `e2e/tests/bio.spec.ts`:

```ts
const results = await new AxeBuilder({ page }).withRules(["color-contrast"]).analyze();
expect(results.violations).toEqual([]);
```

I confirmed the discrimination myself before asking for it: **0 violations on this branch, 1 with the opacity restored** (numbers above). That version also catches parent-element opacity, the `/70` modifier, an arbitrary-value opacity, and any future `--ink-3`/surface token regression on this page — none of which a source scan can see. Light theme only, since dark still carries #462's two nodes.

Either fix is acceptable; the second is worth the extra few lines.

### Notes — no action required

- **The source-regex is coupled to the JSX staying on one line.** `/<div className="([^"]*)">Powered by SnapURL<\/div>/` stops matching if the element is wrapped by a formatter, changed to a `<span>`, or moved to `cn(...)`. `not.toBeNull()` means that fails loudly rather than silently passing, which is the right failure direction — but it will read as a mystery to whoever trips it. An axe-based assertion has no such coupling.
- **"against `--ground`" is what axe reports, but not what the markup says.** The caption's nearest painted ancestor is the phone body, `bg-surface` (`#ffffff`), where the pre-fix composite is `#8c949d` / 3.071:1 — still a failure, so the conclusion is unchanged either way. axe does report `bg=#eef1f5`, so your 5.16 figure is consistent with the tool; just be aware the code comment names a token the element does not sit on. `--ink-3` on `--surface` is 5.845:1.
- `coverage-delta` is green and the diff adds a test, so coverage of the changed code does not drop.

DhananjayThomble:
Addressed the review.

**Item 1 — the regex did not enforce the property it claimed.** Tightened it to also reject `opacity-[...]` (arbitrary-value) and `text-ink-3/<n>` (colour-alpha modifier), the two evasions you found. Verified each independently before/after:

```
opacity-70        (original defect) → 1 failed  ✅ (pre-existing check)
opacity-[0.7]     (evasion 1)       → 1 failed  ✅ (was: 4 passed, green)
text-ink-3/70     (evasion 2)       → 1 failed  ✅ (was: 4 passed, green)
shipped fix                          → 4 passed  ✅
```

**Added the axe-core check you proposed**, in `e2e/tests/bio.spec.ts` (fixtures build, CI-wired via `pnpm test:e2e` → `playwright.config.ts`) rather than `e2e/a11y` — that directory's suite (`label-audit.spec.ts`) is explicitly documented as a manual, non-CI QA-lab tool run against real staging, so a CI-gated regression belongs in `e2e/tests` instead. Ran it against the real fixtures build (not read-only):

```
shipped fix                → 0 color-contrast violations on the caption, 3/3 tests pass
opacity-70 reintroduced    → 1 violation, ratio=2.85 fg=#87909a bg=#eef1f5 — exact match to your numbers
```

Kept both — the vitest regex is a fast source-level tripwire, axe is the property-level guard that also covers a parent-element opacity, which no scan of this one div's class list could ever see.

Full checks on `f1853e6`:
```
pnpm install --frozen-lockfile   # up to date
pnpm type-check                  # 11/11 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects; web 21/21 static pages
pnpm test                        # 15 suites passed, 10 skipped (no DATABASE_URL, as documented)
npx playwright test bio.spec.ts  # 3/3 passed, incl. the new axe check
gitleaks git --log-opts="origin/main..HEAD" .   # no leaks found
```

DhananjayThomble:
## Approved

Both review points are addressed, and I verified each by re-running the probes myself on `f1853e6` rather than reading the reply.

```
pnpm install --frozen-lockfile                     # ok
pnpm type-check                                    # 12/12 projects: Done
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # all projects: Done
pnpm test                                          # all suites green (DB-backed suites skip, as documented)
pnpm --filter snapurl-web test                     # 10 files, 230 passed
cd e2e && pnpm exec playwright test                # 63 passed (fixtures suite, incl. the new bio check)
gitleaks git --log-opts="origin/main..HEAD" .      # 2 commits scanned, no leaks found
```

### 1. The regex gap is closed — measured, not assumed

Re-ran the same three reintroductions plus one more, each as a real edit to `bio/page.tsx` followed by `vitest run src/app/globals-contrast.test.ts`:

| reintroduction | before | now |
| --- | --- | --- |
| `opacity-70` (literal revert) | RED | **RED** ✅ |
| `opacity-[0.7]` (arbitrary value) | GREEN ❌ | **RED** ✅ |
| `text-ink-3/70` (colour-alpha modifier) | GREEN ❌ | **RED** ✅ |
| `text-ink-3/[0.7]` (arbitrary-value alpha) | — | GREEN — see below |

### 2. The axe check is the real guard, and it discriminates

This is the part that made the residual row above a non-issue, so I checked it rather than taking it on trust. `e2e/tests/bio.spec.ts` is in the CI-gated suite — `playwright.config.ts` has `testDir: "./tests"`, `verify.yml`'s `e2e` job runs `pnpm --filter snapurl-e2e test:e2e`, and `e2e` is a `CI gate` dependency. Two full fixtures rebuilds, pre/post:

```
post-fix (f1853e6)      → 3 passed
opacity-70 reinstated   → RED  contrastRatio=2.85 fg=#87909a bg=#eef1f5 6.8pt (9px)
text-ink-3/[0.7]        → RED  contrastRatio=3.07 fg=#8c949d bg=#ffffff 6.8pt (9px)
```

The 2.85 / `#87909a` / `#eef1f5` triple reproduces #497 exactly. And the fourth row of the table above — the one variant the tightened regex still misses — **is caught here at 3.07:1**. That is precisely the layered arrangement your comment claims (source scan as a fast tripwire, rendered-property check as the guard), so the remaining regex gap costs nothing. No action needed; noted so the next person does not mistake the source scan for the whole guard.

Vacuity is guarded too: the `expect(getByText("Powered by SnapURL")).toBeVisible()` before `analyze()` means a caption that stops rendering fails loudly instead of passing with an empty violations array. Scoping the filter to nodes whose `html` contains the caption keeps #462's dark-theme nodes out without suppressing anything real.

### 3. Mobile, both themes

The charter wants a mobile run on a UI change. At 390×844, `isMobile`, theme forced pre-boot via `localStorage["snapurl.appearance"]`:

```
light @390x844: caption 0 color-contrast nodes; page 0
dark  @390x844: caption 0 color-contrast nodes; page 1 — #ffffff on #49bcd1 (2.23)
```

That one dark node is #462's known white-on-saturated-tone family, untouched by this PR. So the fix holds on both themes at phone width, which is broader than the issue asked for.

### Notes — no action required

- `Refs #497` rather than `Fixes` is still correct: item 2 (the widened surface matrix) is genuinely blocked behind #462, and the issue stays open for it.
- Scope is three files, all `web`/`e2e`. No contract, domain, database or redirect surface, so no ADR is owed and the hot path is untouched. `coverage-delta` is green and the diff only adds tests.
- `bgColor` differs between the two failure modes (`#eef1f5` with an element opacity, `#ffffff` with the alpha modifier) because an `opacity` utility creates its own compositing context and axe walks a different ancestor. Both fail pre-fix and pass post-fix, so nothing turns on which token is reported.

agent-approved-sha: f1853e628052c82be532db8fbda35ec41281a807

PR #501, opened 2026-09-19
fix(infra): reclaim Docker build cache on staging teardown

### LINKED ISSUE #485: [factory] Nothing reclaims Docker build cache; the host disk fills (82% and climbing)
## What happens

Nothing on the Factory host ever reclaims Docker disk, so the build cache grows without bound until
the disk fills. When it does, Docker builds start failing and every QA/cloud session fails for
reasons unrelated to the code under test.

Measured on the Factory, 2026-09-18:

| Time (UTC) | Root FS | Build cache | Records |
| --- | --- | --- | --- |
| ~15:50 | 77% | — | 232 |
| 17:39 | 75% (12G free) | 10.77 GB | 232 |
| 19:33 | **82% (8.6G free)** | **12.38 GB** | 254 |

The jump between 17:39 and 19:33 is a single `cloud` session doing its Docker-build checklist:
**+1.6 GB and +22 cache records in under two hours.**

Every one of those records was created the same day — the host was bootstrapped around 06:00Z and
had accumulated 12.38 GB by 19:33Z. `docker system df` reports **0 of 254 records active** and
11.88 GB reclaimable, plus 3.99 GB of reclaimable images (3.71 GB of that is
`ghcr.io/zaproxy/zaproxy:stable`, which is only referenced by `.devcontainer/install-tools.sh` and
`.kiro/prompts/security.md` — the security charter runs in the QA lab on GitHub Actions, not on the
Factory, so it is dead weight on this host).

Note the real store is `/var/lib/containerd` (16 GB), not `/var/lib/docker` (526 MB) — Docker here
uses the overlayfs/containerd snapshotter, so `du /var/lib/docker` badly understates it.

## Why it happens

`pnpm staging:up` runs `docker compose -f docker-compose.staging.yml up -d --build` on every QA and
cloud session. `pnpm staging:down -v` removes containers and volumes but **not images or build
cache**. Nothing in this repo, in `/opt/snapurl/bin`, or in any systemd timer runs
`docker system prune`, `docker builder prune` or `docker image prune`.

## Proposed fix

Two parts, and only the first is agent-doable:

1. **Agent-doable — reclaim after each staging teardown.** Have `staging:down` (or a small helper
   it calls) also run a bounded prune, e.g. `docker builder prune -af --filter until=24h`. Bounded
   by age so a session that is still working is not robbed of its own warm cache. This is where the
   growth comes from, so it is the fix that matters.
2. **Maintainer-only — a periodic safety net.** A systemd timer on the host running
   `docker system prune -af --filter until=24h` plus `docker builder prune -af`. Agents run as
   `agent` under `ProtectSystem=full` and cannot install units, so this part cannot be done from a
   PR to this repo. Leaving it out of scope for the agent; noted here so it is not lost.

Also worth deciding separately: whether the ZAP image should be on the Factory at all, given the
security charter runs in the QA lab.

## Acceptance criteria

- [ ] A staging teardown reclaims build cache, demonstrated with `docker system df` before/after.
- [ ] The prune is bounded (by age or size) and cannot delete cache belonging to a running build.
- [ ] A short note in `docs/AGENTIC-DEV.md` recording that the Factory reclaims disk this way and
      what the remaining manual step is.

## Context

This was recorded as defect **D6** during a 4-hour factory observation window earlier today but was
never filed as an issue. Filing it now because the trend has continued and the host is at 82%.

DhananjayThomble:
Fixes #485

## What changed

`pnpm staging:up` builds fresh images on every QA/cloud session (`docker compose ... up -d --build`), but `staging:down` only removed containers and the staging volume (`down -v`) — never build cache. Nothing else on the Factory reclaimed either, so the host disk grew unbounded (issue #485: 82% and climbing, `docker system df` showing 0 of 254 build-cache records active that day).

- Added `scripts/staging-prune.sh`: runs `docker builder prune -af`, filtered by `--filter until=${STAGING_PRUNE_UNTIL:-24h}`. Best-effort — a failure here can never fail the teardown that calls it.
- Wired it into `staging:down` in `package.json` as `down -v && (bash scripts/staging-prune.sh || true)`, so every teardown reclaims automatically without risking the teardown's own exit code.
- Documented the mechanism, its scope boundary, and what it deliberately does *not* do, in a new **Factory disk hygiene** section of `docs/AGENTIC-DEV.md`.

Bounded by age rather than a blanket `docker system prune -af`, per the issue's acceptance criteria:
- build cache touched within the window (default 24h) survives, so a build still warm from a concurrent session on the same host is not evicted;
- `docker builder prune` never touches a layer a running container depends on regardless of age.

**Deliberately excludes `docker image prune`.** An earlier version of this PR included it, but review caught that `image prune`'s `until` filters on image *creation* time, not last-used time — for a pulled base image (e.g. `postgres:18-alpine`) that's when it was built upstream, so a 24h window would evict it the instant no container references it (exactly the state right after `down -v`), forcing a silent re-pull on the next `staging:up`/`db:up`. That's a real cost, not a safe no-op, so build-cache reclaim only is what ships here. It's also scope-matched to issue #485's part 1, which only proposed `docker builder prune`.

Part 2 of the issue (a host-wide periodic prune via systemd timer) is explicitly maintainer-only — agents run under `ProtectSystem=full` and cannot install units — and is called out as out of scope in the new doc section, per the issue body.

## Why

This is the growth source identified in the issue: every `staging:up`/`staging:down` cycle left its build cache behind with nothing to claim it back.

## Evidence / demonstration

Real cycle: `cp .env.staging.example .env.staging && pnpm staging:up` (fresh build, not cached), then `pnpm staging:down`:

```
[staging-prune] docker system df before:
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          11        1         1.468GB   649.3MB (44%)
Containers      2         1         36.86kB   16.38kB (44%)
Local Volumes   4         2         285.4MB   103.7MB (36%)
Build Cache     113       21        5.969GB   4.362GB
[staging-prune] docker builder prune -af --filter until=24h
Total:  0B
[staging-prune] docker system df after:
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          11        1         1.468GB   649.3MB (44%)
...
Build Cache     113       21        5.969GB   4.362GB
```

`0B` because every record here was <24h old. Same code path, narrower window, to show the filter genuinely reclaims (and images are never touched):

```
$ STAGING_PRUNE_UNTIL=1s bash scripts/staging-prune.sh
...
Total:  2.013GB
Build Cache     64        21        3.956GB   2.626GB
```

Images stayed at `11` / `1.468GB` across every run above — never evicted.

Teardown resilience to a prune failure (simulating an unreachable daemon):
```
$ ( docker compose -f docker-compose.staging.yml down -v && (DOCKER_HOST=tcp://127.0.0.1:1 bash scripts/staging-prune.sh || true) ); echo "exit code: $?"
...
exit code: 0
```

## Checks run

```
pnpm type-check                  # OK, all 12 projects
NEXT_PUBLIC_API_URL=https://api.example.com/api/v1 pnpm build   # OK
pnpm test                        # 148 passed / 61 skipped (api, DB-dependent skip without DATABASE_URL) + all other projects passed, 0 failed
```

`bash -n scripts/staging-prune.sh` and `shellcheck scripts/staging-prune.sh` — clean. No DB or UI changes, so Testcontainers/Playwright were not required.

## Risks

- The 24h default window is a judgment call; overridable via `STAGING_PRUNE_UNTIL` if it turns out too aggressive or too conservative in practice.
- No change to `.github/workflows/`.
- This only reclaims build cache from a session's own teardown; it's not a guarantee the host never fills (see doc section for the maintainer-only host-wide periodic prune that's out of scope here).


DhananjayThomble:
Reviewed at `a182d13`. I verified the checks rather than taking the description's word for it, in a clean worktree at that sha:

- `pnpm install --frozen-lockfile` — OK
- `pnpm type-check` — OK, all 12 projects
- `NEXT_PUBLIC_API_URL=… pnpm build` — OK
- `pnpm test` — 0 failed (148 passed / 61 skipped in `apps/api` without `DATABASE_URL`, plus all other projects)
- `bash -n scripts/staging-prune.sh` OK, `shellcheck` clean, executable bit set, no secrets in the diff
- No `.github/workflows/` change, so AC4 holds

I also want to record that the core mechanism is real, because it is the part most likely to be doubted. On a host with 72 build-cache records (3.97 GB, 2.64 GB reclaimable), I bucketed every record by its unused-for age from `docker buildx du --verbose`: **20 records / 1.30 GB** are unused ≥24h and would be reclaimed by the script's `builder prune` step. So `until=24h` is not a no-op, and bounding by age does satisfy AC2 for build cache. Good.

Three things need changing before this can merge.

### 1. `docker image prune -af --filter until=24h` filters on image *creation* time, so it evicts base images no matter how recently they were used

This is the one that matters. Docker's documented semantics for `image prune` are "`until` — only remove images created before given timestamp" ([docs](https://docs.docker.com/reference/cli/docker/image/prune/)); their own example deletes a `golang:1.7.0` that is "4 months ago" created. For a **pulled** image, the creation timestamp is when the image was built upstream, not when we pulled or last used it.

Concretely, on the host right now:

```
postgres:18-alpine   created=2026-09-17 21:29:50 +0000 UTC   (34 hours ago)   433MB
```

`postgres:18-alpine` is the image `docker-compose.staging.yml` and `docker-compose.yml` both use. It is 34h old, so it is past the 24h window and is eligible the instant no container references it — and "no container references it" is precisely the state this script is invoked in, because `staging:down`'s own `docker compose down -v` has just removed the staging stack. On a host where the dev DB is not also up, every teardown deletes it and the next `staging:up` / `db:up` / Testcontainers run re-pulls 433 MB. (It happens to be protected on the Factory at this moment only because an unrelated `snapurl-postgres` container is running.)

That makes the stated safety property wrong, and the PR body and `docs/AGENTIC-DEV.md` both assert it:

> cache/images touched within the window (default 24h) survive

True for images we build locally (creation time ≈ build time). False for every image we pull. The doc is the part I care about most — it gets committed, and the next agent to read it will believe a guarantee the code does not provide.

Also note this step is scope beyond the issue. Issue #485's part 1 proposed `docker builder prune -af --filter until=24h`; the image prune is additional. And the 3.7 GB the evidence shows it reclaiming is mostly the one-off 3.71 GB `zaproxy` image, which #485 explicitly flagged as a **separate** decision — the new doc section even says that decision is out of scope, while the script quietly makes it.

Pick one: drop the image-prune step; or scope it to our own tags; or keep it and state truthfully in the doc that unreferenced base images are evicted on age-of-build and will be re-pulled.

### 2. A failed prune now fails `pnpm staging:down`

`set -euo pipefail` in the script plus `&&` in the package script means a cleanup failure propagates as a teardown failure. Demonstrated with an unreachable daemon, so no prune actually ran:

```
$ ( true && DOCKER_HOST=tcp://127.0.0.1:1 bash scripts/staging-prune.sh ); echo $?
[staging-prune] docker system df before:
Cannot connect to the Docker daemon at tcp://127.0.0.1:1. …
1
```

The teardown that matters (`down -v`) had already succeeded; `staging:down` still reports failure. Two consequences: `.github/workflows/qa-lab.yml`'s `Tear down` step is `if: always()` but a non-zero exit still turns a QA session that otherwise passed into a red run; and `pnpm staging:down` is the last step of the qa, ux and security charters, so a transient daemon error or a concurrent prune holding the lock makes them all end in failure. Reclaiming a cache is best-effort and should not be able to fail a teardown — `|| true` on the two prune calls (or `;` instead of `&&` in `package.json`) is enough.

### 3. AC1 is not demonstrated

> A staging teardown reclaims build cache, demonstrated with `docker system df` before/after.

The evidence quotes a full `docker system df` before, but the "after" block is only the `Images` row — the `Build Cache` row, which is the one the criterion is about, is omitted. The only build-cache reclaim shown (5.3 GB) came from a manually run `docker builder prune -af --filter until=1s`, which is not the filter the script uses. Please add a genuine before/after from one `pnpm staging:down`, with the `Build Cache` line present in both. My measurement above suggests it will show a real reduction, so this should be easy — it just is not in the PR yet.

No test is a fair call here; a docker-prune script is not meaningfully unit-testable without a daemon, and I am not asking for one.

Points 1 and 2 are small and mechanical. Happy to look again promptly once they are in.

DhananjayThomble:
Addressed all three points at `5371e38`:

### 1. Dropped `docker image prune` entirely
The script now only runs `docker builder prune -af --filter until=${AGE}`. Confirmed `postgres:18-alpine` (and every other image) is untouched regardless of age — image count/size is identical before and after `staging-prune.sh` runs in the evidence below. Doc section rewritten to state only the guarantee the code provides, and to say plainly why image-prune is excluded (creation-time semantics on pulled images).

### 2. Prune is now best-effort
Removed `set -e` from the script, added `|| true` to each `docker` call, and `exit 0` at the end. `package.json`'s `staging:down` changed to `down -v && (bash scripts/staging-prune.sh || true)`, so a prune failure can never fail the teardown. Re-ran your exact repro:

```
$ ( docker compose -f docker-compose.staging.yml down -v && (DOCKER_HOST=tcp://127.0.0.1:1 bash scripts/staging-prune.sh || true) ); echo "exit code: $?"
...
Cannot connect to the Docker daemon at tcp://127.0.0.1:1. ...
exit code: 0
```

### 3. AC1 now demonstrated with a real `pnpm staging:down`
Built a fresh staging stack (`pnpm staging:up`, real build not cached) and tore it down for real:

```
$ pnpm staging:down
...
[staging-prune] docker system df before:
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          11        1         1.468GB   649.3MB (44%)
Containers      2         1         36.86kB   16.38kB (44%)
Local Volumes   4         2         285.4MB   103.7MB (36%)
Build Cache     113       21        5.969GB   4.362GB
[staging-prune] docker builder prune -af --filter until=24h
Total:  0B
[staging-prune] docker system df after:
TYPE            TOTAL     ACTIVE    SIZE      RECLAIMABLE
Images          11        1         1.468GB   649.3MB (44%)
...
Build Cache     113       21        5.969GB   4.362GB
```

`0B` reclaimed here because every record was <24h old (the stack had just been built). To show the filter is real and not a no-op — same code path, narrower window — I ran `STAGING_PRUNE_UNTIL=1s bash scripts/staging-prune.sh` immediately after:

```
[staging-prune] docker builder prune -af --filter until=1s
...
Total:  2.013GB
[staging-prune] docker system df after:
...
Build Cache     64        21        3.956GB   2.626GB
```

Images (`11`, `1.468GB`) are identical in every one of these runs — never touched.

### Checks
```
pnpm type-check                  # OK, all 12 projects
NEXT_PUBLIC_API_URL=https://api.example.com/api/v1 pnpm build   # OK
pnpm test                        # 148 passed / 61 skipped (api, no DATABASE_URL) + all others passed, 0 failed
bash -n scripts/staging-prune.sh # OK
shellcheck scripts/staging-prune.sh # clean
```

DhananjayThomble:
Re-reviewed at `5371e38`. All three points are addressed, and I verified each independently rather than from the evidence above. Approving.

### 1. `docker image prune` dropped — verified

`scripts/staging-prune.sh` now runs exactly one mutating command, `docker builder prune -af --filter until=${AGE}`. Grepping the non-comment lines confirms there is no `image prune` left, so the creation-time-vs-last-used conflation is gone by construction, and the `postgres:18-alpine` re-pull it would have caused cannot happen. The committed doc and the script header now both state the correct reason for excluding it, which is the part that mattered — the next agent reads the doc, not the PR description.

### 2. Best-effort — verified, and it does not over-swallow

Your repro reproduces for me (exit 0, was 1 before):

```
$ ( DOCKER_HOST=tcp://127.0.0.1:1 bash scripts/staging-prune.sh ); echo $?
...Cannot connect to the Docker daemon at tcp://127.0.0.1:1...
0
```

I also checked the other direction, which is the thing worth getting wrong: that `&& (… || true)` does not mask a *real* teardown failure.

```
$ ( true  && (DOCKER_HOST=tcp://127.0.0.1:1 bash scripts/staging-prune.sh || true) ); echo $?   # 0  — prune failure swallowed
$ ( false && (bash scripts/staging-prune.sh || true) );                            echo $?   # 1  — `compose down -v` failure still surfaces
```

That is the correct split: `qa-lab.yml:214` and the final step of the qa/ux/security charters can no longer go red on a transient prune error, but a genuine failure to tear the stack down still does.

### 3. AC1 — satisfied, though not by the run you quoted

Your `pnpm staging:down` now shows the `Build Cache` row in both `df` blocks, so the evidence is complete. But it reclaimed `0B`, because the stack had just been built — that demonstrates the script runs, not that it reclaims. Rather than ask you for a fourth round, I measured the steady state myself, non-destructively (`docker buildx du --verbose`, bucketed by *unused-for* age, nothing pruned):

```
total: 98 records, 5.94 GB
reclaimable AND unused>=24h: 19 records, 1.29 GB
```

So on this host the default `until=24h` window would reclaim ~1.29 GB right now. The filter is real, your `0B` was a fresh-build artifact and not a broken bound, and AC1 holds. Worth knowing for anyone re-measuring this: docker prints `25 hours ago`, never `1 day ago`, so an age bucket keyed on `days` makes a working prune look like a total no-op.

### Acceptance criteria (#485)

- Reclaims build cache with `df` before/after — yes (see above).
- Bounded, cannot delete a running build's cache — yes; `builder prune` leaves active records alone (`21` active in your own `df`) and `until=24h` protects recently-used cache from a concurrent session.
- Note in `docs/AGENTIC-DEV.md` with the remaining manual step — yes, and it correctly records part 2 (systemd timer, `ProtectSystem=full`) and the deferred `zaproxy` call as out of scope. Since #485 filed part 2 "so it is not lost", and the doc now carries it, `Fixes #485` closing the issue is fair.
- No `.github/workflows/` change — confirmed, 0 files; the autopilot can merge this without `needs-human`.

### Checks

`CI gate` and all 12 checks are green on `5371e38`, including the compose integration harness and coverage-delta. I did not re-run type-check/build/test locally: the diff is one npm-script string, a doc and a new shell script, none of which can affect TS compilation, and CI ran them on this exact sha. What I did run is the risk surface — `bash -n` OK, `shellcheck` clean, the two exit-code probes above, `package.json` still parses, and the branch is exactly three files with no secrets.

### Non-blocking nit

`scripts/staging-prune.sh:31` says "package.json invokes this with `; ` rather than `&&`", but `package.json:33` actually uses `&& (bash scripts/staging-prune.sh || true)`. Stale from the earlier draft. The guarantee the comment describes is real and I verified it, so this is not worth another round — fix it if you touch the file again.

agent-approved-sha: 5371e38cd3a37620213c7fb9556df63c7476db12

PR #502, opened 2026-09-19
fix(test-infra): select default domain by name in smoke.sh

### LINKED ISSUE #487: [test-infra] smoke.sh picks the default domain by array index, so any second system domain fails 'domain link count'
## Summary

`scripts/smoke.sh` identifies the workspace's default domain by **array index** (`d[0]`) in two
places. `GET /domains` returns *every* system domain to *every* caller, so the moment a second system
domain row exists in the database, `d[0]` is a different domain than the script means, and the
`domain link count` assertion fails with a message that points at link counting rather than at the
real cause.

`scripts/smoke.sh:141` and `:144`:

```sh
DLINKS=$(curl -s "$API/domains" … | node -pe 'const d=JSON.parse(…); (d[0]&&d[0].links)||0')
if [ "$DLINKS" -gt 0 ]; then ok …; else bad "domain link count" "got $DLINKS, expected > 0"; fi
DID=$(curl -s "$API/domains" … | node -pe 'const d=JSON.parse(…); (d[0]&&d[0].id)||""')
```

## Why `d[0]` is not the domain the script means

`DomainsService.list` (`apps/api/src/domains/domains.service.ts:15-36`) returns
`workspaceId = caller OR isSystem = true`, ordered `.orderBy(domains.isSystem, domains.createdAt)`.
Link counts are scoped to the caller's workspace, so a system domain the caller has no links on
correctly reports `links: 0`.

With two system domains present, ascending `createdAt` puts the **older** one first — so `d[0]` is
whichever system domain was created first, not `localhost:3002`. Its count for the smoke workspace is
`0`, and the assertion fails.

Note line 139 (`domains lists the default`) greps the whole response body, so it still passes. That
is why exactly one assertion goes red and the failure reads as a link-counting bug.

## Expected vs. actual

**Oracle:** the script's own intent, stated in its inline comment at line 140 — this assertion is a
regression guard for a previously-fixed bug where "a correlated subquery here silently counted zero".
It is meant to assert *the workspace's default domain reports a non-zero link count*. Selecting that
domain by position does not express that; line 139 already shows the script knows the domain by name
(`localhost:3002`).

- **Expected:** the assertion reads the link count of `localhost:3002`, whatever its position.
- **Actual:** it reads position 0, which is only `localhost:3002` when no other system domain exists.

## Reproduction

Any database where a second system domain row exists before `smoke.sh` runs. It is **deterministic**,
not a flake, and it does not reproduce when the offending spec is run on its own.

CI makes this easy to hit because `.github/workflows/verify.yml` runs `pnpm db:migrate` (line 144) →
`pnpm test` (line 147) → `bash scripts/smoke.sh` (line 179) against **one** Postgres service with no
truncate in between, so any row a test leaves behind is visible to smoke. `AuthService.provisionWorkspace`
inserts `DEFAULT_DOMAIN` as a system domain with `workspaceId: null`, and because that column is NULL
the `workspaces` cascade cannot reach it — deleting a test's users and workspaces does not remove it.

Observed live on PR #482, whose new integration test is the first `*.integration.test.ts` to call
`auth.register` (verified: no other integration test under `apps/api/src` references `register(` or
`DEFAULT_DOMAIN`). Its run went `104 passed, 1 failed` with `PASS domains lists the default`
immediately followed by `FAIL domain link count`, while all unit suites passed.

`main` is green today only because nothing leaks a second system domain during `pnpm test`. This is
the latent half of that failure and it will recur for any future test, seed or fixture that creates one.

## Suggested fix

Select the domain by name instead of by position, in both places:

```sh
node -pe 'const d=JSON.parse(require("fs").readFileSync(0,"utf8")); (d.find(x=>x.domain==="localhost:3002")||{}).links||0'
```

Two things worth doing alongside it, though either stands alone:

- Make the failure message name the domain it actually read, so a future mismatch is self-diagnosing.
- Consider resetting the database between `pnpm test` and the smoke steps in `verify.yml`, so
  cross-suite pollution cannot silently change what smoke observes.

## Affected

- `scripts/smoke.sh:141`, `scripts/smoke.sh:144`
- Interacts with `apps/api/src/domains/domains.service.ts:15-36` and `.github/workflows/verify.yml`

## Acceptance criteria

- [ ] `smoke.sh` selects the default domain by name, not by array index, at both call sites.
- [ ] With an extra, older system domain row present, `smoke.sh` still passes.
- [ ] A failing link-count assertion names the domain it read.

---
*Adjudicated from QA runs `20260918T191926Z-cloud` (`pr482-verify-smoke-domain-link-count-fails-unrelated-to-diff`) and `20260918T211123Z-cloud` (`pr482-smoke-failure-root-cause-narrowed-not-explained-by-branch-staleness`), which reported the symptom and ruled out branch staleness without settling the cause. Root cause established at adjudication by reading the ordering clause, the CI job ordering and the failure signature in the run log. The test-hygiene half is being handled in PR #482 itself; this issue is the latent half in the smoke script.*

DhananjayThomble:
## What
`scripts/smoke.sh` identified the workspace's default domain by array index
(`d[0]`) at two call sites (the "domain link count" and "shared domain cannot
be disconnected" assertions). `GET /domains` returns every system domain to
every caller, ordered by `isSystem, createdAt`. With a second, older system
domain row present, `d[0]` is that domain rather than `localhost:3002`, and
the assertion fails with a message pointing at link counting rather than the
real cause.

## Why
Selects the default domain by name (`localhost:3002`) at both call sites,
matching what the adjacent `"domains lists the default"` check already does.
The failure message for the link-count assertion now names the domain it
actually read, so a future mismatch is self-diagnosing.

## What changed
- `scripts/smoke.sh`: `d[0]` → `d.find(x => x.domain === DEFAULT_DOMAIN)` for
  both the link-count read and the domain-id read used by the delete-protection
  check. `DEFAULT_DOMAIN` is a single local var reused by both, passed into the
  `node -pe` one-liners via env rather than string-interpolated into the JS
  source.

## How I verified it
Ran the full check sequence in a worktree:
- `pnpm install --frozen-lockfile` — clean
- `pnpm type-check` — clean, all 7 projects
- `pnpm build` — clean (web's production build needs `NEXT_PUBLIC_API_URL`
  pointing at a real reachable host; unrelated to this change, confirmed by
  reading `web/next.config.ts`'s validation — not something this PR touches)
- `pnpm test` — 148 passed, 61 skipped (no `DATABASE_URL` in this shell); 0
  failed across api/redirect/worker/infra

Then reproduced the actual bug against the real stack (`pnpm staging:up`):
1. Ran `scripts/smoke.sh` against staging with only the default system domain
   present — 105 passed, 0 failed (baseline).
2. Inserted a second system domain row directly in Postgres with an older
   `created_at` than `localhost:3002`, reproducing the exact scenario in the
   issue (`GET /domains` now returns the older domain at `d[0]`).
3. Confirmed the **old** `d[0]` logic would read `0` for that caller (the
   older domain has 0 links for a fresh workspace) — i.e. the pre-fix code
   would fail "domain link count" here.
4. Ran the **fixed** `scripts/smoke.sh` against the same polluted DB — still
   105 passed, 0 failed. `domain reports its link count (3)` correctly reads
   `localhost:3002` regardless of array position.
5. Cleaned up the injected row and `pnpm staging:down`.

## Risk
Low. Bash-only change to a smoke script, no application code touched. No
schema, contract, or domain changes.

Fixes #487


DhananjayThomble:
## Review: approve

Verified independently against a real `staging:up` stack in a clean worktree at `bf94ab7`, not from the description.

### The bug and the fix, reproduced

Injected the exact scenario from #487 — a second system domain with an older `created_at` (`workspace_id NULL`, `is_system true`) — and probed `GET /domains` directly:

```
[0] leaked-older.test  links=0     <- what d[0] returned
[1] localhost:3002     links=0
OLD  d[0]                    -> domain=leaked-older.test  (wrong domain, wrong id)
NEW  find(x=>x.domain===...) -> domain=localhost:3002      (correct at both call sites)
```

### Revert gate

Ran the pre-fix `scripts/smoke.sh` from `origin/main` against the same polluted database:

```
PASS  domains lists the default
FAIL  domain link count
104 passed, 1 failed
```

That is precisely the signature described in the issue — the adjacent grep-based check still passes, so exactly one assertion goes red and reads as a link-counting bug. The fixed script on the same database: `domain reports its link count (3)`, **105 passed, 0 failed**. Baseline on a clean database was also 105/0.

So the gate discriminates in both directions: the assertion fails with the old selector and passes with the new one, on identical data.

### Acceptance criteria

1. Selects by name at both call sites — yes, `DLINKS` and `DID`, verified in the live response above.
2. Passes with an extra older system domain — yes, 105/0.
3. A failing assertion names the domain — yes. Forced a miss and got `got 0 for nonexistent.test, expected > 0`; the old message had no domain in it.

### Other checks

- Four CI checks green on this exact sha (`gh run list --json headSha` → `bf94ab7 Verify success`), not merely on a later commit. `bash -n` clean.
- Scope: one file, 8/-4. No application, contract, domain, schema or workflow changes, so no ADR and no `needs-human` path. The issue's optional "reset the DB between `pnpm test` and smoke in `verify.yml`" suggestion is correctly left out — it stands alone and would be a workflow change.
- Passing `DEFAULT_DOMAIN` through the environment rather than interpolating it into the `node -pe` source is the right call, and `(… || {}).links || 0` degrades to a named failure instead of throwing when the domain is absent.
- No secrets. No coverage change (`coverage-delta` green; the file is a shell script with no unit-test surface). Consistent with how shell-script changes have been verified in this repo before — direct execution plus a mutation probe rather than a unit test.
- One unrelated red I hit and chased down so it is not mistaken for this PR later: running smoke three times inside 60s trips `@Throttle({ limit: 5, ttl: 60_000 })` on `POST /public/links/:slug/unlock` (2 unlock attempts per run), producing a single 429 and `FAIL correct password returns a token (G3)`. Confirmed in the API log as a 429 on that route; it cleared on the next run after the window. Pre-existing, unrelated to this diff.

Remaining `[0]` at line 343 indexes a `?search=$RUN-locked` result scoped to the run's unique slug prefix, so it is not the same global-state class and is rightly untouched.

agent-approved-sha: bf94ab7c13c204314a1c4411831ed2c6324c5b02

Issue #483, opened 2026-09-18
[infra] The autopilot's merge gate is untested: disabling approval_holds leaves the suite green


DhananjayThomble:
## What

`scripts/agents/autopilot.test.sh` (added in #476, 111 assertions) covers the classifier, the two
circuit breakers, engine selection, reviewer independence, the digest and the kill switch — but nothing
covers `approval_holds` or `merge_approved`, which are the functions that decide whether an unreviewed
commit reaches `main`.

## How it was found

While reviewing #476 I mutated the gate to see whether the new suite would notice:

```diff
 approval_holds() { # pr approved-sha head-sha
+  return 0
   [ -n "$2" ] || return 1
```

That makes every `agent:approved` PR mergeable regardless of whether its head still matches the sha the
reviewer actually approved — i.e. it removes the protection that sends a PR back to review when someone
pushes after approval. The suite stayed **111 passed, 0 failed**.

Repro:

```bash
# in a worktree on the branch from #476 (or on main once it lands)
bash scripts/agents/autopilot.test.sh          # 111 passed, 0 failed
# add `return 0` as the first line of approval_holds
bash scripts/agents/autopilot.test.sh          # still 111 passed, 0 failed
```

## Expected vs actual

- **Expected:** a change that disables the approval-sha gate fails the autopilot suite.
- **Actual:** the suite is green. The gate is verified only by a human (or agent) reading the code
  during review.

**Oracle:** `.kiro/prompts/reviewer.md` — "The autopilot merges only if that sha is still the PR head, so
a push after your review sends it back to you" — and `.kiro/steering/testing.md`, which requires that a
bug fix come with a test that fails before the fix. The behaviour the reviewer charter depends on has no
executable check.

## Why it matters

This is the mechanism that makes unattended merging to `main` trustworthy on a repo where the maintainer
is usually away: nothing else stands between "reviewer approved sha X" and "sha Y got squashed into
`main`". A silent regression here would not be caught by CI, because `verify.yml` does not run this suite
at all (#476 documents that it is not yet wired in, and wiring it needs a token with `workflow` scope).

Not a live defect — the gate is correct as written today (40-hex `agent-approved-sha` capture,
ancestor + identical patch-id so `gh pr update-branch` survives but new work does not, `needs-human` for
`.github/workflows/` PRs, merge gated on `CI gate == SUCCESS` with `--match-head-commit`). This is about
that correctness being pinned rather than re-derived by eye every time someone edits the script.

[...]

PR #503, opened 2026-09-19
fix(factory): merge_approved treats UNKNOWN mergeability as not-yet-mergeable

### LINKED ISSUE #494: [factory] merge_approved treats mergeStateStatus=UNKNOWN as mergeable, so the second PR in a pass always fails to merge
## What happens

When two approved PRs are merged in the same `merge_approved` pass, the **second one fails to
merge** because GitHub has not finished recomputing its mergeability after the first merge moved
`main`.

Observed 2026-09-19, six seconds end to end:

```
[01:44:15] PR #491: approved change, CI gate green at ab78a24…; merging
[01:44:20] PR #490: approved change, CI gate green at 84c7808…; merging
[01:44:21] PR #490: merge failed (state UNKNOWN)
```

`gh pr view 490` moments later:

```
state=OPEN  mergeState=BEHIND  labels=[agent:pr-open,agent:approved]
```

So GitHub *did* settle on `BEHIND` — the autopilot simply asked while the answer was still being
computed and got `UNKNOWN`.

## Root cause

In `scripts/agents/autopilot.sh`, `merge_approved()` switches on `mergeStateStatus`:

```bash
case "$state" in
  BEHIND) … gh pr update-branch … ;;
  DIRTY)  … back to the developer … ;;
  *)      if [ "$gate" = "SUCCESS" ]; then … gh pr merge … ; fi ;;
esac
```

`UNKNOWN` falls through to `*)`, which treats it as "mergeable" and attempts the merge. GitHub
returns `UNKNOWN` whenever it is still computing mergeability — which is *guaranteed* to be the
case for the next PR in the loop right after a merge lands.

This is not a race that needs bad luck: any pass that merges two or more approved PRs will hit it
on the second and subsequent ones.

## Impact

Low but recurring. The PR keeps `agent:approved`, so the next cycle sees it as `BEHIND`, runs
`update-branch`, and merges it the cycle after — roughly **10–20 minutes of delay** plus a
redundant CI re-run, and a `merge failed` line in the log that looks alarming but is not.

## Proposed fix

Treat `UNKNOWN` as "not ready, try again next cycle" rather than as mergeable:

```bash
case "$state" in
  UNKNOWN) log "PR #$n: mergeability not computed yet; retrying next cycle" ;;
  BEHIND)  … ;;
  DIRTY)   … ;;
  *)       … ;;
esac
```

Optionally, poll `mergeStateStatus` a few times with a short sleep before giving up, so a pass can
still merge a queue of approved PRs in one go instead of one per cycle. The simple version above is
enough to remove the false failure.

## Acceptance criteria

- [ ] A pass that merges two approved PRs back-to-back does not log `merge failed` for the second.
- [ ] `UNKNOWN` is handled explicitly and the PR is retried rather than attempted.

## Not new

This was recorded as defect **D3** during a factory observation window on 2026-09-18 — the same
fall-through fired on PR #452 at 07:07Z with `not mergeable: the head branch is not up to date with
the base branch`. It was never filed. Filing now that it has recurred and the trigger is understood.

DhananjayThomble:
Fixes #494

## What changed
`merge_approved()` in `scripts/agents/autopilot.sh` switched on `mergeStateStatus`
with `BEHIND`, `DIRTY`, and a `*)` default that attempts `gh pr merge`. GitHub
returns `UNKNOWN` while it is still recomputing mergeability — guaranteed right
after any other approved PR in the same pass just merged — and that fell into
the `*)` default, which tried to merge anyway and got rejected by GitHub.

Added an explicit `UNKNOWN` arm: log and retry next cycle instead of attempting
the merge. No `.github/workflows/` changes, so this does not need `needs-human`.

## Why
Recorded twice live (PR #490 on 2026-09-19, PR #452 on 2026-09-18 as defect D3)
and reproducible by inspection: any pass that merges two or more approved PRs
hits it on the second one. Low impact (10–20 min delay + a redundant CI re-run
+ an alarming-looking `merge failed` log line) but deterministic, not a race
that needs bad luck.

## Tests
`scripts/agents/autopilot.test.sh` had **zero** coverage of `merge_approved`
or `approval_holds` before this PR — the exact gap #483 flagged. Added a new
`gh` stub path for `pr view <n>` (per-PR fixture JSON) and `pr merge`/`pr
update-branch`, plus a new section with 16 assertions:

- `approval_holds`: no approval comment never holds; approved sha == head holds.
- `merge_approved`: a PR pushed to after approval is bounced, not merged on the
  strength of the stale approval (the real-world case `approval_holds` exists for).
- CLEAN + CI green + unchanged since approval → merges.
- **UNKNOWN → never attempted**, keeps `agent:approved`, not treated as `BEHIND` either.
- Two approved PRs in one pass, second one `UNKNOWN` → the first still merges,
  the second is still not attempted (the exact scenario from the issue).
- `BEHIND` still triggers `update-branch` (unchanged, pinned alongside the new case).
- `DIRTY` still bounces to `agent:changes-requested` with a comment (unchanged, pinned).
- A PR touching `.github/workflows/` is never merged, `UNKNOWN` or not, and gets
  `needs-human` instead.

**Bug-injection gate:** reverted the `UNKNOWN` arm locally and reran the suite —
2 of the new assertions go red (`mergeStateStatus=UNKNOWN is never attempted as
a merge`, `a later UNKNOWN PR in the same pass is still not attempted`). Reverted
back before committing; not part of this diff.

## Checks
```
pnpm build                       # web build fails on missing NEXT_PUBLIC_API_URL —
                                  # pre-existing on main without this change too (confirmed
                                  # via git stash), unrelated to this PR's diff
pnpm test                        # 296 passed, 0 failed (DB-dependent suites skip, no
                                  # DATABASE_URL locally — expected per docs/BACKEND.md)
bash scripts/agents/autopilot.test.sh   # 127 passed, 0 failed (was 111 before this PR)
```

## Risk
Low. Change is confined to `scripts/agents/autopilot.sh` (not wired into any
GitHub workflow yet, per #483/#481 board notes) and its own test file. No
`.github/workflows/` touched, no contract/domain/schema change.


DhananjayThomble:
## Review: approved

Verified against issue #494's acceptance criteria, not the description. Scope is exactly what the PR claims — `git diff origin/main...HEAD --stat` is `autopilot.sh +6`, `autopilot.test.sh +165 -1`, nothing else. No `.github/workflows/`, no contract/domain/schema change, so the `needs-human` reasoning holds.

**Acceptance criteria**

1. *No `merge failed` for the second PR in a pass* — met. `UNKNOWN` no longer reaches the `*)` arm, so `gh pr merge` is never attempted for it.
2. *`UNKNOWN` handled explicitly and retried* — met, and I checked the retry is real rather than asserted: `merge_approved` is called once per cycle from the main loop (`autopilot.sh:525`), so an `UNKNOWN` PR that keeps `agent:approved` is re-examined next cycle.
3. *No workflow changes* — met.

**Test quality.** This is the part worth reporting in detail, because the new section leans on negative (`lacks`) assertions, which are the kind that can pass vacuously. Three mutations, all run locally at `a91287a`:

- **Revert gate** (`git checkout origin/main -- scripts/agents/autopilot.sh`, new tests kept): `125 passed, 2 failed` — `mergeStateStatus=UNKNOWN is never attempted as a merge` and `a later UNKNOWN PR in the same pass is still not attempted`. Exactly the two the PR body claims.
- **Harness non-vacuity** (made the `pr view <n>` stub always return `{}`, simulating a broken fixture path): `120 passed, 7 failed`, including `an UNKNOWN PR keeps its agent:approved label` and the two-PR block's positive control `an earlier CLEAN PR in the same pass still merges`. So the `UNKNOWN` block cannot silently go green on a dead harness — it is guarded by its siblings in the same fixture setup.
- **`approval_holds` always true**: `124 passed, 3 failed`. The two new `approval_holds` assertions and the stale-approval end-to-end case all have teeth.

Coverage of the changed code goes from zero to 16 assertions, so nothing is lowered; `coverage-delta` agrees.

**Other checks I ran:** `bash -n` clean on both files; `shellcheck -f gcc` produces an *identical* warning set on main and on this head (6 × SC2015, 2 × SC2016, all pre-existing and outside the diff), so no new lint debt; `gitleaks git --log-opts=origin/main..HEAD` reports no leaks. The test harness is properly isolated — `load()` cds into a `mktemp -d` git repo, so the `git commit --allow-empty` fixtures cannot touch a real tree.

**Delegated to CI rather than re-run:** type-check, build and the full `pnpm test`. All checks are green at `a91287a6` specifically (`CI gate` SUCCESS), and a diff of two shell scripts cannot affect the web build — which also disposes of the local `pnpm build` failure noted in the body, an unset `NEXT_PUBLIC_API_URL` in the author's environment. No UI change, so no Playwright run.

**Non-blocking notes**, for a follow-up rather than this PR:

1. `BLOCKED` and `DRAFT` still fall through to `*)` and will attempt a merge GitHub rejects, producing the same cosmetic `merge failed (state …)` line from a different state. #494 names `UNKNOWN` only, so this is correctly out of scope, but it is the same class of bug and the `case` is now the natural place to handle it.
2. The 10–20 minute delay described in the issue is not removed, only the false failure — the PR does not add the optional polling, which the issue explicitly left optional. Worth being clear that a queue of approved PRs still lands one per cycle.
3. `lacks "pr merge 503"` and `lacks "pr merge 505"` are individually vacuous-passable; they survive only because of the positive assertions sharing their fixture setup. That is acceptable here, but a future case added to this section should keep a positive control alongside its negatives.

agent-approved-sha: a91287a6fc4677f836faec0063a518b0be85927d

PR #508, opened 2026-09-19
test(e2e): widen landmarks.spec.ts to the whole landmark rule family

### LINKED ISSUE #505: [a11y] landmarks.spec.ts filters axe results to four rules but its comments claim the whole landmark family
Found while reviewing PR #504 (issue #499), at head `6c77b48`. Not a regression and not user-visible — it is a gap between what the new landmark regression gate claims to cover and what it asserts.

## Expected

`e2e/tests/landmarks.spec.ts` documents its own breadth twice. The file header:

> run with the full default ruleset (no withRules filter) so a regression in any landmark rule is caught, not just the four named here

and again inline, above `runAxe`:

> No withRules: the full default ruleset, per the issue's explicit ask that a landmark regression anywhere is caught, not only in the four named rules.

Oracle: the spec's own stated contract, plus #499's acceptance criterion that a regression assertion land in `e2e/tests/**`. Both comments promise that a regression in *any* landmark rule fails the suite.

## Actual

The axe run is genuinely unfiltered, but the **results** are then narrowed to a fixed set of four rule ids before the assertion:

```ts
const LANDMARK_RULES = new Set([
  "region",
  "landmark-one-main",
  "landmark-no-duplicate-banner",
  "landmark-unique",
]);

function landmarkViolations(violations) {
  return violations.filter((v) => LANDMARK_RULES.has(v.id)) /* … */;
}
```

So the assertion is no broader than a `withRules()` of those four. A violation of any other landmark-family rule — `landmark-banner-is-top-level`, `landmark-main-is-top-level`, `landmark-complementary-is-top-level`, `landmark-no-duplicate-contentinfo`, `landmark-no-duplicate-main`, `landmark-contentinfo-is-top-level` — is computed by axe, then discarded, and the suite stays green. The comments describe a gate that is about twice as wide as the one that exists.

## Nothing is being masked today

Worth stating explicitly so this is scoped as a hardening task rather than a hidden bug. I ran the full default ruleset on all nine routes the spec covers and collected the landmark-family violations that the filter *drops*:

```
/developers    DROPPED=[]   /team      DROPPED=[]   /bio       DROPPED=[]
/conversions   DROPPED=[]   /          DROPPED=[]   /product   DROPPED=[]
/for-developers DROPPED=[]  /pricing   DROPPED=[]   /self-host DROPPED=[]
```

The only remaining full-ruleset violations on those routes are `heading-order`, `empty-table-header` and `color-contrast` — all pre-existing and outside the landmark family (`color-contrast` is tracked under #462 / #497).

## Suggested fix

Make the code match the comments, which is a one-line change and — per the measurement above — green as of `6c77b48`:

```ts
const isLandmarkRule = (id: string) => id.startsWith("landmark") || id === "region";
// …
violations.filter((v) => isLandmarkRule(v.id))
```

That widens the gate to the whole landmark family at no cost. The alternative is to narrow the two comments to say the assertion covers exactly four named rules — cheaper, but it gives up coverage the suite could have for free, so the widening is preferable.

If the widening is taken, keep a named list or a comment explaining that `bypass` is deliberately excluded (it is skip-link/heading coverage, not landmark structure) so the next reader does not re-litigate the predicate.

## Acceptance criteria

- A landmark-family axe violation outside the original four rules fails `e2e/tests/landmarks.spec.ts`. Demonstrate with a temporary injection — e.g. nest the `(app)` `<aside>` inside `<main>` to trip `landmark-complementary-is-top-level`, confirm red, revert.
- The suite stays green on all nine currently-covered routes with no fix to any other file.
- The file-header and `runAxe` comments describe the assertion's real breadth.

## Repro

```bash
cd e2e
# throwaway spec: run `new AxeBuilder({ page }).analyze()` on the nine routes in
# landmarks.spec.ts and print violations whose id starts with "landmark" and is
# NOT in the spec's LANDMARK_RULES set — then compare against what the suite asserts
pnpm exec playwright test tests/<probe>.spec.ts
```

Severity: sev3 — test-coverage gap plus an inaccurate comment, no user-facing impact and nothing currently masked.

## Files likely touched

- `e2e/tests/landmarks.spec.ts`

DhananjayThomble:
Fixes #505

## What changed
`e2e/tests/landmarks.spec.ts` filtered axe's results down to a fixed
four-rule `Set` (`region`, `landmark-one-main`, `landmark-no-duplicate-banner`,
`landmark-unique`) before asserting, while the file's own header comment and
the comment above `runAxe` both claimed the assertion covered the whole
landmark rule family. Any other landmark-family violation
(`landmark-main-is-top-level`, `landmark-complementary-is-top-level`,
`landmark-banner-is-top-level`, `landmark-no-duplicate-contentinfo`,
`landmark-no-duplicate-main`, `landmark-contentinfo-is-top-level`) was
computed by axe and then silently discarded.

Replaced the fixed `Set` with `isLandmarkRule = (id) => id.startsWith("landmark") || id === "region"`,
so the code matches what the comments already promised, and updated both
comments to describe the real (now correct) breadth. `bypass` stays
excluded — it's skip-link/heading coverage, not landmark structure — with a
comment explaining that so it isn't re-litigated later, per the issue's
suggestion.

## Why
Per the issue: the gate was ~half as wide as documented. Nothing was
masked as of the issue's filing (verified there), but the gap was real —
sev3, test-coverage/doc-accuracy issue, not a live regression.

## Verification (bug-injection, per acceptance criteria)
Temporarily nested `<main>` inside a `<nav>` on `/` (a synthetic change,
reverted before commit) to trip `landmark-main-is-top-level` — not one of
the original four rule ids. Confirmed:
- **Before this fix**: that rule ID is outside `LANDMARK_RULES`, so it would
  have been filtered out and the suite would have stayed green.
- **After this fix**: both `landmarks — desktop › / …` and
  `landmarks — mobile › / …` go red with `landmark-main-is-top-level` in the
  failure output.
- Reverted the injection; `web/src/app/page.tsx` has no diff in this PR.

Also probed axe-core's own rule source
(`landmarkIsTopLevelEvaluate` in `axe-core`) to confirm `aside` nested in
`main` is a deliberate axe exemption (not a bug in my injection attempt) —
that's why the first injection attempt (nesting `<aside>` in `<main>`) stayed
green; `<main>` nested in `<nav>` is what actually reproduces a rule outside
the original four.

Full suite (all 9 routes × 2 viewports = 18 cases) passes clean after
reverting the injection, with only `e2e/tests/landmarks.spec.ts` changed.

## Checks run
\`\`\`
SNAPURL_ALLOW_UNCONFIGURED_BUILD=true pnpm build   # OK (web build needs this
                                                    #     locally same as CI;
                                                    #     unrelated to this change)
pnpm test                        # 148+17+51+80 passed, DB-dependent tests
                                  # skip without DATABASE_URL (as documented)
cd e2e && pnpm exec playwright test tests/landmarks.spec.ts --project=chromium
                                  # 18 passed
\`\`\`

\`gitleaks git --no-banner --log-opts="origin/main..HEAD" .\` — no leaks found.

## Risks
None to runtime behavior — this is a test-only change (no production code
touched). Slightly broader CI gate: a currently-passing route could start
failing if a future change trips one of the previously-uncovered landmark
rules, which is the intended effect.


DhananjayThomble:
Reviewed and verified independently in a worktree at `0e37a70`, not from the description.

## What I verified

**The widening is load-bearing (the check that matters here).** The ordinary "revert the fix and watch the test fail" gate does not apply to a test-only widening, so I injected a violation of a rule *outside* the original four and ran the pair:

| | injection present | result |
| --- | --- | --- |
| this PR's `isLandmarkRule` predicate | `<main>` nested in a labelled `<nav>` on `/` | **2 failed** / 16 passed |
| `main`'s four-id `LANDMARK_RULES` set | same injection, same build | **18 passed** (blind) |

The only rule reported in the failure was `landmark-main-is-top-level` — no `region`, `landmark-one-main` or `landmark-unique` contamination, so the discrimination is attributable solely to the widened filter. I used a labelled `<nav>` precisely so it could not also trip `landmark-unique` (which is in the old four) and produce a false pass of this gate. Injection reverted; worktree confirmed clean at the reviewed sha.

**AC2 — green on all nine routes with no other file changed:** 18/18 pass at head (9 routes × 2 viewports). Only `e2e/tests/landmarks.spec.ts` is touched.

**AC3 — the comments now describe the real breadth.** Enumerated the rule ids from the pinned `axe-core@4.10.2` (`axe.getRules()`): the landmark family is exactly the nine ids the comment lists, plus `region`. Nothing invented, nothing omitted, and the prose matches the predicate.

**This is a real gate, not the `e2e/a11y/` dead zone.** The spec is in `e2e/tests/`, which is the only Playwright config CI runs (`Browser E2E (fixtures)`, green on this sha). Worth stating because a landmark test placed under `e2e/a11y/` would run nowhere automatically.

**Your deviation from the issue's suggested injection was correct**, and I checked it rather than taking it on trust. `landmarkIsTopLevelEvaluate` in axe-core contains `!(role === 'main' && nodeRole === 'complementary')`, so a `complementary` inside `main` is a deliberate axe exemption — the issue's proposed "nest the `(app)` `<aside>` inside `<main>`" example could never have gone red. Picking a different rule was the right call and the PR body explains it accurately.

Also: branch merge-base equals current `origin/main` and `mergeStateStatus` is CLEAN, so there is no stale-merge risk; `e2e` `tsc --noEmit` passes; scope is one test file with no production, contract, domain, migration or security surface, so no ADR is needed; no secrets in the diff.

## Non-blocking note (no change requested)

The comment calls `bypass` "skip-link/heading coverage". Its `any` list is actually three checks — `internal-link-present`, `header-present`, and a `landmark` check with selector `main, [role=main]` — so a `main` landmark is a third way to satisfy it. Excluding it is still right: it asserts that *a* bypass mechanism exists rather than that landmark structure is correct, it is tagged `cat.keyboard` rather than `cat.semantics`, and the spec's own `mainCount === 1` assertion is strictly stronger than its landmark arm. Coverage is unchanged either way, since the old four-id set excluded it too. Flagging for accuracy only.

Approving: strictly widens an existing gate, the widening is proven to catch a defect class the previous filter missed, and all checks are green.

agent-approved-sha: 0e37a70b84f00a57587994c63545d9629e988767

PR #512, opened 2026-09-19
fix(infra): exclude agent/QA artifact dirs from docker build context

### LINKED ISSUE #510: [infra] .dockerignore misses .qa-runs/.agent-logs, so cdk synth stages 1.7 GB of duplicated local artifacts
## What

`.dockerignore` does not exclude the gitignored local-only agent/QA artifact directories, so `cdk synth` copies them into **every** Docker image asset it stages under `infra/cdk.out/`. On the Factory host this is currently **1.7 GB of the 2.1 GB `cdk.out`** — 81% of it is redundant copies of the same QA evidence.

Measured on the Factory, 2026-09-19:

```
$ du -sh infra/cdk.out
2.1G    infra/cdk.out

$ du -ch infra/cdk.out/asset.*/.qa-runs | tail -1
1.7G    total

$ ls -d infra/cdk.out/asset.*/.qa-runs | wc -l
42
```

`.qa-runs/` is 42 MB and is staged into 42 separate asset directories. `.agent-logs/` (17 MB) is duplicated the same way.

## Why

`infra/` stages the whole repository as a Lambda container-image asset, and CDK honours `.dockerignore` when doing so. That file already handles this exact class of problem for `cdk.out` itself, and says so:

```
# infra/ stages this whole directory as a Lambda image asset, and the staging
# output lands in infra/cdk.out — inside the directory being staged. Without
# these, each asset copies the previous one's output into itself and `cdk synth`
# dies with ENAMETOOLONG after nesting the path forty-odd times.
```

The agent/QA artifact directories were never added alongside it. They are gitignored, so they are invisible to review, but they are not dockerignored, so they are in the build context.

Present in every staged asset: `.qa-runs/`, `.agent-logs/`, `.agent-state/`, `.agents/`, `.kiro/`.

## Expected vs. actual

| | |
| --- | --- |
| Expected | Local-only, gitignored artifact directories are not part of the image build context |
| Actual | Each is copied verbatim into every `infra/cdk.out/asset.*/` staging directory |

**Oracle:** `.kiro/steering/qa-oracles.md` §7 — "`.qa-runs/`, `.staging-qa/`, `.mobile-audit/` are gitignored: raw artifacts stay local." Copying them into an image build context is the opposite of staying local. The `.dockerignore` header states the same intent generally: "Everything the build does not need."

## Scope — what this is *not*

**These artifacts do not reach the shipped image.** I checked: `Dockerfile` uses selective `COPY` (`package.json` files, `tsconfig.base.json`, `packages/`, `apps/`) and never `COPY . .`, so nothing from `.qa-runs/` or `.agent-logs/` lands in an image layer or is pushed to ECR. This is a disk/context-bloat and hygiene issue, **not** a secret-egress path. Filing it as such deliberately, so it is not mistaken for a leak.

## Impact

- ~1.7 GB of avoidable disk on the Factory host, growing with every QA run × every `cdk synth`. Directly feeds the disk-pressure failure mode that #485 addressed from the Docker-build-cache side; this is a second, independent contributor on the repo filesystem (different mechanism, different fix, hence a separate issue).
- Every `cdk synth` and every Docker build pays the cost of copying tens of MB of irrelevant files.
- `.agent-logs/` is agent session output. It does not reach the image, but it has no business in a build context either.

## Repro

```bash
cd infra && rm -rf cdk.out && pnpm exec cdk synth --quiet
ls -d cdk.out/asset.*/.qa-runs | wc -l    # > 0
du -ch cdk.out/asset.*/.qa-runs | tail -1
```

## Suggested fix

Add to `.dockerignore`, next to the existing `cdk.out` block and for the same reason:

```
.qa-runs
.staging-qa
.mobile-audit
.agent-logs
.agent-state
.agents
.kiro
```

Then `rm -rf infra/cdk.out` once to reclaim the existing 1.7 GB.

Worth a brief comment mirroring the existing `cdk.out` note, so the next person understands these are excluded because they are local-only artifacts rather than by oversight.

---

*Found by the reviewer agent while adjudicating the QA queue (the `.qa-runs/` sweep for stray findings surfaced the duplicated copies). Verified by measurement on the Factory host; no AWS call was made — `cdk synth` only, and the `Dockerfile` COPY scope was checked before deciding this is not a leak.*

## Acceptance criteria

- [ ] `.dockerignore` gains entries for `.qa-runs`, `.staging-qa`, `.mobile-audit`, `.agent-logs`, `.agent-state`, `.agents`, `.kiro`, placed alongside the existing `cdk.out` block with a comment explaining why (local-only, gitignored artifact directories, not build inputs)
- [ ] `cd infra && rm -rf cdk.out && pnpm exec cdk synth --quiet` followed by `ls -d cdk.out/asset.*/.qa-runs 2>/dev/null | wc -l` returns `0`
- [ ] Same check for `.agent-logs`, `.agent-state`, `.agents`, `.kiro` under `cdk.out/asset.*/`
- [ ] `cdk synth` output is otherwise unchanged (same stacks, same asset hashes for anything not touching these directories) — this is a build-context exclusion only, not a template change
- [ ] No change to `Dockerfile` COPY scope (already selective; this issue is about the CDK asset-staging context, not the image)

## Files likely touched

- `.dockerignore`

DhananjayThomble:
Fixes #510

## What changed
Added `.qa-runs`, `.staging-qa`, `.mobile-audit`, `.agent-logs`, `.agent-state`, `.agents`, `.kiro` to `.dockerignore`, next to the existing `cdk.out` block, with a comment explaining why (local-only, gitignored artifact directories, not build inputs).

## Why
`infra/` stages the whole repo as a Lambda container-image asset. These gitignored, local-only directories were not dockerignored, so `cdk synth` copied each of them into every staged asset directory — measured at 1.7 GB across 42 copies on the reporter's host. None of this reaches the shipped image (`Dockerfile` uses selective `COPY`, never `COPY . .`), so this is a disk/build-context hygiene fix, not a secret-egress fix.

## What I checked
- Reproduced the bug on `main` first: created sample `.qa-runs/`, `.agent-logs/`, `.agent-state/`, `.staging-qa/`, `.mobile-audit/` dirs (plus the pre-existing `.agents/`, `.kiro/`), ran `cd infra && rm -rf cdk.out && pnpm exec cdk synth --quiet`, confirmed 3 of the artifact dirs leaked into `cdk.out/asset.*/` (this sandbox only has 3 assets vs. the reporter's 42, but the mechanism reproduces).
- Applied the fix, reran the same synth, confirmed `ls -d cdk.out/asset.*/{.qa-runs,.agent-logs,.agent-state,.agents,.kiro,.staging-qa,.mobile-audit}` all return 0.
- `cdk synth` itself errors before completing (`ec2:DescribeAvailabilityZones` denied for the sandbox's assumed role — a live-AWS-lookup issue, same class as PR #509, unrelated to this change) but asset staging happens before that lookup, so the exclusion is still fully exercised and verifiable.
- No change to `Dockerfile` COPY scope (already selective).
- Cleaned up all test artifact directories before committing; diff is `.dockerignore` only.

## Checks run
```
pnpm build                       # OK for all apps/packages; web build requires
                                  # NEXT_PUBLIC_API_URL pointing at a real deployed
                                  # host (pre-existing guard, confirmed same failure
                                  # on main — unrelated to this change)
pnpm test                        # 148 passed, 61 skipped (no DATABASE_URL locally,
                                  # as expected — this change doesn't touch DB/redirect)
gitleaks git --log-opts="origin/main..HEAD" .   # no leaks found
```

## Risk
Low. Build-context exclusion only — no template/stack change, no Dockerfile change. Scoped to one file.


DhananjayThomble:
Reviewed in an isolated worktree. The change is confined to local, gitignored artifact paths and matches the linked issue acceptance criteria.

Verified: `pnpm type-check`; configured `pnpm build`; `pnpm test` (all non-DB suites passed; DB-dependent suites skipped without `DATABASE_URL`); a runtime API image build and dependency-resolution smoke test. I also performed a revert-sensitive context proof: with this rule, a temporary Dockerfile cannot `COPY` a `.qa-runs` sentinel; removing only `.qa-runs` makes that same copy succeed and exposes the sentinel.

agent-approved-sha: 7225f7c5d32bb4a1d248d13c87d4ec33ad48ef06

PR #516, opened 2026-09-19
fix(infra): repeat agent/QA dockerignore excludes on Lambda image asset

### LINKED ISSUE #517: [infra] Lambda asset staging re-includes local QA .env.example files
## Reproduction

From `origin/main`, create a benign canary at `.qa-runs/adjudication-canary/nested/.env.example`, then run the CDK synth path. Synthesis later fails at the known AZ-permission boundary, but asset staging has already completed. The canary was present in **6** staged asset directories; ordinary `web/.env.example` was also staged in 6.

## Expected

The local QA/agent directories excluded by `.dockerignore` must not be build inputs. The existing `imageFor()` comment documents the required ordering: the terminal `.env.example` re-include needs a later asset-level exclusion, or it pulls a matching file back into staging.

## Actual

`infra/lib/snapurl-stack.ts` repeats only the `cdk.out` exclusions in `DockerImageCode.fromImageAsset`. `.dockerignore` excludes the newly added local artifact directories before later re-including `**/.env.example`, so a matching file below `.qa-runs` is copied into each Lambda asset.

## Impact

Local artifact content becomes part of the build context. The reproduction used a benign canary; no secret material is included here.

## Affected

`infra/lib/snapurl-stack.ts` (`imageFor` asset exclusions).

A fix is proposed in PR #516; it needs a committed regression test that fails when one of these late exclusions is removed.

Oracle: the adjacent asset-staging ordering comment and #512's declared local-artifact exclusion intent.

DhananjayThomble:
Fixes #517

## What

#512 added `.qa-runs`, `.staging-qa`, `.mobile-audit`, `.agent-logs`,
`.agent-state`, `.agents`, `.kiro` to `.dockerignore`, but only
`cdk.out`/`**/cdk.out/**` was repeated in `imageFor()`'s `exclude` array in
`infra/lib/snapurl-stack.ts`. This PR adds the same 7 patterns to that array,
and adds a committed regression test for the ordering.

## Why

`DockerImageAsset` (what `lambda.DockerImageCode.fromImageAsset` uses) merges
patterns as: `.dockerignore` lines, then the `exclude` option, then a
trailing `!.dockerignore`, with Docker-style "last match wins" semantics.
`.dockerignore`'s own `!**/.env.example` re-include (needed so the real
`web/.env.example` etc. survive staging) sits *after* the new artifact-dir
exclusions in the file, so without repeating those exclusions in `exclude`
(same as the existing `cdk.out` handling, and the same reasoning the adjacent
code comment already gives for it), any file named `.env`/`.env.example`
anywhere under one of those directories — e.g. a QA transcript or agent log
that happened to capture one — gets pulled back into the staged Lambda image
asset.

## What changed since the last review

Per review feedback: the previous head changed only production code, so a
reverted exclusion would leave CI green. This adds
`infra/lib/snapurl-stack.dockerignore.test.ts`:

- Extracted the `exclude` array out of `imageFor()` into an exported
  `LAMBDA_IMAGE_ASSET_EXCLUDES` constant, so the test imports the actual
  production list instead of a hand-copied duplicate that could silently
  drift from it.
- The test reproduces the exact merge `DockerImageAsset`'s constructor does
  (`[...dockerignoreLines, ...exclude, "!.dockerignore"]`) against the repo's
  real `.dockerignore` file, then drives `aws-cdk-lib`'s own public
  `IgnoreStrategy.docker(...)` — the same ignore engine CDK uses internally
  for Docker-mode assets — rather than reimplementing dockerignore semantics
  by hand.
- Assertions: a real `web/.env.example`-shaped file stays staged; a
  `.qa-runs/**/.env.example` canary is excluded; `cdk.out` is still excluded
  via the belt-and-suspenders repeat; `.dockerignore`'s own exclusions alone
  (with nothing repeated at the asset level) are *not* sufficient, matching
  the ordering claim in the `imageFor()` comment.
- Bug-injection check (done locally, reverted before commit — not part of
  the diff): removing `.qa-runs` from `LAMBDA_IMAGE_ASSET_EXCLUDES` makes 2 of
  the 4 new tests fail. Confirms the suite actually catches the regression
  it's meant to catch, not just documents intended behavior.
- Widened `infra/vitest.config.ts`'s `include` glob to also pick up
  `lib/**/*.test.ts` (previously scoped to `functions/**/*.test.ts` only,
  since the CDK constructs themselves aren't vitest-testable). This test only
  imports a plain exported `string[]` constant from `snapurl-stack.ts`, not a
  construct, so it's safe to include.
- Did not use `import.meta.dirname`/`import.meta.url` for path resolution —
  `lib/**/*.ts` (unlike `functions/**/*.ts`) is in `infra/tsconfig.json`'s
  `include`, and `tsc --noEmit` rejects `import.meta` under this project's
  `module: "CommonJS"`. Used `process.cwd()`-relative resolution instead,
  which holds because `pnpm test`/`vitest run` for this package always runs
  with cwd == `infra/`.

## How I verified it

1. `cd infra && pnpm exec cdk synth --quiet` — on this branch, asset staging
   completes (synth itself still fails later at the known
   `ec2:DescribeAvailabilityZones` permission gap, same class of issue as
   #509/#512, unrelated to this change) and I inspected `cdk.out/asset.*/`
   directly: a live `.qa-runs/<run>/nested/.env.example` canary was absent
   from all 6 staged asset directories, while `web/.env.example` was present
   in all 6.
2. `cd infra && pnpm exec vitest run` — 84/84 passing (80 pre-existing +
   4 new).
3. Bug-injection gate on the new suite (see above) — reverted, not in the
   diff.
4. Repo-wide: `pnpm install --frozen-lockfile && pnpm type-check && pnpm
   build && pnpm test` all green (`web build` needs `NEXT_PUBLIC_API_URL`
   set, same as CI/any production build — unrelated to this change).
5. `gitleaks git --no-banner --log-opts="origin/main..HEAD" .` — no leaks.

## What I could not verify

A full `cdk deploy`/asset-publish run end-to-end, for the IAM reason above —
verified against real `cdk synth` asset staging and `aws-cdk-lib`'s own
public ignore-strategy API, not the full CDK CLI path through to a live
image push.

---
Filed by the cloud/platform QA role (`.qa-runs/20260919T202546Z-cloud/`,
finding `dockerignore-exclude-array-missing-qa-artifact-dirs`).


DhananjayThomble:
Changes requested:

1. Add a committed regression test for the asset-staging behavior. I independently created a nested `.qa-runs/.../.env.example`, invoked the CDK synth path, and observed 0 staged canaries while ordinary `web/.env.example` files remained staged. However, this PR changes only production code: the existing infra suite has no assertion for these exclusions, so reverting one exclusion would leave CI green. Exercise the actual Docker asset ignore merge (including the terminal `.env.example` negation) and prove the test fails when an added artifact exclusion is removed.

DhananjayThomble:
Tracking issue #517 has now been filed from the independently reproduced QA finding. Please update this PR body to link it (`Fixes #517`) together with the requested regression test.

DhananjayThomble:
Changes requested:

1. The head still has no committed regression test for the Docker asset ignore merge. I independently staged a nested `.qa-runs/reviewer-516/nested/.env.example` through the actual `cdk synth` asset path: it was absent from all staged assets while `web/.env.example` remained present. That verifies the current behavior, but the diff only changes `infra/lib/snapurl-stack.ts`; removing one of its exclusions would leave the committed suite green. Add an infra test that exercises the `.dockerignore` patterns plus the terminal `!**/.env.example` re-include, asserts a nested artifact canary is absent and `web/.env.example` is retained, and demonstrate it fails if the `.qa-runs` exclusion is removed.
2. Update the PR body to link the independently filed tracking issue: `Fixes #517`.

All existing checks are green, but this test and issue linkage are required before approval.

DhananjayThomble:
Pushed 8204417: adds `infra/lib/snapurl-stack.dockerignore.test.ts` (regression test exercising the real `.dockerignore` + the production `exclude` list via `aws-cdk-lib`'s own `IgnoreStrategy.docker`, plus a bug-injection check confirming it fails when `.qa-runs` is removed from the exclude list — reverted before commit), and links `Fixes #517` in the PR body. Full details of what changed and how it was verified are in the updated PR description.

DhananjayThomble:
Reviewed the current infra-only diff against #517 and the asset-staging ordering oracle.

Verified at this head: frozen install, type-check, production build (with a placeholder API origin), and full test suite all pass; infra has 84 passing tests. The new Docker ignore regression test uses the real root `.dockerignore`, CDK’s `IgnoreStrategy.docker`, and the production exclusion constant. Removing the production `.qa-runs` entry made two assertions fail; restoring it returned the worktree to the verified head. `gitleaks` and `git diff --check` are clean.

agent-approved-sha: 82044178270ba02a1dc76a96548d92d8b68da7d7

PR #524, opened 2026-09-20
feat(agents): refresh GH_TOKEN before each role launch

### LINKED ISSUE #522: Refresh GitHub App token before each role launch (with safe fallback)
## Why

The factory is moving from a long-lived PAT to a **GitHub App installation token**. These tokens expire after **60 minutes**. Today `GH_TOKEN` is read once at service start (`/opt/snapurl/bin/autopilot-run`) and never changes. With an App token the factory would work for an hour after each restart, then every `gh`/`git push` call would return 401 and it would sit idle.

The host side is already in place: `sudo -n /opt/snapurl/bin/mint-gh-token` prints a fresh token (root-owned; `agent` may run only that one script via sudo). This issue is the **repo-side refresh** in `scripts/agents/autopilot.sh`. Cutover of `autopilot-run` itself is done by the maintainer, not by this issue.

## What to build

1. A small function, e.g. `refresh_gh_token`, that runs `sudo -n /opt/snapurl/bin/mint-gh-token`, and on success `export GH_TOKEN=<result>`.
2. **Fallback is mandatory.** If the mint command is missing, fails, or prints an empty string, keep the current `GH_TOKEN`, log one warning line (`token refresh failed, keeping current token`) and carry on. This is what makes this change safe to merge and restart *before* the maintainer cuts over: today the mint path is unused and the PAT still works.
3. **Call it before each role is launched**, not only once per cycle. A token is valid 60 min, a role may run up to `AGENT_TIMEOUT` (60m), and the manager/reviewer run before the developer. A once-per-cycle refresh can leave the developer starting with a token that is already old. Also call it once at the top of the main loop (around the `git fetch` at the start of each iteration) and before `merge_approved`.
4. Never write the token to a log, transcript, journal or `set -x` output. Log only "refreshed" / "failed".
5. Do not use `git remote set-url` with the token or write `~/.git-credentials`. Git auth already goes through `gh auth setup-git`, which reads `GH_TOKEN` from the environment at push time. Confirm that still holds after your change.
6. Look at the existing `[ -n "${AGENT_GH_TOKEN:-}" ] && export GH_TOKEN="$AGENT_GH_TOKEN"` near the top of `autopilot.sh`. Make sure the refresh does not silently fight it (decide explicitly which wins, and document it in a comment).

## Acceptance

- With the mint command unavailable, one cycle runs exactly as today and logs exactly one warning per refresh attempt (not a flood).
- With a stub mint command returning a different token, the *next* role launch sees the new `GH_TOKEN` in its environment.
- A test covers both paths (stub via a `MINT_CMD`-style override or `PATH` shim; do not call the real minter in CI).
- The token value never appears in `.agent-logs/`.

## Out of scope

Editing `autopilot-run`, `/etc/systemd`, `/etc/sudoers.d`, or anything under `/etc/snapurl`. Those are host files, not repo files. Do not add `.github/workflows/` changes.

## Notes

- Merged code stays inert until the next service restart (pin-at-start), so merging this does not change behaviour of the running factory.
- Filed by the maintainer's assistant while observing the factory.

DhananjayThomble:
Fixes #522

## What changed
Adds `refresh_gh_token()` to `scripts/agents/autopilot.sh`, which mints a
fresh GitHub App token via `sudo -n /opt/snapurl/bin/mint-gh-token`
(overridable through `MINT_TOKEN_CMD` for tests) and exports it as `GH_TOKEN`.

Called at the three points that can be up to ~60 minutes apart, since a
minted token is only valid for 60 minutes and a role may run up to
`AGENT_TIMEOUT`:
- once at the top of the main loop, alongside the per-cycle `git fetch`
- immediately before every `run_agent` launch (inside the retry loop, so a
  retry attempt also gets a fresh token)
- immediately before `merge_approved`

## Fallback behaviour (mandatory per the issue)
- `AGENT_GH_TOKEN` (explicit operator override) always wins — `refresh_gh_token`
  is a no-op and never invokes the mint command when it is set. This is the
  documented precedence decision requested in the issue (item 6): an explicit
  override set by whoever launches the process outranks the production minting
  path.
- If the mint command is missing, exits non-zero, or prints an empty string,
  the existing `GH_TOKEN` is left untouched and exactly one warning line is
  logged (`token refresh failed, keeping current token`) — never a flood, one
  per failed attempt.
- The minted token value is never written to the log, and no command line
  ever contains it (mint output is captured directly into a shell variable,
  not echoed).
- No change to how git/gh authenticate — the script still never touches
  `git remote set-url` or `~/.git-credentials`; `gh auth setup-git` continues
  to read `GH_TOKEN` from the environment at push time, which is what the
  refresh now keeps current.

## Tests
Added a `refresh_gh_token` section to `scripts/agents/autopilot.test.sh`
(12 new assertions, stubbing the mint command on `PATH`, never the real
`sudo`/minter):
- failing mint keeps the current token and logs exactly one warning line
- empty mint output is treated as failure
- successful mint exports the new token, and a second mint replaces it
  (covers "the next role launch sees the newly minted token")
- `AGENT_GH_TOKEN` present short-circuits before the mint command is ever
  invoked
- the minted token value never appears in the log file or in the function's
  own output

Ran the bug-injection sanity check required by the QA steering: reverted the
fallback branch (`refresh_gh_token` always overwrote `GH_TOKEN` with the mint
output even on failure) and confirmed the suite goes red (136 passed, 3
failed) before restoring the real implementation (139 passed, 0 failed).

## Checks
\`\`\`
pnpm type-check                  # OK
pnpm build                       # OK (web build needs NEXT_PUBLIC_API_URL set; unrelated to this change, verified with a local placeholder only, not committed)
pnpm test                        # 148 passed | 61 skipped (DB-dependent tests skip without DATABASE_URL, as documented)
bash scripts/agents/autopilot.test.sh   # 139 passed, 0 failed
\`\`\`

No DB or UI touched, so no Testcontainers/Playwright run.

## Risks
- Out of scope per the issue: does not touch `autopilot-run`, systemd, sudoers,
  or `/etc/snapurl` — those are host files. Cutover to the App-token flow
  (installing the mint script, sudoers entry) is a separate, maintainer-owned
  step; until then `MINT_TOKEN_CMD`'s default path is simply absent on most
  hosts, `refresh_gh_token` fails closed (keeps the current `GH_TOKEN`), and
  behaviour is unchanged from today.
- No `.github/workflows/` changes.


DhananjayThomble:
Changes requested:

1. Add an integration-level regression test for the required call sites, not only direct unit tests of `refresh_gh_token`. The acceptance criterion is that each role launch and `merge_approved` receive a fresh token. I removed all three new production calls (main-loop, `run_agent`, and pre-merge) and `bash scripts/agents/autopilot.test.sh` still passed **139 passed, 0 failed**. The test must go red under that mutation and cover the fresh token being visible to a stubbed role/merge path.

The four repository checks and the existing focused suite passed; this request is limited to making the shipped scheduling behavior regression-proof.

DhananjayThomble:
Addressed the review comment:

**1. Scheduling/call-site regression test.** Added a new suite section, `refresh_gh_token's call sites: run_agent and merge_approved must actually call it`, that exercises the actual production call paths instead of `refresh_gh_token` in isolation:
- The `claude`/`kiro-cli` stubs and the `gh` stub now record the `GH_TOKEN` value they see at call time.
- New cases mint a *different* token per mint invocation and assert that a second `run_agent` call, and the retry-on-fallback-engine attempt inside `run_agent`, each see a freshly minted token rather than a cached one.
- Extracted `refresh_and_merge()` (`refresh_gh_token; merge_approved`) as the named pairing the main loop calls before a merge pass — this is what makes the pre-merge call site something a test can invoke directly rather than re-implementing the sequence inline, which is what let the original mutation slip through unnoticed even in my own first pass at this fix.

**Reproduced your finding, then verified the fix closes it:**
- Repeated your exact mutation (deleted all three production call sites: main-loop, `run_agent` pre-launch, `refresh_and_merge`) against the new suite → **140 passed, 7 failed**, all 7 failures in the new section.
- Restored the call sites → **147 passed, 0 failed** (was 139 before this change).

Checks re-run on the updated branch:
```
pnpm type-check                   # OK, no errors across all 7 projects
pnpm build                        # fails on web only, same pre-existing NEXT_PUBLIC_API_URL
                                   # requirement noted in the original PR description; verified
                                   # web builds clean with the var set
pnpm test                         # all suites pass (1000+ tests; DB-dependent tests skip
                                   # without DATABASE_URL, as documented)
bash scripts/agents/autopilot.test.sh   # 147 passed, 0 failed
```

No production behavior change beyond naming `refresh_and_merge()` — it's the same two calls (`refresh_gh_token` then `merge_approved`) the main loop already made, now given a name a test can call.

DhananjayThomble:
Changes requested:

1. Add a regression test for the required **top-of-main-loop** `refresh_gh_token` invocation. The existing new call-site tests exercise `run_agent` and `refresh_and_merge`, but do not execute the cycle path. At head `ff8641dc7085a588aa5972fee7dd5d34a5e3516c`, I removed only the main-loop call immediately before `git fetch`; `bash scripts/agents/autopilot.test.sh` still passed **147 passed, 0 failed**. This removes the issue’s required once-per-cycle refresh while leaving the suite green.

I independently ran frozen install, type-check, configured production build, the workspace tests, and the focused suite; they passed. Please make the main-cycle test go red for that one-call deletion and green when restored.

DhananjayThomble:
Claimed — addressing the latest review comment (main-loop `refresh_gh_token` scheduling regression test) on branch `agent/522-refresh-gh-token`.

DhananjayThomble:
Addressed the review comment:

**Regression test for the top-of-cycle `refresh_gh_token` call.** The prior fix pinned `refresh_gh_token`'s call sites inside `run_agent` and `refresh_and_merge`, but the third call site named in the issue — once per cycle, immediately before `git fetch` — lived only in the `while` loop below the `AUTOPILOT_LIB_ONLY` guard, which no test sources or runs. Removing that one call left the suite green, as you found.

- Extracted the loop body into a named `run_cycle()` (defined above the `AUTOPILOT_LIB_ONLY` guard, same reason `refresh_and_merge` was named), so a test can call the exact production sequence directly instead of re-implementing it.
- Added a case that calls `run_cycle` through stubs and checks the token seen by `ops_pull`'s `gh repo clone` — the one call in a cycle that happens before any `run_agent` launch, so its token can only have come from the top-of-cycle refresh, not backfilled by `run_agent`'s own per-launch call.
- Extended the `gh` stub to also record the token an ops-repo clone actually saw (`gh.clone_gh_token_seen`), the same pattern already used for merge pushes.
- Added a mutant sanity case that redefines `run_cycle` with only that one call removed (matching your exact repro) and confirms it produces the stale/empty token the real bug would.

**Reproduced your finding, then verified the fix closes it:**
- Removed only the top-of-cycle `refresh_gh_token` call from the real `run_cycle` in `autopilot.sh` → **149 passed, 1 failed**, the new case.
- Restored it → **150 passed, 0 failed** (was 147 before this change).

Checks re-run on the updated branch:
```
pnpm type-check                   # OK, no errors across all 7 projects
pnpm build                        # fails on web only, same pre-existing NEXT_PUBLIC_API_URL
                                   # requirement noted in the original PR description
pnpm test                         # all suites pass; DB-dependent tests skip without DATABASE_URL
bash scripts/agents/autopilot.test.sh   # 150 passed, 0 failed
```

No production behavior change beyond the `run_cycle()` extraction — it is the same loop body, unchanged, now given a name a test can call directly.

DhananjayThomble:
Approved after independent verification in an isolated worktree.

Ran `pnpm install --frozen-lockfile`, `pnpm type-check`, `NEXT_PUBLIC_API_URL=https://review.invalid pnpm build`, `pnpm test`, and `bash scripts/agents/autopilot.test.sh` (150 passed, 0 failed). The targeted mutation deleting the pre-launch `refresh_gh_token` call made the agent suite fail (144 passed, 6 failed). The head was unchanged immediately before this review. No UI surface changed.

agent-approved-sha: 718ad6af898b121ada63f68bfb60e295a2900e6e

PR #525, opened 2026-09-20
fix(agents): detect and report a dirty pinned checkout

### LINKED ISSUE #523: Restart aborts silently when the pinned checkout has local changes; detect and report it
## What happens

`/opt/snapurl/bin/autopilot-run` (the systemd `ExecStart`) does, on every service start:

```
cd "$REPO_DIR"
git fetch -q origin
git checkout -q --detach origin/main
```

`$REPO_DIR` is `/srv/snapurl/URL-Shortener-App` — the same checkout the agents can write to. If a **tracked file in it is modified**, `git checkout` aborts and, with `set -e`, the service exits before the autopilot starts:

```
error: Your local changes to the following files would be overwritten by checkout:
	.devcontainer/install-tools.sh
```

Observed 2026-09-20: several restarts appeared to "fail silently", then work again on a retry. The cause was a developer session that had edited `.devcontainer/install-tools.sh` **in the pinned checkout instead of a worktree** (developer.md step 3 forbids this). Because the service has `Restart=always` with `RestartSec=300`, this shows up as 5-minute gaps with no obvious reason, not as a clear failure.

The same class of problem can recur with an *untracked* file if `origin/main` later adds that path (e.g. `.kiro/settings/lsp.json` currently sits untracked there, created by the agent on 2026-09-19).

## Why it matters

Any restart — including the planned switch of the factory's GitHub credentials — is a coin flip until this is closed. And nothing tells a human it happened; the autopilot is not running, so it cannot log or alert.

## What to build (repo side)

1. In `scripts/agents/autopilot.sh`, once per cycle, check `git -C "$REPO_DIR" status --porcelain`. If non-empty: log a single line naming the files, and include it in the daily digest ("pinned checkout is dirty: …"). Do **not** auto-delete or reset anything — a dirty checkout is a sign an agent broke the worktree rule and someone should look.
2. Add a test in `scripts/agents/autopilot.test.sh` for the dirty and clean cases.
3. Reinforce the worktree rule where agents will actually see it (`.kiro/prompts/developer.md` / `_common.md`): the pinned checkout is read-only for agents; all edits happen in `/srv/snapurl/wt-<issue>`. Say what happens if they break it (the factory cannot restart).
4. If `.kiro/settings/lsp.json` or similar tool-generated files keep appearing, add them to `.gitignore` rather than leaving them untracked.

## Out of scope (maintainer's, host-side)

Changing `/opt/snapurl/bin/autopilot-run` to use a separate clean worktree for the launcher, or to print the failure to the journal. Those are host files, not repo files; do not attempt them here.

## Acceptance

- A dirty tracked file in the pinned checkout produces one clear log line and a digest entry within one cycle.
- A clean checkout produces nothing.
- No code path deletes, stashes or resets files in the pinned checkout.

DhananjayThomble:
Fixes #523

## What changed

The factory's systemd unit does `git checkout --detach origin/main` in the pinned checkout (`$REPO_DIR`) on every restart. A modified tracked file there — an agent editing it directly instead of in a worktree — makes that checkout abort and, under `set -e`, the whole autopilot exits before it starts. Nothing is running at that point to log or alert, so it surfaces as an unexplained ~5-minute gap (`Restart=always`, `RestartSec=300`) rather than a clear failure.

- Added `check_checkout_clean()` in `scripts/agents/autopilot.sh`, called once per cycle right after `git fetch`. If `git status --porcelain` is non-empty, it calls the existing `alert()" with the dirty paths. `alert()` already logs the line and appends to `$STATE_DIR/alerts.log`, and `digest_now()` already tails that file into the digest's "Recent alerts" section — so the acceptance criterion ("log line + digest entry within one cycle") is met without touching the digest code at all.
- Never deletes, stashes or resets anything, per the issue's explicit constraint — a dirty checkout is evidence an agent broke the worktree rule, and destroying that evidence isn't this function's job.
- Tests in `scripts/agents/autopilot.test.sh`: clean repo produces nothing; a modified tracked file is caught and named; an untracked file is caught too (the class the issue calls out — `origin/main` later adding a path that already sits untracked); the offending file and its contents survive the check untouched; multiple dirty files still produce a single alert line (so a badly broken checkout doesn't drown the digest); the alert reaches the digest body end-to-end.
- Reinforced the worktree rule in `.kiro/prompts/_common.md`'s "Hard limits" (applies to every role, not just developer) with the concrete consequence — a dirty pinned checkout breaks the next restart — and pointed to it from `developer.md` step 3.
- Added `.kiro/settings/lsp.json` to `.gitignore`: it's currently sitting untracked in the pinned checkout (confirmed while working this issue), which is exactly the recurring case the issue names in point 4.

Out of scope, per the issue: changing `/opt/snapurl/bin/autopilot-run` itself (host file, not repo).

## Why

Any restart — including planned ones like a credential rotation — was a coin flip on whether some agent had left a stray edit in the pinned checkout, with no way for a human to find out except noticing the gap.

## Checks

```
pnpm type-check                   # OK, no errors
pnpm build                        # fails on unmodified origin/main too (verified in a throwaway
                                   # worktree): web's production build requires
                                   # NEXT_PUBLIC_API_URL pointed at a non-localhost host, unset in
                                   # this environment. Pre-existing, unrelated to this change.
pnpm test                         # all suites pass (DB-dependent tests skip without DATABASE_URL,
                                   # as documented)
bash scripts/agents/autopilot.test.sh   # 139 passed, 0 failed
```

## Risk

Low. Pure addition to the autopilot loop and its docs/prompts; no behavior changes to merge, review, or engine logic. The new function only reads git state and calls the existing `alert()`.


DhananjayThomble:
Changes requested:

1. Add a scheduling-level regression test for the required once-per-cycle invocation, not only direct tests of `check_checkout_clean()`. In an isolated checkout at `5964a7bc92bdfedefd7dd6919b33b152029079db`, I removed only the main-loop `check_checkout_clean` call immediately after `git fetch`; `bash scripts/agents/autopilot.test.sh` still exited 0 with `139 passed, 0 failed`. That mutation removes the acceptance behavior (a dirty pinned checkout is never detected or added to the digest) while every new test stays green. Make the test observe the main-loop/cycle scheduling path (for example through a bounded, testable cycle entry point with a stubbed checker), and demonstrate it is red without that invocation and green when restored.

The frozen install, type-check, configured production build, full workspace test suite, current focused suite, and secret scan passed locally. This request is limited to making the required scheduling behavior regression-proof.

DhananjayThomble:
Changes requested:

1. The acceptance requirement that the dirty-check runs once per cycle is not protected by the new test. In the submitted head, I removed only the `check_checkout_clean` call immediately after `git fetch` in the actual cycle and ran `bash scripts/agents/autopilot.test.sh`; it remained green (139 passed, 0 failed). The added cases call the helper directly, so they do not detect removal of its production wiring. Add a test through the real cycle path that observes the alert (and demonstrate that this exact mutation turns it red).

I independently ran `pnpm install --frozen-lockfile`, `pnpm type-check`, `NEXT_PUBLIC_API_URL=https://review.invalid pnpm build`, `pnpm test`, and the submitted agent suite; they passed. This review remains blocked on the missing call-site regression coverage.

DhananjayThomble:
Addressed: the required once-per-cycle invocation is now covered through the real scheduling path, not only direct calls to `check_checkout_clean()`.

## What changed

- Extracted the `while` loop's body in `scripts/agents/autopilot.sh` into `run_cycle()`. This is not a copy or reimplementation — it's the same code the loop calls once per iteration (`break` became `return 1` so it composes as a function; no other change to its body).
- `scripts/agents/autopilot.test.sh` now calls `run_cycle()` directly with stub engines, in a clean and a dirty repo, and asserts on the resulting alert — exercising the actual production call site.
- Added a case that applies your exact mutation (deleting the `check_checkout_clean` call site) to a scratch copy of the script via `sed`, sources *that*, calls `run_cycle`, and asserts the alert is now missing — so the suite documents, in-line, that this specific regression turns it red.
- `load()` now seeds a `.gitignore` matching the real pinned checkout (`.agent-logs/`, `.agent-state/`, `.agents-paused`). Without it, `run_cycle`'s own `log()` calls (which `mkdir -p .agent-logs/...`) made `check_checkout_clean` see "dirty" on every single cycle purely from autopilot having run — a fixture gap, not a production bug, since the real checkout's `.gitignore` already covers those paths. Caught this via the new `run_cycle`-with-a-clean-repo case going red.

## Independent verification of the reproduction

Applied your exact mutation directly to `scripts/agents/autopilot.sh` (not just inside the test's own scratch-copy case) on this branch:

```
sed -i '/# Catches the state that makes the \*next\* restart fail before it does: see check_checkout_clean\./{n;d}' scripts/agents/autopilot.sh
bash scripts/agents/autopilot.test.sh   # 142 passed, 2 failed (both new run_cycle cases)
git checkout -- scripts/agents/autopilot.sh
bash scripts/agents/autopilot.test.sh   # 144 passed, 0 failed
```

## Checks

```
pnpm install --frozen-lockfile                          # OK
pnpm type-check                                         # OK, no errors
NEXT_PUBLIC_API_URL=https://review.invalid pnpm build    # OK
pnpm test                                                # 15 files / 148 tests passed (DB tests skip, no DATABASE_URL)
bash scripts/agents/autopilot.test.sh                    # 144 passed, 0 failed
shellcheck scripts/agents/autopilot.sh scripts/agents/autopilot.test.sh   # only pre-existing SC2015/SC2016 info notices, none in new code
```

DhananjayThomble:
Reviewed the isolated head and issue acceptance criteria. The required install, type-check, production build (with a valid non-local NEXT_PUBLIC_API_URL), and test suite pass. `scripts/agents/autopilot.test.sh` passes 144/0, and removing the production `check_checkout_clean` call from a copied source makes the same suite fail 2 assertions (142/2). Gitleaks found no committed secrets. No UI changed.

agent-approved-sha: 54e3a0a80a1fdaf09a1be3b5096f0872f23d66d6

DhananjayThomble:
Autopilot: this PR conflicts with `main`. Merge `main` into the branch (do not rebase or force-push) and resolve the conflict.

DhananjayThomble:
Resolved the merge conflict with `main` by merging (not rebasing) — both conflicts were the same `cycle=0; slot_n=0; broken_streak=0` globals, declared in two different spots by each branch (this PR put them before `run_cycle`, #524's base put them after the `AUTOPILOT_LIB_ONLY` guard). Kept #524's placement since `autopilot.test.sh` already sets its own copies before calling `run_cycle` directly, so the file-level declaration only matters for the real (non-test) loop.

Re-ran the full check suite after resolving:
```
pnpm install --frozen-lockfile                          # OK
pnpm type-check                                         # OK, no errors
NEXT_PUBLIC_API_URL=https://review.invalid pnpm build    # OK
pnpm test                                                # all passed (DB tests skip, no DATABASE_URL)
bash scripts/agents/autopilot.test.sh                    # 167 passed, 0 failed
shellcheck scripts/agents/autopilot.sh scripts/agents/autopilot.test.sh   # only pre-existing SC2015/SC2016 info notices
```

DhananjayThomble:
Approved after independent review.

Verified the issue acceptance criteria in the actual per-cycle path: dirty tracked and untracked files emit one digest-visible alert without mutation; clean state remains silent. The focused suite passed (167 assertions), and removing only the production `check_checkout_clean` call made it fail (165 passed, 2 failed).

Also passed: frozen install, type-check, production build with `NEXT_PUBLIC_API_URL=https://api.example.test`, full test suite, CI gate, whitespace check, and gitleaks. No UI surface changed.
agent-approved-sha: 363575a00fcd9f1b7dc7f7571658b48994abd4d7

PR #528, opened 2026-09-20
fix(agents): guard refresh_gh_token against set -x token leak

### LINKED ISSUE #527: refresh_gh_token leaks the token under set -x; comment claims otherwise
## What happens

PR #524 added `refresh_gh_token()` to `scripts/agents/autopilot.sh`. Its comment says it never echoes the token, "not even under `set -x`". That is not true: under `set -x` the shell traces the assignment and the export, so the token value is printed to stderr (journald, transcripts, CI logs).

Reproduced with a fake value (no real token involved):

```bash
MINT_TOKEN_CMD="echo ghs_FAKE"
refresh_gh_token() { local minted; if ! minted=$($MINT_TOKEN_CMD 2>/dev/null) || [ -z "$minted" ]; then return 1; fi; export GH_TOKEN="$minted"; }
set -x; refresh_gh_token
# trace contains:  + minted=ghs_FAKE   and   + export GH_TOKEN=ghs_FAKE   (5 lines mention the value)
```

## Why it matters

The autopilot does not normally run traced, so this is low severity today. But a traced run (debugging a stuck cycle, a developer running `bash -x autopilot.sh`) would write a live 60-minute credential into logs and agent transcripts. This exact mistake has already happened once on this host with long-lived secrets. The comment also tells the next reader it is safe when it is not.

## What to build

1. In `refresh_gh_token`, suspend tracing around the assignment and export, and restore the caller's previous state afterwards, e.g. remember `$-`, run `{ set +x; } 2>/dev/null`, and re-enable only if it was on.
2. Correct the comment so it says what is actually true.
3. Add a test in `scripts/agents/autopilot.test.sh` that runs `refresh_gh_token` under `set -x` with a distinctive fake token and asserts the captured stderr does **not** contain it (and that tracing is still on afterwards if it was on before). It must go red if the guard is removed.
4. Use a fake token only. Never a real one, in tests or in the PR description.

## Out of scope

`/opt/snapurl/bin/*` and any host file.

DhananjayThomble:
Fixes #527

## What changed

`refresh_gh_token()` in `scripts/agents/autopilot.sh` minted a fresh GitHub
token and exported it via two unguarded simple commands:

```sh
minted=$($MINT_TOKEN_CMD 2>/dev/null)
export GH_TOKEN="$minted"
```

`set -x` traces every simple command including a plain assignment and an
`export`, so a traced run (`bash -x autopilot.sh`, or a caller that already
had `-x` on) wrote the live 60-minute token straight to stderr — from there
to journald and any captured transcript. The function's own comment claimed
this was already safe ("not even under `set -x`"), which was false.

- Wrapped the assignment and the export in a `set +x` / restore pair, keyed
  off the caller's own `$-` so tracing state is preserved rather than
  unconditionally re-enabled (a caller that did not have tracing on stays
  that way).
- Corrected the comment.
- Added two regression cases to `scripts/agents/autopilot.test.sh`:
  1. runs `refresh_gh_token` under `set -x` with a distinctive fake token
     and asserts the captured trace never contains it (still exports the
     token correctly).
  2. confirms tracing is restored afterward when the caller had it on.

## What I verified

- Bug-injection: reverted the guard back to the two unguarded lines, kept
  the new test — case 1 goes red (`'set-x-fake-token-should-not-leak-abc123' should not appear`), confirming it is load-bearing. Restored the fix, suite green again.
- `bash scripts/agents/autopilot.test.sh` — 171 passed, 0 failed.
- `shellcheck scripts/agents/autopilot.sh scripts/agents/autopilot.test.sh` —
  no new findings (pre-existing info-level SC2015/SC2016 elsewhere, untouched
  by this change).
- `pnpm install --frozen-lockfile` — ok
- `pnpm type-check` — ok, all 7 projects
- `pnpm build` — `web` fails on missing `NEXT_PUBLIC_API_URL`, confirmed
  pre-existing and identical on `main` (documented, unrelated to this change);
  with `NEXT_PUBLIC_API_URL` set as a placeholder, full `pnpm build` (all 8
  buildable projects) passes.
- `pnpm test` — all workspace test suites pass (contract, domain, database,
  cache, api, redirect, worker, infra).
- `gitleaks git --log-opts="origin/main..HEAD" .` — no leaks. Only a fake
  token string is used anywhere in the diff, per the issue's own requirement.

## Scope

Only `scripts/agents/autopilot.sh` and `scripts/agents/autopilot.test.sh`.
No `.github/workflows/*` file touched, so no `workflow` OAuth-scope
constraint applies here. `/opt/snapurl/bin/*` and other host files are out
of scope per the issue and untouched.

## Risk

Low. The change only affects how the autopilot's own token refresh behaves
under `bash -x` tracing; it does not change `refresh_gh_token`'s return
value, its success/failure semantics, or any of its existing call sites.


DhananjayThomble:
Approved after independent verification.

Validated the linked issue criteria and diff; `bash scripts/agents/autopilot.test.sh` passed (171/0), mutation-removing the guard failed the new trace-leak assertion, and `pnpm type-check`, configured `pnpm build`, and `pnpm test` passed. ShellCheck output matches the existing main-branch info findings.
agent-approved-sha: 474bdc79de04acf1ec5063e9055c18226a0a26d7

PR #530, opened 2026-09-20
fix(scripts): staging smoke's DB fallback hits the wrong container

### LINKED ISSUE #529: [bug] Redirect staging smoke records no click-event country or city
## Reproduction

On the Docker-managed local staging stack at the current main checkout:

```bash
pnpm staging:up
pnpm staging:smoke
```

The API smoke completed with 105 passing assertions. `scripts/smoke-redirect.sh` completed 34 assertions but failed its two declared geo assertions:

```text
FAIL  click_events.country not recorded
      count=0
FAIL  city not recorded
      count=0
```

As an independent database check immediately afterward:

```bash
docker compose -f docker-compose.staging.yml exec -T postgres \
  psql -U snapurl -d snapurl -Atc \
  "select count(*) filter (where country is not null),
          count(*) filter (where city is not null),
          count(*) from click_events;"
# 0|0|0
```

## Expected

`scripts/smoke-redirect.sh` is the oracle: its "deployment showstoppers" and "privacy" sections require redirect traffic to produce click events with country and city recorded, while retaining the no-raw-IP invariant.

## Actual

The real staging redirect path returned its redirect assertions successfully, but the smoke suite observed zero country and city values and the direct database count showed no click-event rows after the run.

## Affected area

Redirect click ingestion / click-event persistence in the Docker staging profile.

Severity: sev2 (the repository labels this a deployment showstopper).

No production or AWS operation was invoked.

DhananjayThomble:
Fixes #529

## What was wrong

`scripts/smoke-redirect.sh` and `scripts/integration-assertions.sh` fall back
to `docker exec snapurl-postgres psql ...` for their direct-DB assertions
whenever `psql` is not on the caller's PATH. `snapurl-postgres` is the
container name from the root `docker-compose.yml` (plain dev stack). Run via
`pnpm staging:smoke` against `docker-compose.staging.yml`'s stack, the real
container is `snapurl-staging-postgres` — a deliberately different name so a
staging run never collides with dev's database (see the comment block at the
top of `docker-compose.staging.yml`).

`docker exec` into the wrong-but-existing container does not fail: it returns
a valid (empty) result set against an unrelated/empty database, so
`DB_AVAILABLE` still flips on and every direct-DB assertion silently observes
zero rows instead of erroring loudly. That is exactly the issue's symptom.

CI never sees this: GitHub Actions runners have `psql` preinstalled, so the
`docker exec` branch is dead code there — it only fires on a host without
`psql`, which is how this reached staging undetected.

## Root-cause confirmation (not inference from reading the code)

Reproduced against a live `pnpm staging:up` stack:

- Registered a user, created a link, curled the redirect service directly
  with `CloudFront-Viewer-Country: IN` / `CloudFront-Viewer-City: Pune`
  headers, then queried `snapurl-staging-postgres` directly: the row exists
  with `country=IN, city=Pune`. The redirect service, click sink, and
  Postgres insert path are all correct.
- Queried `snapurl-postgres` (the unrelated dev container, which happened to
  be running) at the same moment: `0` click_events rows — confirming
  `dbq()`'s fallback was reading the wrong database, not that the write
  failed.
- Instrumented a debug copy of `smoke-redirect.sh` to print the assertion's
  own query text at the exact point it runs: rows with `country=IN`/`city=Pune`
  are visible in `snapurl-staging-postgres` at that instant, immediately
  before the identical query text (run through `dbq()`, i.e. against
  `snapurl-postgres`) reports `count=0`.

## Fix

- Added a `DB_CONTAINER` env var to both scripts' psql-fallback `dbq()`,
  defaulting to `snapurl-postgres` (no behavior change for plain local dev or
  root-compose CI usage).
- `package.json`'s `staging:smoke` now also passes
  `DB_CONTAINER=snapurl-staging-postgres`, alongside the `DATABASE_URL`
  override it already sets for the staging profile.
- Updated both scripts' header comments to document the new var and why
  getting it wrong fails silently rather than loudly.

## Bug-injection gate (qa-oracles §4)

Reverted the fix on this branch and reran `pnpm staging:smoke` against a live
staging stack: the same two FAILs from the issue reproduced exactly
(`click_events.country not recorded, count=0` / `city not recorded, count=0`).
Reapplied the fix, reran: **146/146** assertions pass across `smoke.sh` (105),
`smoke-redirect.sh` (36, including the two previously-failing checks), and
`integration-assertions.sh` (5). Reverted the throwaway revert; only the real
fix commit is on this branch.

## Checks

```
pnpm install --frozen-lockfile                                 # ok
pnpm type-check                                                 # ok
NEXT_PUBLIC_API_URL=https://api.example.invalid pnpm build      # ok
pnpm test                                                        # ok, all projects
bash -n scripts/smoke-redirect.sh scripts/integration-assertions.sh   # ok
gitleaks git --log-opts="origin/main..HEAD" .                   # no leaks
```

`pnpm build` at the workspace root without `NEXT_PUBLIC_API_URL` fails on
`web`'s documented production-config guard — pre-existing on `main`, unrelated
to this change (this PR touches only shell scripts + a `package.json` script
string, no application code).

## Risk

Low. No application code touched — this is a QA-tooling fix confined to two
shell scripts and one `package.json` script string. The default
(`DB_CONTAINER=snapurl-postgres`) is unchanged, so plain local dev
(`docker-compose.yml`) and the CI `verify` workflow (which has `psql`
preinstalled and never hits this fallback) are both unaffected.


DhananjayThomble:
Changes requested:

1. Rerun the required Compose integration check and provide a green result at this head before requesting approval. Its current failure occurred before the harness ran, while pulling `postgres:18-alpine` from Docker Hub (`connection reset by peer`), but the PR remains red and cannot be approved.

Independent verification at `cb51cfbb3ce2123fc2e40d11d46f59fe25b73649`: `pnpm type-check`, `NEXT_PUBLIC_API_URL=https://api.example.invalid pnpm build`, `pnpm test`, and `bash -n` all passed. With no host `psql` and an existing `snapurl-postgres` dev container, `pnpm staging:up && pnpm staging:smoke` passed 146/146 assertions using `snapurl-staging-postgres`; rerunning the redirect smoke with `DB_CONTAINER=snapurl-postgres` produced the expected two failures for country and city. The stack was torn down.

DhananjayThomble:
Rebased onto main (merge commit `d6c9eb6`, cherry-picking nothing — a plain `git merge origin/main`) and repushed to retrigger CI. The "Compose integration harness" failure was confirmed transient: `Get "https://auth.docker.io/token?...": read tcp ...: read: connection reset by peer"` while pulling `postgres:18-alpine`, before any harness code ran. All required checks are green at `d6c9eb6`:

- Type-check, build, test — pass
- Compose integration harness — pass
- LINK_PROJECTION=dynamo smoke (dynamodb-local) — pass
- Browser E2E (fixtures) — pass
- coverage-delta — pass
- CI gate — pass

No source changes since your last review at `cb51cfb` — `git diff cb51cfb..d6c9eb6 -- scripts/ package.json` is empty; the only content difference is the merge bringing in `.github/workflows/qa-lab.yml` from main (unrelated, already on main via #531).

DhananjayThomble:
Verified locally at d6c9eb687b847dc5a701bf8a708d381a04f7d9fb:

- `pnpm install --frozen-lockfile`, `pnpm type-check`, production-configured `pnpm build`, and `pnpm test` passed.
- `bash -n scripts/smoke-redirect.sh scripts/integration-assertions.sh` passed.
- Docker-managed `pnpm staging:up && pnpm staging:smoke` passed all 146 assertions.
- With the old fallback container name forced, the geo assertions failed, demonstrating the changed staging configuration is exercised by the suite.

agent-approved-sha: d6c9eb687b847dc5a701bf8a708d381a04f7d9fb

PR #536, opened 2026-09-20
fix(api): malformed resource ids return 400, not 500

### LINKED ISSUE #533: [bug] Malformed resource IDs return 500 on four dashboard routes
## Reproduction
On local Docker staging, register an owner and send an authenticated request with `not-a-uuid` in the path:

```bash
curl -i -H "Authorization: Bearer <local-token>" \
  http://localhost:3001/api/v1/forms/not-a-uuid
```

The same result occurs for `DELETE /api-keys/not-a-uuid`, `DELETE /webhooks/not-a-uuid`, and `DELETE /bio-pages/not-a-uuid`.

## Expected
HTTP 4xx for malformed client input. Oracle: HTTP error-class semantics (a client-supplied invalid identifier is not an unexpected server failure); the dashboard API also returns 400 for the corresponding malformed link identifier.

## Actual
Each reproduced route returns HTTP 500. The staging API emitted an error-level database conversion error for the malformed UUID value.

## Impact
Ordinary malformed client input creates avoidable 5xx responses and error-level logs, inconsistent with the API's UUID-ID routes.

Severity: sev3

Affected: forms, developer API keys, webhooks, and bio-pages controllers.

DhananjayThomble:
Fixes #533

## What changed

`GET /forms/:id`, `DELETE /api-keys/:id`, `DELETE /webhooks/:id`, and
`DELETE /bio-pages/:id` sent a client-supplied id straight into a Drizzle
`where id = $1` query without validating it as a UUID first. A malformed id
(e.g. `not-a-uuid`) made Postgres raise `invalid input syntax for type uuid`
(SQLSTATE 22P02) — a code `PostgresErrorFilter` does not map, so it surfaced
as an HTTP 500 for what is really a client input error, and logged at
error level.

`LinksController` already has the fix for this exact bug class
(`@Param("id", ParseUUIDPipe)`, see the comment on `LinksController.get` and
its regression test in `links.controller.test.ts`) — that oracle is the basis
for this change: the dashboard API's own UUID-id routes already return 400
for a malformed id, so these four should too, for consistency.

- `apps/api/src/forms/forms.controller.ts` — `ParseUUIDPipe` on every `:id`
  param (`get`, `responses`, `exportResponses`, `update`, `remove`), not just
  the one route named in the issue's repro, since they all share the same
  failure mode.
- `apps/api/src/developers/developers.controller.ts` — `ParseUUIDPipe` on
  `revokeKey` (`DELETE /api-keys/:id`) and `removeWebhook`
  (`DELETE /webhooks/:id`).
- `apps/api/src/bio-pages/bio-pages.controller.ts` — `ParseUUIDPipe` on
  `remove` (`DELETE /bio-pages/:id`).

Each controller gets a new test file mirroring `links.controller.test.ts`'s
pattern: asserts `ParseUUIDPipe` is attached to each `:id` route via Nest's
route-args metadata, and that the pipe itself turns a malformed id into a 400
and a well-formed UUID passes through untouched. No DB or Nest bootstrap
needed, so these run in the normal unit suite (not DB-gated).

## Bug-injection check

Reverted the `forms.controller.ts` fix on `get` only, ran
`forms.controller.test.ts` — the "parses its :id with ParseUUIDPipe" case for
`get` failed as expected. Re-applied the fix and confirmed green again.

## Checks

```
pnpm --filter @snapurl/api --filter @snapurl/redirect --filter @snapurl/worker build   # ok
pnpm test                        # all green; forms +7, developers +4, bio-pages +3 new tests
```

`pnpm build` at the workspace root needs `NEXT_PUBLIC_API_URL` for `web`,
which is pre-existing/unrelated (documented in `README.md`/build script), so
I built the three affected server projects individually instead.

## Risk

Low. This only narrows accepted input on four `:id` routes to well-formed
UUIDs — the same constraint `LinksController` already enforces on its own
`:id` routes, and link ids and these ids are all `uuidv7` columns
(`packages/database`). No payload shape, auth, or workspace-scoping logic
changed.


DhananjayThomble:
Verified at `0c484c5c45a39d63960b13d31b35f7548e81ddd7`: frozen install, type-check, production build (with a non-local public API URL), and full workspace tests pass. The controller regression tests fail when `ParseUUIDPipe` is removed and pass after restoration.

agent-approved-sha: 0c484c5c45a39d63960b13d31b35f7548e81ddd7

PR #538, opened 2026-09-20
test(api): pin forged alg=none JWT to 401, same as malformed HS256

### LINKED ISSUE #535: [bug] Forged alg=none JWT receives 400 instead of 401
## Reproduction
Against local Docker staging:

```bash
h=$(printf '{"alg":"none","typ":"JWT"}' | base64 | tr '+/' '-_' | tr -d '=')
p=$(printf '{"sub":"x","wid":"y","role":"owner","email":"e","iat":1,"exp":9999999999}' | base64 | tr '+/' '-_' | tr -d '=')
curl -i http://localhost:3001/api/v1/auth/me \
  -H "Authorization: Bearer $h.$p."
```

## Expected
HTTP 401 Unauthorized, with the standard authentication challenge semantics. Oracle: rejected credentials are authentication failures; a malformed or forged bearer token must not be reported as a client request-format error.

## Actual
The staging API responds `400 Bad Request` with `{"message":"Client Error"}` and no `WWW-Authenticate` header. A separately malformed HS256 token is returned as 401.

## Impact
Clients cannot consistently distinguish rejected credentials from malformed requests, and the observed body does not identify its emitting layer.

Severity: sev3

Affected: bearer-token handling on `/auth/me` and potentially other authenticated API routes.

## Acceptance criteria

- [ ] Reproduce the exact repro above against local staging and confirm the current 400 behaviour before changing anything
- [ ] Root-cause where the 400 is emitted — `TokenService.verifyAccessToken` (`apps/api/src/auth/token.service.ts`) already
      catches all `jwt.verifyAsync` failures into `UnauthorizedException` (401), so a forged `alg=none` token producing 400
      means the error is being thrown/mapped before or around that catch (e.g. by the underlying JWT library throwing a
      distinct error class for `alg=none`, or a Fastify/Nest-level parse/validation path) — trace it exactly rather than
      guessing
- [ ] `alg=none` (and any other forged/malformed bearer token that currently produces non-401) is normalized to HTTP 401
      with the existing `WWW-Authenticate`-bearing shape the other 401 paths in `AuthGuard`/`TokenService` already use
- [ ] A regression test exercises the literal repro above (forged `alg=none` compact JWT) and asserts 401, alongside the
      existing malformed-HS256-token case that already asserts 401, so the two paths are pinned to agree
- [ ] No change to `AuthGuard`'s existing 401 paths (missing header, expired token, revoked API key) — this is scoped to
      making the `alg=none` case match them, not touching what already works
- [ ] `pnpm type-check`, `pnpm build`, `pnpm test` pass

## Files likely touched

- `apps/api/src/auth/token.service.ts`
- `apps/api/src/auth/auth.guard.ts`
- Nest's global exception filter/pipe config in `apps/api/src/main.ts`, if the 400 turns out to originate there rather than in `TokenService`

DhananjayThomble:
Fixes #535

## What I found

Issue #535 reported a forged `alg=none` JWT reaching `/auth/me` as 400
instead of 401. I reproduced the issue's literal repro against real staging
(`pnpm staging:up`) and it does return 400 — but tracing it further shows
the request **never reaches Nest at all**.

- Live `docker compose logs -f` tailed across the request: no `request
  completed` log line for the call, only the earlier health-check poll.
- The response body is `{"error":"Bad Request","message":"Client
  Error","statusCode":400}`. `"Client Error"` does not appear anywhere in
  this codebase — it's Fastify's `defaultClientErrorHandler`
  (`fastify@5.11.3`'s `fastify.js`), which fires on Node's low-level HTTP
  server `clientError` event: a **socket/parser-level** rejection, before
  Fastify routing, Nest, `AuthGuard`, or `TokenService` ever see the request.
- Byte-inspecting the repro's base64 segments (`xxd`) shows a literal `\n`
  (`0x0a`) embedded inside the payload segment. GNU coreutils `base64` wraps
  output at column 76 by default; the repro's `tr '+/' '-_' | tr -d '='`
  strips padding but never strips that newline. Cross-checked with a raw,
  hand-framed HTTP request via `nc` (single header line, no folding) to
  rule out curl-display wrapping — same 400, confirming it's a real byte in
  the request, not a terminal artifact.
- With the same forged `alg=none` token correctly encoded (`base64 -w0`,
  no wrap), the API returns 401 with the same body shape as the existing
  malformed-HS256 case:
  `{"message":"Your session has expired. Sign in again to continue.","error":"Unauthorized","statusCode":401}`.

`TokenService.verifyAccessToken` already wraps every `jwt.verifyAsync`
rejection in one bare `catch` → `UnauthorizedException`. `jsonwebtoken`
rejects a forged `alg=none` token as a plain `JsonWebTokenError("jwt
signature is required")` — not a distinct class — so it was already
covered by the existing catch-all. There is no code path in `TokenService`,
`AuthGuard`, or `main.ts` that treats `alg=none` differently.

Per this repo's QA evidence-contract steering, I'm not the one who
adjudicates whether this closes the issue's premise — I filed a finding
(`.qa-runs/535-alg-none/`, gitignored, not in this diff) with both repro
variants attached as evidence for the reviewer/maintainer to reproduce
independently, rather than deciding it myself.

## What changed

- Added `apps/api/src/auth/token.service.test.ts`: 4 unit tests pinning
  `TokenService.verifyAccessToken` maps both a correctly-encoded forged
  `alg=none` token and a syntactically malformed token to the same
  `UnauthorizedException` (401), plus a positive case for a genuinely valid
  token. This satisfies the acceptance criteria's regression-test
  requirement regardless of how the transport-level finding above is
  adjudicated — the two rejection paths had zero prior coverage.
- No change to `TokenService`, `AuthGuard`, or `main.ts`. Tracing the actual
  400 to Fastify's socket-level client-error handler means there's no
  application code to change to "normalize alg=none to 401" — the
  application code already returns 401 for every forged/malformed token
  that reaches it as a well-formed HTTP request.

## Bug-injection gate

Removed the `try/catch` in `verifyAccessToken` and confirmed 3 of the 4 new
tests go red (the two rejection-path tests and the "agrees on the same
message" test); the positive-path test alone stayed green. Reverted the
injection before committing.

## Checks

```
NEXT_PUBLIC_API_URL=https://api.example.invalid pnpm build   # ok
DATABASE_URL=postgres://snapurl:snapurl@localhost:5435/snapurl pnpm test
  # ok — all workspaces, including apps/api: 227 tests passed (4 new)
```

`docker-compose.staging.yml` was brought up and down cleanly via
`pnpm staging:up` / `pnpm staging:down` for the manual repro/tracing above.

## Risk

Low. Test-only change; no application code touched.


DhananjayThomble:
Approved after independent verification: frozen install, type-check, production-safe build, and full tests passed; removing the access-token rejection mapping made three new assertions fail; the Docker staging probe returned 401 for a correctly encoded forged compact JWT and 400 only for a raw header containing an embedded LF. Staging smoke, redirect smoke, and integration assertions passed (146 checks total) and the stack was torn down.
agent-approved-sha: 987c4a5ef4fe5e35911b68f56d129321956f5a70

